PR #43, opened 2026-09-06
feat: add Elixir static analysis support

mfreeman451:
Elixir repositories currently contribute no parsed definitions or call edges. This adds `.ex` and `.exs` support using the official [tree-sitter-elixir](https://github.com/elixir-lang/tree-sitter-elixir) grammar, vendored at `4b0c7118760af58a2e7081bbc8396e136f820b37` with its license and notice.

The extractor recognizes modules, protocols, functions, macros, guards, delegates, and literal ExUnit tests. It records local calls (including executable default expressions), static remote calls, and pipes; filters out declaration heads, quoted AST, module attributes, and protocol implementation bodies; and handles block and keyword bodies. The integration includes metrics, embedded queries, doctor, language catalogs, fuzz registration/seeds, and cache invalidation.

This is syntax-based support: macro expansion, alias/import/use resolution, dynamic dispatch, and protocol implementation dispatch are not inferred. Bare identifiers outside pipes are omitted because they may be variables or zero-arity calls. Arity narrowing remains disabled. These limits are documented in the README and architecture guide.

Validation:

- Before the review fixes: normal and Release builds; the Elixir gate passed on both and under ASAN.
- Full-repository ASAN scan; three byte-identical full-repository outputs and XML validation.
- Before the review fixes, the full gate suite plus failed-check reruns: 554 passed, two skipped, one pre-existing failure. `w3fixlegendcheck.sh` reports different `--dead-code=src` counts for relative and absolute roots under `/Users/mfreeman/src/ripwire`; the original HEAD binary reproduces the same 1-versus-8 counts with caching disabled.
- Formatting, dependency pinning, embedded-query coverage, cache identity, doctor, documentation, and output parity checks pass.
- Runtime implementation at `96cc09f2a`: Release build and Elixir regression gate passed; ASAN build, Elixir gate, and full-repository scan passed. README drift and generated-command checks passed. The installed Release binary also indexed ServiceRadar successfully.
- After review, a development build and focused Elixir, isolated-binary, configured fuzz-target, and extraction-cache checks passed. Added regressions verify nested/top-level `defimpl` isolation and executable defaults, including guarded defaults.
- New behavioral gate covers definitions, call edges and negatives, call-site mutation, metrics, cold/warm determinism, malformed input, and keyword-body signature elision. It fails against the pre-Elixir binary.

Implementation and generated output updates are separate commits, following CONTRIBUTING.md.

Upstream CI has not run yet: [run 34068757111](https://github.com/redhat-et/ripwire/actions/runs/34068757111) reports `action_required` with zero jobs, pending maintainer approval for this fork contribution.

Follow-up at `2baf1d80b` documents touched first-party functions and clarifies Elixir naming. The self-contained gate now compares every committed query name and exact contents with the generated embedded table while retaining isolated Elixir extraction assertions. Development and Release builds, installed-binary Elixir checks, build/installed self-containment checks, and a missing-query negative control passed. Formatting passed; advisory clang-tidy crashed locally (exit 139).

CodeRabbit docstring coverage is currently 61.76%, below its 80% threshold. Vendored grammar sources remain pristine; their undocumented functions contribute to the remaining warning.


joyful-ii-V-I:
**Merged** in `35f32aa1`. GitHub did not mark this PR merged automatically because the commits were rebased onto a moving main and their SHAs changed — your authorship is preserved on every one of them, and you are in the contributors graph. Closing it manually for that reason, not because anything was rejected.

Merged in `35f32aa1`. Thank you — this is the most complete language port this project has received.

Two things stood out. You wired it in the places contributors normally miss: `kExtSurfaceLangSlots`
derived rather than hardcoded, `symbolTally` sizing, the clones-mask `static_assert`, the lint
catalog, fuzz seeds. And you *improved* three existing gates while you were in there, replacing
brittle source-text counts with real behavioural assertions — the CMake file-API fuzz-target check
and the executed query interface are better than what they replaced.

Three changes on landing, all mine and all explained in the commits:

- **Extraction identity landed at 78, not 83.** Your re-pin log argued 78 had to be skipped as
  "77's rich value". `parserVerFor()` does derive rich as `kParserVer+1`, but `main.cpp`'s A4-P4
  split gives lean and rich *separate cache files*, so a lean-78 blob can never reach a rich-77
  reader — main's own history is 74→75→76→77, four consecutive +1 bumps. The house rule on rebase
  is to re-bump to the next free number over main's. Worth knowing for next time.
- **`defimpl` bodies are now indexed**, under the module name Elixir actually generates
  (`Protocol.For`). They were being dropped entirely, which lost every clause and call edge inside
  a protocol implementation. If there's a case where scoping that way is wrong, please say so —
  I changed a deliberate decision of yours and rewrote the `boundaries` assertions to match.
- **`elixircheck.sh` ran `--doctor` under `set -e`**, and `--doctor` exits 1 whenever the binary
  isn't the `ripwire` on PATH — including when there is none, which is every CI runner. That's why
  your `release` jobs never finished; they're the ones that run the gate battery. Fixed the way
  `doctorcheck.sh` does it.

Open follow-up if you're interested: `use`/`import`/`alias`/`require` currently mint no import
edges, so `--impact` and `--arch` can't see Elixir module dependencies. That's the next round here
and you'd be very welcome on it.

joyful-ii-V-I:
Landed in **[v0.5.0](https://github.com/redhat-et/ripwire/releases/tag/v0.5.0)** — thank you, this is the first outside contribution the project has shipped.

Closing the PR rather than merging it because the branch was rebased onto main's tip and landed as its own commits, so GitHub can't link it automatically. Your work is `f3cc02e3` (vendored grammar + call-graph extraction) through `0d67849c`, plus your review-round commits `d9c135c6` and `df9cfe1d`, all reachable from the tag:

    375660a2  merge(elixir): land PR #43 rebased — extraction identity 78, not 83
    f3cc02e3  feat: add vendored Elixir grammar and call graph extraction
    3021ee5f  feat(elixir): protocol implementations are indexed

Two notes on what changed in the rebase, so nothing is hidden:

- **Extraction identity is 78, not the 83 the fork carried.** Main had moved underneath, and the parser version had to follow the merged tip rather than the branch point.
- **`defimpl` protocol implementations are now indexed as their own definitions**, which was a gap neither of us could see from the fork — it only showed up against main's newer resolver.

Elixir also picked up **dependency edges** in the same release (`alias`/`import`/`require`/`use` resolve against the corpus's own `defmodule` index), alongside Bash, Lua and Ruby.

You're named in the [changelog](https://github.com/redhat-et/ripwire/blob/main/CHANGELOG.md) beside what you found. If the Elixir handling gets anything wrong on a real codebase, that's exactly the report worth filing.

PR #44, opened 2026-09-07
feat(platform): native Windows support (MSVC ABI + Clang, zero external deps)

lennix1337:
## Summary

This refresh advances the native Windows x64 port onto the current `main` line and hardens the Windows build, filesystem, process, cache, MCP, and test-harness boundaries without changing the Linux contract.

The implementation uses Clang/clang-cl with the MSVC ABI. The core CLI runtime needs only Windows system libraries; the optional `--run-trace` mode uses `sh.exe`/`bash.exe` when available and falls back to `cmd.exe`.

## What changed

- Added the native Windows platform compatibility layer for paths, UTF-8 filesystem operations, sockets, polling, process handles, and file locking.
- Added Windows-safe atomic cache/index writes, cache-directory selection, cache isolation, and stale-cache recovery.
- Added Win32 Job Object process-tree cleanup for timed-out subprocesses.
- Added native Windows handling for MCP edit locks, stale-index checks, watcher checks, and atomic qsnap publication/prefetch.
- Added fail-closed intermediate reparse-point inspection for Windows sidecar reads/writes: inspection errors refuse the operation instead of being treated as a clear path.
- Made Windows `--run-trace` use `sh.exe`/`bash.exe` when available, with correct Windows argument quoting so the documented `sh -c` contract, exit 127, timeout, and command-not-found behavior match the POSIX path.
- Added a native-Python XML checker fallback for `runtracecheck.sh` when Git for Windows does not provide `xmllint`.
- Fixed the Windows gate harness to preserve MSYS drive paths such as `/d/...` for `RIPWIRE_HEADBIN` instead of interpreting them as `C:/d/...`.
- Added the Windows companions needed by POSIX/FIFO-oriented gates while preserving the existing gate contracts.
- Updated CMake/portable flags and the Clang/MSVC build path, including the Windows manifest and `.exe` tool resolution.
- Hardened `test/pargates.py` scheduling, native stdin-pipe adapters, Windows process cleanup, per-gate budgets, and tree-write detection.
- Fixed `portablebuildcheck.sh` to use the configured native Python executable instead of relying on the broken Microsoft Store `python` alias.

## Validation

Environment: Windows 11 x64, Clang 22.1.8/clang-cl, MSVC ABI, Visual Studio 2019 Build Tools, `CMAKE_BUILD_TYPE=` and `RIPWIRE_LTO=OFF`.

- `cmake --build build --clean-first -j 6`: PASS. Both `ripwire_probe.exe` and `ripwire.exe` link successfully inside the `vcvars64` environment.
- `test/manifestcheck.sh`: PASS.
- `test/pargatescheck.sh`: ALL PASS.
- `test/portablebuildcheck.sh`: ALL PASS; only the expected Apple-Silicon/Darwin-only checks are skipped on this host.
- `sidecarsymlinkcheck.sh`: ALL PASS, 89/89 Windows checks, including intermediate/direct-volume reparse points and racing writes.
- `runtracecheck.sh`: ALL PASS, including `sh` exit 127, timeout, frameless failure, deterministic pricing, and XML validation.
- `mcpeditracecheck.sh`: PASS, 3 race trials.
- `mcpstalecheck.sh`: PASS.
- `mcpwatchercheck.sh`: PASS.
- `qsnapprefetchcheck.sh`: PASS.
- Final CLI smoke checks: `--version`, `--help`, and `--doctor` all pass.
- Final determinism check on `test/fixture`: two outputs are byte-identical.
- Final XML well-formedness check with `test/xmlcheck.py --noout`: PASS.
- No generated binaries or build artifacts are included in the branch diff.

The local test binary is also refreshed at `C:/Users/swluc/bin/ripwire.exe` and reports `built_from=d6e3f48d`.

## Full-suite status

The parametrized harness currently plans 631 gates across four deterministic shards, with a predicted budget of about 2,485 seconds per shard. A single `test/regression.sh` run exceeded the local 420-second execution window, so this description deliberately does not claim a full-suite pass. The focused Windows gates and harness contract checks above are the completed evidence from this workstation.

## Review notes

- The branch was synchronized with the current upstream `main` before the Windows changes were sent.
- `git diff --check` is clean.
- The remote PR head is now `d6e3f48d0c5c198fda81864eb6bb39af97aeb71d`.


joyful-ii-V-I:
Thank you for this, and for the scale of it — native Windows with the MSVC ABI and no external dependencies is a serious piece of work, not a port sketch, and it clearly took real effort.

I want to be straight with you rather than leave it sitting: **the open question on this one is not technical.** Adding a platform is a support commitment as much as a code change — every future PR, every release, and every bug report inherits it — and that is a scope decision I need to settle on our side before I can honestly review the code. It would be unfair to review it as though the answer were already yes.

So: no timeline, and no promise. What I can commit to is that you get a real answer either way rather than an indefinite open PR, and that if the answer is no it will be for that reason and stated plainly, not left to expire.

In the meantime, the Linux build runs unmodified under WSL2, and that is now documented in the README — which is a workaround, not an answer to what you built.

Sorry to have left this quiet for as long as I did.

lennix1337:
I understand. Thank you for the clear answer. In the meantime, I'll try to keep my branch updated.

matbeedotcom:
I ended up just using the wsl2 binary directly from windows- in case anyone comes across this in the future.

https://gist.github.com/matbeedotcom/e754922f39edeed297715d43382b9142

antoniojosedev:
Thanks for documenting the WSL2 workaround. I tested the current release path from a Windows host with WSL2 and Ubuntu 24.04.4 LTS.

Observed:

- The official installer works inside WSL: ripwire v0.6.0, checksum verification succeeded, `--doctor` passed all 8 checks, and a real `--for` query succeeded against a Windows checkout mounted under `/mnt/c`.
- From native PowerShell, the current release does not provide a `ripwire.exe`. To integrate it with a repository workflow that invokes a bare `ripwire` command, I had to add a machine-local `.cmd` shim forwarding arguments to the WSL binary. The shim works, including relative paths such as `src`, but it is tied to the local distro and installation path.
- The current source-build route is therefore not a practical Windows installation path until the native Windows work lands; the published installer also currently documents macOS/Linux targets.

This is not a request to duplicate the native-port work in this PR. It is installation and integration evidence from a real Windows consumer. Once native Windows support is accepted, a Windows release asset/installer would remove the remaining integration friction; for the interim, a short documented WSL-to-PowerShell wrapper recipe could help users whose tools expect `ripwire` on PATH.

joyful-ii-V-I:
this is in the plan soon now

what we would like to do in general is put all of the operating system switching at compile time in a single inline header (consteval and/or #if) , so windows/unix switching is not strung all over the code  and try to match posix/unix if possible for api calls into the header
(there maybe a framework worth pulling in, but it does not look like it is going to be that long)

this submission is close if we can move just move the operating system switching to a single compile time inline header
0.6.1 will be the last release for x86 Mac that way load in the CI will not increase
I actually do not have a windows machine to test on, but I * think * the GitHub CI can handle well, going to rely on the community.

joyful-ii-V-I:
Hi @lennix1337, thank you for sticking with this port through several refreshes onto a moving `main`. Native Windows support is what we're targeting for 0.6.2, and Windows x64 is taking the CI slot Intel macOS is giving up.

Before it merges we're restructuring where the OS switching lives. **We'll do that work ourselves, on this branch** (maintainer edits are enabled), so you don't have to redo it:

- **One header.** All OS-specific code moves into `src/infra/os.h`, namespace `rw::os`, with POSIX names, signatures and errno behaviour. Call sites read `os::lstat( p, &st )` and `os::rename( a, b )`, and `#if defined( _WIN32 )` appears only in that header.
- **What it replaces:** the force-included `platform_compat.h`, the `src/infra/compat/` POSIX shims, the libc-renaming macros (`#define rename rw_rename` and friends), and the ~70 `_WIN32` sites currently spread across 33 files.
- **Your implementations carry over.** Your Windows code moves into that header as-is wherever your Windows gates already prove it, e.g. the 89/89 `sidecarsymlinkcheck` run. Your 17 commits stay in the history: we merge, never squash or force-push.
- **Build and tests:** the `if(WIN32)` CMake blocks become one `cmake/Windows.cmake`, and the Windows branches in the test harness move into one shared helper.

**Two asks:**
1. Please hold new pushes to `win32-port` until our refactor commits are up (we'll comment here). If you have unpushed work, tell us now and we'll fold it in.
2. Once they're up, please run the Windows build and the gates you listed in the PR description on your machine. You have the only Windows workstation in this loop, and your run is the evidence we trust most.

Queue: #44 lands last — #231 → #237 → #236 → #235 + #233 → #224 → #44 — so the refactor rebases once over a settled `main` instead of six times.

lennix1337:
Hi @joyful-ii-V-I — I read your latest comment and followed the request to hold pushes to `win32-port`.

The PR branch on the remote is unchanged. Locally, I consolidated the work into one commit:

- local branch: `win32-port`
- local commit: `e795983f87d31696209a20e6874360ae26b780c6`
- parent: `d6e3f48d0c5c198fda81864eb6bb39af97aeb71d`
- remote status: local branch is 1 commit ahead of `fork/win32-port`; nothing was pushed
- working tree: clean
- commit size: 80 files, 3691 insertions, 1733 deletions

This commit should be treated as a local evidence/behavior snapshot, not as a competing replacement for the `src/infra/os.h` refactor you described. I will not push it over the PR while your maintainer sequence is landing.

What is implemented and locally validated:

- native Windows UTF-8/wide-path handling, including non-ASCII and long filesystem paths;
- centralized filesystem, process, socket, watcher, cache and executable-resolution compatibility;
- Win32 reparse-point protection for cache roots and ACL operations;
- `NUL` stdin for `--run-trace`;
- Job Object error handling and descendant cleanup;
- MCP watcher invalidation, root replacement coverage and linked-worktree `gitdir: C:/...` handling;
- Windows PATH/PATHEXT handling and valid JSON output from `wrap cursor`/`wrap opencode`;
- Windows/NMake/ClangCL standalone test-driver support;
- CRLF-preserving edit previews and native ACL-based edit-plan failure tests;
- Windows-native release installer and Python/path boundary handling;
- worker-budget changes that reduced the measured peak from 28–29 threads on 16 logical CPUs to 20.

The extreme temporary-path case is now functional rather than fail-fast. With real `TMP`, `TEMP` and `TMPDIR` values longer than 320 characters:

- normal map: `rc=0`, 2234 bytes;
- `--run-trace`: `rc=0`, 1308 bytes.

The focused Windows gates currently pass, including:

`runtracecheck`, `cacheisolationcheck`, `mcpwatchercheck`, `mcpstalecheck`, `qsnapprefetchcheck`, `githardencheck`, `opencodewrapcheck`, `docscommandscheck`, `editpreviewcheck`, `editplanrollbackmsgcheck`, `decltodefcheck`, `clonebandcheck`, `clonelexcheck`, `columnarcommacheck`, `connectcorecheck`, `dynmapsimdcheck`, `emitescapecheck`, `releaseinstallcheck`, `childwalkscalecheck`, `manifestcheck` and `pargatescheck`.

Builds that pass locally:

- clean Windows Release + ThinLTO/ClangCL (`/O2`, x86-64-v3, `/fp:precise`);
- Windows default build;
- Linux WSL Release + LTO;
- Windows ASAN;
- deterministic XML smoke and ASAN smoke.

Performance on the representative 480-file / 188640-byte corpus:

Windows native, same NTFS, 5 runs:

- cold: 126.857 ms
- warm retrieval: 41.967 ms
- report: 125.498 ms
- quality delta: 247.184 ms
- dead code: 136.455 ms
- MCP: 1.205 ms

Linux WSL, same NTFS, 5 runs:

- cold: 1441.196 ms
- warm retrieval: 1262.542 ms
- report: 1427.657 ms
- quality delta: 5251.653 ms
- dead code: 1918.417 ms
- MCP: 2112.960 ms

Windows is therefore substantially faster than WSL Linux when both use the same NTFS. Linux on its native WSL ext4 filesystem is still faster (cold 105.106 ms, warm 7.804 ms), so I am not claiming a universal Windows-over-ext4 result.

Remaining limitations:

- the complete `pargates` and `regression.sh` runs are still too long on this workstation; after roughly an hour they reported several general golden/table failures and were stopped without a trustworthy aggregate summary;
- `docdriftcheck` and `docdemotecheck` still need separate golden/count review;
- before the local commit, `--quality-delta` against `d6e3f48d` reported `gating=27`, and `--test-gate` reported 235 tests plus 1288 untested impacted symbols. This is not merge-ready evidence yet.

Recommended integration path for your plan:

1. land the maintainer refactor in the stated order: `#231 → #237 → #236 → #235 + #233 → #224 → #44`;
2. map the behavior above into `rw::os` instead of carrying both `platform_compat` and `os.h`;
3. preserve the Windows regression cases for reparse points, UTF-8/long paths, NUL stdin, Job Objects, watchers, worktrees, ACLs and CRLF;
4. rerun the focused Windows gates and the representative benchmark on the refactored branch;
5. only then decide whether any local behavior needs to be reapplied before the PR is merged.

The important result for the refactor is that the Windows behavior is now covered by concrete probes and numbers, while the PR itself remains untouched as requested.

joyful-ii-V-I:
Jeepers!  that is quite a measurement

Windows native, same NTFS, 5 runs:
cold: 126.857 ms
warm retrieval: 41.967 ms
report: 125.498 ms
quality delta: 247.184 ms
dead code: 136.455 ms
MCP: 1.205 ms


Linux WSL, same NTFS, 5 runs:
cold: 1441.196 ms
warm retrieval: 1262.542 ms
report: 1427.657 ms
quality delta: 5251.653 ms
dead code: 1918.417 ms
MCP: 2112.960 ms

code is close, pulled operating system switching to a single file or close, could use help testing it, it would be awesome if you can help testing it before it goes to main  (I don't have a windows machine) that would be awesome!!!


'will direct to do this: land the maintainer refactor in the stated order: #231 → #237 → #236 → #235 + #233 → #224 → #44;

joyful-ii-V-I:
@lennix1337, thank you for holding the branch and for the detailed snapshot. The native vs WSL numbers on the same NTFS volume (126 ms vs 1,441 ms cold, 1.2 ms vs 2,113 ms MCP) are a great argument for this port.

**One request so none of that work gets lost:** please push `e795983f` to a *separate* branch on your fork, for example `win32-port-snapshot`, leaving `win32-port` where it is. Our refactor will fold its fixes in commit by commit, with you as the author: UTF-8 and long paths, reparse protection, Job Object cleanup, PATH/PATHEXT, watcher invalidation, the Windows installer. That's better than rebuilding them from the older head. A few pieces also change behaviour on Linux and macOS (CRLF-preserving previews, for example). Those will go out as small separate PRs credited to you, so each gets its own review and gate.

**Landing order:**
1. Hardening and crash fixes, plus the contributor PRs already in review: #231 → #236 → #235 + #233 → #224. #237 has already merged.
2. The POSIX-only `src/infra/os.h` seam lands on `main`. It is output-identical on Linux and macOS.
3. We push the refactor of this PR onto `win32-port`, merged with your snapshot and your commits intact, with the Windows code in `os.h` / `os_win32.cpp`.
4. **Windows testing, where your machine is the key:** the Windows gates you listed, plus a small UTF-8 probe we'll attach. Then a Windows CI leg.

We'll comment here at step 3.

lennix1337:
Hi @joyful-ii-V-I — following your request, I pushed the local evidence snapshot to a separate branch on my fork:

- branch: `lennix1337/ripwire:win32-port-snapshot`
- commit: `e795983f87d31696209a20e6874360ae26b780c6`
- branch URL: https://github.com/lennix1337/ripwire/tree/win32-port-snapshot

I did not open a second PR. PR #44 and its `win32-port` branch remain untouched at the current head.

I’m holding here and waiting for your signal at step 3, after the `src/infra/os.h` refactor is ready on `win32-port`. Once you signal, I’ll run the Windows build, the listed Windows gates, and the UTF-8 probe on the refactored branch and report the results here. No further pushes to `win32-port` until then.

joyful-ii-V-I:
Got it, thank you @lennix1337. `win32-port-snapshot` @ `e795983f` is exactly what we needed. The refactor now builds on your snapshot instead of the older head, merged with your authorship intact, and the pieces that change Linux/macOS behaviour are being split into their own small PRs credited to you. We'll post here when the refactored branch is ready for a Windows run.

joyful-ii-V-I:
@coderabbitai pause

joyful-ii-V-I:
Status for you, @lennix1337, since you've been holding pushes since the 16th.

**The queue ahead of this PR is clear.** Everything we said would land first has landed: #231, #235 + #233, #236, #224 and the rest, plus a run of resolver fixes. `main` is now at `57d713dd`.

**The `os.h` refactor is real and on today's main.** The branch that moves every OS-specific call behind one header (`src/infra/os.h`, `rw::os`, POSIX names and errno behaviour, inline passthroughs so POSIX pays nothing) has been refreshed onto current `main` and pushed. It still owes two proofs before it lands: outputs byte-identical against a `main` build, and the zero-cost check on POSIX. Those are next.

**Then your Windows work goes behind that header**, built on your snapshot `win32-port-snapshot` @ `e795983f`, exactly as described on the 16th: your implementations carry over where your gates prove them, your commits stay, we merge rather than squash, and the pieces that change Linux/macOS behaviour are already split into their own credited branches.

**Timing.** We're token-limited until about 8pm EST today. From then, native Windows to `main` is the priority, not something else.

**What would help, whenever convenient:**
1. Keep holding pushes to `win32-port` until we comment that the refactor commits are up.
2. If you have anything unpushed beyond `e795983f`, say so now and we'll fold it in rather than land around it.
3. Be ready for the Windows run when we post: the build plus the gate list from your PR description. Yours is the only Windows machine in this loop, and that run is the evidence we trust most.

Thanks for the patience, and for the snapshot with its validation list — building on it instead of the older head saved us a lot of guessing.

joyful-ii-V-I:
@lennix1337, it's ready for a Windows run.

**Where things stand**
- **The `os.h` seam is on `main`** (#290). Every OS call goes through `src/infra/os.h` (`rw::os`), and it costs nothing on POSIX: we checked that byte-identical output and codegen hold.
- **The eight Linux/macOS fixes split out of your port landed in #292,** each credited to you, each with a gate that fails before the fix. They are the empty-corpus check, `--index-out` with `--no-ignore`, `wrap`'s JSON escaping, the `sliceBodyLines` bound, the edit-lock contention refusal, `readFilePrefix`'s `ferror` check, the ingest memory release, and the TypeScript runtime extensions.
- **Your Windows implementations are behind that seam** on **`lane/win32-d3-onto-os` @ `f5d44d6b`**. It's built on your `win32-port-snapshot`, and your commits keep your authorship; our adaptations are separate commits. The Windows code lives in `os_win32.cpp` / `os_win32_logic.h` / `cmake/Windows.cmake`. The pure-logic part runs under a new gate on macOS and Linux (29 cases). Since the snapshot we added:
  - long-path (`\\?\`) support;
  - `ERROR_IO_PENDING` → `EWOULDBLOCK`;
  - path normalisation at intake for `--in`, `--doc-drift` and `--exclude`;
  - a disclosed (not silent) `--doctor` row where `which` goes through Git Bash.

**What we can't prove without you:** nobody has built or run this on Windows yet. The only compile check so far was against a hand-written mock SDK. Your run is the evidence we trust most.

**One scope change:** native Windows is now a real target. clang-cl is the primary toolchain, and we'd *also* like plain `cl.exe` (Visual Studio) to build. Item 2b asks you to try it and just report what happens. Any `cl.exe` fixes are our work, not yours.

**The checklist** is below: numbered, with the exact command and expected result for each item. Please:
- keep holding pushes to `win32-port`, since we work from your snapshot;
- report results in whatever shape is easiest (a list of pass/fail by item number is perfect);
- include the first error text for anything that fails.

Items 25–26 are two known gaps to confirm, not fix.

Thank you for staying with this through weeks of a moving `main`.

<details><summary>Windows-run checklist</summary>

# The Windows run — for @lennix1337

Thank you again for the port (#44) and for staying with it through weeks of a moving `main`. Everything below is
what still needs a real Windows machine to prove: this branch (`lane/win32-d3-onto-os` @ `f5d44d6b`)
compiles clean against a hand-written mock SDK and passes every gate that runs on Linux/macOS, but nobody has run
it on Windows yet. This list replaces the scattered notes from the D2/D3 reports and their review — it is the one
list to work from. Items 1–2 are the build. Items 3–24 are gate/behavior checks, each with the exact command and
what a pass looks like. Items 25–26 are two known gaps this round did not fix — please just confirm the repro and
report back rather than trying to fix them (they're small, but scoped for a separate round).

## 1–2. Build

**1. Configure and build with clang-cl (MSVC ABI), from a clean tree.**
```
cmake -S . -B build -G Ninja -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DRIPWIRE_LTO=OFF
cmake --build build --clean-first
```
Expect: `ripwire.exe` and `ripwire_probe.exe` both link, zero errors. Report the warning count (there should be
very few; the codebase is `-Wall -Wextra` clean on macOS/Linux).

**2. Also build: Release + ThinLTO, the ASan flavour, and the test target.**
```
cmake -S . -B build-rel -G Ninja -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_BUILD_TYPE=Release -DRIPWIRE_LTO=ON
cmake --build build-rel

cmake -S . -B build-asan -G Ninja -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DRIPWIRE_ASAN=ON
cmake --build build-asan

cmake -S . -B build-test -G Ninja -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DRIPWIRE_TESTS=ON
cmake --build build-test
ctest --test-dir build-test
```
Expect: all four build, and `ctest` passes every target including `ripwire.oswin32logic` (the pure-logic test
that already passes on Linux/macOS — this just proves the same source compiles and passes under clang-cl too).

**2b. Also build with plain `cl.exe` (MSVC) — new owner requirement (2026-09-19).** clang-cl stays the primary, tested
toolchain, but the project now also wants a plain Visual Studio build to work. Please try:
```
cmake -S . -B build-msvc -G "Visual Studio 17 2022" -A x64
cmake --build build-msvc --config RelWithDebInfo
```
Expect: we don't know yet. Please report the first errors (or success) as-is; any cl.exe-specific fixes will be our
work in a follow-up, not yours.

## 3–13. The snapshot harness, run against this branch's binary

This branch carries none of `win32-port-snapshot`'s Windows harness (that's D4, still to come). Until D4 lands,
run these from a `win32-port-snapshot` checkout with the gate scripts' binary set to **this branch's**
`build/ripwire.exe`. Every gate takes the binary as its first argument, for example (from Git Bash, in the snapshot
checkout):
```
bash test/mcpeditracecheck.sh /c/path/to/this-branch/build/ripwire.exe
```

3. `manifestcheck` — pass.
4. `pargatescheck` — pass.
5. `portablebuildcheck` — pass.
6. `sidecarsymlinkcheck` — pass, 89/89.
7. `runtracecheck` — pass (see item 20 for the new adversarial cases to add to it).
8. `mcpeditracecheck` — pass, including the race trials. **Also check the lock-contention code by hand**, because
   `main` now refuses an edit only when the lock is proven held by someone else, and on Windows that proof depends
   on `LockFileEx` failures (`ERROR_LOCK_VIOLATION`, `ERROR_IO_PENDING`) mapping to `EWOULDBLOCK`. Hold the file's
   edit lock from a second process (e.g. a small PowerShell script that opens `<cache>/…/edit.lock` and calls
   `LockFileEx` exclusive, or a second `ripwire.exe --mcp` mid-edit), then send a `replace_symbol_body` over MCP.
   Expect: the edit **refuses** with JSON-RPC error `-32603` and a message containing "edit lock unavailable", and
   the target file is byte-identical afterwards. Release the lock and resend: expect it to apply. If it applies
   while the lock is held, the error mapping is wrong; please include the Win32 error code you see.
9. `mcpstalecheck` — pass.
10. `mcpwatchercheck` — pass.
11. `qsnapprefetchcheck` — pass.
12. `--version`, `--help`, `--doctor` — each runs and prints something sane. For `--doctor`'s binary-path row
    specifically, see item 23 below — don't expect a clean `ok="1"` there, that's the known, disclosed gap.
13. Determinism (`ripwire.exe test\fixture` twice, `fc` or `cmp` the two outputs) and well-formedness
    (`xmlcheck.py --noout` on the output) — both pass.

## 14–19. From the 09-16 review comment

14. `cacheisolationcheck` — pass.
15. `githardencheck` — pass.
16. `opencodewrapcheck` — pass.
17. `docscommandscheck` — pass.
18. `editpreviewcheck` — pass.
19. `editplanrollbackmsgcheck`, `decltodefcheck`, `clonebandcheck`, `clonelexcheck`, `columnarcommacheck`,
    `connectcorecheck`, `dynmapsimdcheck`, `emitescapecheck`, `releaseinstallcheck`, `childwalkscalecheck` — each
    pass (these are the ones the 09-16 review flagged as never run on Windows).

## 20. The adversarial `--run-trace` / MSYS quoting cases (LOW-5 evidence, see item 25)

Run `--run-trace` with a command containing `\"`, a trailing `\`, a `%`, and a `*`. Check the child process
receives each argument byte-for-byte unchanged, and that `*` is never glob-expanded. `runtracecheck` may already
carry some of these vectors — if it does, just confirm they pass; if not, add them. This is the evidence LOW-5
(item 25) needs, not a fix — MSYS bash re-parsing `spawn_sh`'s command line is a separate, harder problem, not
addressed this round.

## 21. CRLF through the one `fdopen` read path

Confirm CRLF bytes pass through `os::fdopen(fd, "r")` (`pathguard.h:795`) unchanged: run `--edit-check` (or the
sidecar path) on a CRLF-line-ending file and confirm `bytes=` matches the POSIX run on the same file, plus one
`--grep` hit on that file landing at the correct column.

## 22. Long paths — MED-1, now fixed this round (re-test the ENOENT case)

**This is the one item whose expected result CHANGED from earlier reports.** Earlier notes said "expect ENOENT" —
that was describing the bug. `NativePath` now applies the `\\?\` extended-length prefix past ~260 characters, so:

- With `LongPathsEnabled=0` (the Windows default): create and read a file whose full path is 261+ characters.
  **Expect success** — `ripwire` should crawl/read it with no `ENOENT`. (Before this fix round, this failed.)
- With `LongPathsEnabled=1`: same file, still expect success (this path was already fine before the fix; it
  should stay fine).
- A UNC path (`\\server\share\...`) past the same length: expect success too — the fix prefixes UNC paths as
  `\\?\UNC\server\share\...`.
- A *relative* path that happens to be long: this fix does not and cannot help it (relative paths aren't
  prefixable) — it still needs `LongPathsEnabled=1`, unchanged from before.

If any of the "expect success" cases still fails, that's a real regression in this round — please report it with
the exact path length and whether it's a drive path or UNC.

- **The root spelling matters too:** run the drive-path case twice, once with an absolute root
  (`ripwire.exe C:\deep\...\root`) and once from inside the directory with a relative root (`ripwire.exe .`). The
  relative-root form cannot be prefixed, so on `LongPathsEnabled=0` it may still fail past 260 characters. Please
  report both results; the relative case is a known limit, not a regression.

## 23. `--doctor`'s binary-path row — now disclosed, not silently wrong

Run `--doctor` from a normal PATH install. The `binary-path` row should now carry `degraded="1"
degrade_reason="win32-which-spelling"` (Git Bash's `which` never prints `.exe`, so `same_file`/`same_bytes` can
legitimately disagree with a correct install). Confirm the attribute is present and that a correct install is no
longer silently reported as a clean mismatch without that context. This is a disclosure, not a real fix — the
real fix (a native `os::which` PATH/PATHEXT search) is future work (D4).

## 24. The rest of D3's own new-on-Windows surface (never run before this branch)

24a. `openat` + `O_DIRECTORY`: `--edit-plan` (`editplan.h`'s `parseEdit` → `readWholeBeneathNoFollow`;
     `editplancheck`, `editplanpayloadconfinecheck`), including a junction as an intermediate or final path
     component (must be refused).
24b. `pthread_create`/`pthread_join` with a sized stack: `--grep` regex, `--scan-skill`, `--arch`
     (`grepcheck`, `regexguardcheck`, `skillscanreadcheck`).
24c. `os::path_is_root`: launch MCP with cwd `C:\` — must refuse the assumed root.
24d. `os::path_is_absolute`: a linked worktree whose `.git` says `gitdir: C:/…` (`mcpstalecheck`/`mcpwatchercheck`).
24e. `os::program_path`: confirm no `\` appears in any `p=` attribute of a map.
24f. Memory-stream publishing through `rw::MemoryStream`: every `--for`/MCP verb returns non-empty bodies.
24g. `normalize_path_arg` for roots given as `C:\x`, `/c/x`, and `c:/x` — all three should resolve to the same
     place. **New this round:** also try `--in=src\quality.h`, `--doc-drift=C:\docs\README.md`, and
     `--exclude=C:\vendor\` (backslash spellings) — these three flags are newly normalized at intake and were not
     covered by lennix1337's original list.
24h. The UTF-8 active-code-page probe: does the UCRT's narrow `std::fopen` honor the manifest's
     `activeCodePage=UTF-8` for a non-ASCII (CJK/emoji) path? Check a non-ASCII root and file name through argv,
     MCP `path`, the crawl, and the four narrow `std::ifstream` sites (`--arch=`, `--eval-*`, `--doctor`,
     `binstale`).
24i. `TMP`/`TEMP`/`TMPDIR` set to a path over 320 characters: map and `--run-trace` should both still exit 0 (the
     child gets a short temporary directory instead, per `isLongTemporaryEntry`).
24j. `--listen`: restart on the same port within `TIME_WAIT` should succeed; a second process binding the same
     port should fail (`SO_EXCLUSIVEADDRUSE`).
24k. A linked worktree whose `.git` says `gitdir: C:/…` (already 24d) — also confirm `gitdir: /c/…` never appears
     from Git Bash; skip that sub-case if it doesn't.

## 25–26. Known gaps — please confirm and report, do not try to fix

**25. LOW-5 — MSYS re-parses `spawn_sh`'s command line.** `spawn_sh` quotes its command for `CreateProcessW` by
MSVCRT rules, but the process that reads it is MSYS bash, which re-parses `GetCommandLineW` with its own rules.
The two disagree on `\"` sequences and a trailing backslash. Repro: run `--run-trace` with a command argument
ending in a single backslash right before a closing quote (e.g. an argument that is itself `C:\some\path\`), and
with an argument containing an escaped quote (`a\"b`). Compare what the child process actually received (echo it
back, or log argv in a tiny test script) against what was passed. Report whether they diverge and how. `popen` is
not affected (it goes through a script file, not the command line).

**26. LOW-6 — `openat` fails on a volume without a drive letter.** `openat`'s `GetFinalPathNameByHandleW(...,
VOLUME_NAME_DOS)` call fails when the directory is on a volume mounted without a drive letter (a mounted folder,
or a VHD mounted into an NTFS folder rather than a drive letter). Repro: mount a folder or VHD without a drive
letter, point `--index-out=` or a crawl root's sidecar writes at a directory inside it, and confirm the no-follow
descriptor chain (`pathguard.h`) refuses cleanly (no crash, a clear error) rather than doing something silently
wrong. This is expected to fail today — we want to confirm it fails *safely*, not fix the failure.

---

Everything above either already passes on Linux/macOS (the portable logic in `os_win32_logic.h`, tested by
`test/oswin32logiccheck.sh` — 29 cases, 4,480,667 assertions) or is compile-checked against the mock SDK. What's
missing is exactly what's in this list: a real Windows machine actually running it. Thank you for closing that
gap — and please push back on anything above that looks wrong from where you're sitting; you know this code
better than any of us reviewing it without a Windows box.

</details>

lennix1337:
@joyful-ii-V-I I re-read the checklist and ran the requested validation against the exact Windows ref `lane/win32-d3-onto-os` at `f5d44d6b`.

Environment used: VS 2022 Build Tools 17.12.4, MSVC 14.42.34433, Windows SDK 10.0.22621.0, clang-cl 22.1.8, Ninja 1.12.1, CMake 4.2.3.

Results:

- 1. clang-cl debug/Ninja: FAIL. First error: `src/ingest.cpp(263,23): error: cannot initialize a parameter of type 'const char *' with an rvalue of type 'const value_type *' (aka 'const wchar_t *')`. Warning diagnostics: 1.
- 2a. Release + ThinLTO: FAIL. Same `std::filesystem::path` `wchar_t*` to narrow `char*` mismatch in `src/ingest.cpp`/`src/wrap.h`. Warning diagnostics: 1.
- 2a. ASan: FAIL. First error: `clang-cl: error: invalid argument '-MDd' not allowed with '-fsanitize=address'`.
- 2a. Tests + ctest: FAIL. First link error: missing `Diagnostics::ConsoleLog::handleAssert(...)` while linking `ripwire_test_oswin32logic.exe`; ctest was not reached.
- 2b. Plain cl.exe / VS 2022: FAIL. First error: `cl : command line error D8021: invalid numeric argument '/Werror=switch'`.

No `ripwire.exe` or `ripwire_probe.exe` was produced from the unmodified ref. Therefore:

- 3–13 (manifest/pargates/portable/sidecars/runtrace/MCP/version/help/doctor/determinism/XML): BLOCKED/NOT RUN.
- 14–19 (cache isolation, git hardening, opencode, docs, edit preview/rollback, decl-to-def, clone, comma, connect, SIMD, escape, release install, child walk): BLOCKED/NOT RUN.
- 20–24 (adversarial run-trace, CRLF, long paths, doctor disclosure and D3 Windows surface): BLOCKED/NOT RUN.
- 25 / LOW-5: not reproduced; not fixed.
- 26 / LOW-6: not reproduced; not fixed.

So the exact-ref run is currently a build/portability blocker, not a passing Windows validation. I did not push or modify `win32-port`.

If you want, I can take all non-25/26 fixes locally in a separate worktree, rebase the PR onto the current `main`, and stabilize the PR for a second Windows validation pass. I would keep the original `f5d44d6b` result separate and label the second pass as post-fix validation. For the cl.exe-specific follow-up, I will follow whatever scope you prefer from item 2b.

joyful-ii-V-I:
**Update: both MSVC front ends now build ripwire, and it ships in 0.6.2.**

@lennix1337 — your validation run is what made this possible. It established that the port did not compile at all, which turned a vague "Windows support" into four concrete failures: a narrow/wide `std::filesystem::path` mismatch, the sanitizer flavour's runtime-library choice, a test target missing a source, and clang-style flags handed to `cl.exe`. Each is fixed by one mechanism rather than a special case.

Where it stands now:
- **clang-cl and `cl.exe` both build and run**, on CMake plus Ninja and nothing else — no WSL2, no MSYS, no prebuilt binary. Both CI legs gate.
- Clearing `cl.exe` took four more walls past the flag dialect: `__BASE_FILE__`, `__PRETTY_FUNCTION__`, `__int128`/`__BYTE_ORDER__` (removed rather than ported — the wide type bought nothing, and `std::endian` is a standard answer), and `/bigobj`, attached to both MSVC-ABI front ends because the object-format limit belongs to the format.
- POSIX is unaffected: byte-identical output across two corpora, both legend postures, `--json` and MCP.

**What a user should not conclude, stated in README, CHANGELOG and CONTRIBUTING:** the 647-gate suite does **not** run on Windows, and ASan is compiled there but never executed. It is a build proof plus a smoke test on a clean NTFS volume with short ASCII paths — long paths, UNC shares, junctions, the edit lock, CRLF and non-ASCII paths are untested rather than merely unlisted.

**That harness is the high-priority 0.6.3 item here**, along with running a sanitizer on Windows. Thank you — this was a contributor report that changed a release.

PR #45, opened 2026-09-07
fix: resolve named JavaScript and TypeScript import aliases

PollyBot13:
## Summary
Fix named ES import aliases resolving to unrelated same-named functions in the call graph.

```ts
// owner.ts
export function saveMemory() { return 1; }
// consumer.ts
import { saveMemory as persistMemory } from './owner.js';
export function caller() { return persistMemory(); }
// decoy.ts
export function persistMemory() { return 2; }
```

Before this change, `--callers=saveMemory` finds no caller, while `--callers=persistMemory` attributes `caller` to the unrelated decoy. The graph reports neither ambiguity nor an unresolved call.

### Reproducer result
`--callers` (same three-file fixture, cold parse):

```text
                 before           after
saveMemory       []               [caller]
persistMemory    [caller]         []
```
The direct-import control continues to resolve.

## Change
- Capture named import local/export/module identities and direct exported declarations.
- Resolve against a uniquely identified module using existing workspace-alias context and its actual exported declaration; handle runtime `.js`/`.mjs`/`.cjs` spellings of source files conservatively.
- Preserve lexical shadowing, including parameters, closures, loops, function/class expressions and default-parameter scope. Unmodelled targets remain unresolved instead of falling back to a global nickname match.
- Persist the new binding fact and invalidate old extraction caches.

The regression is part of the existing TypeScript import gate. It exercises TS/TSX/MTS/CTS and JS/JSX/MJS/CJS, wrong-target decoys, scoped shadows, non-exported/type-only/missing targets, ambiguous source/runtime files, input mutation, and cold/warm equivalence.

## Boundaries
This changes call-graph relationships, not ranking policy or general TypeScript semantics. Re-export/barrel traversal, default/namespace imports and new compiler/tsconfig integration are not added (existing cross-root aliases are preserved). `--uses` still matches written reference names, so it can omit an alias site when queried by its target name; this PR does not redesign that index.

The source/runtime file choice is deliberately conservative: competing candidates remain unresolved. Shadowed local calls may also remain unresolved; this change does not infer local callback values.

## Scope and recovery
Binding extraction owns import/export/shadow identity; graph resolution consumes it before the global name fallback. This is one correctness change: the extractor, cache schema and graph consumer must move together, and shadow/ambiguity checks prevent replacing one false edge with another. No dependencies, installer hooks, ranking defaults or deployed runtime change. Revert both commits to restore prior behavior; extraction caches self-invalidate.

AI-assisted implementation, with independent adversarial source review and executable CLI probes.

## Verification
- macOS ARM64: plain, Release and `-DRIPWIRE_ASAN=ON` builds pass (vendored dependencies, disconnected configure).
- `bash test/lib/jsimportalias.sh <binary>` passes with all three builds; the new regression failed before the implementation.
- `LSAN_OPTIONS=suppressions=lsan_suppressions.txt ./asan/ripwire .`: exit 0, no sanitizer findings (an old extraction cache was correctly rejected/reparsed).
- Three default-map runs, cold then warm: byte-identical, valid XML.
- `scripts/formatcheck.sh`: passes with pinned clang-format 22.1.8.
- Independent source review and executable cross-root/shadow/ambiguity probes: no remaining must-fix finding.
- Full gate suite: `python3 test/pargates.py . ./build/ripwire -j 6` — **561 gates, 558 pass, 3 failures**, classified below; no alias-specific or unexplained failure.
  - `legendcoveragecheck`: missing `scan-skills` legend definitions for `f@rule` / `f@sev`; reproduced identically using unchanged upstream `6488f6f4`.
  - `argvdiffcheck` and `selectorchaincheck` were explicitly activated against the pre-change binary. Their byte-identical-old-output contract is inappropriate for changed extraction/cache behavior. Rerunning their standard no-baseline configurations yields the documented argv skip and selector **ALL PASS** (only its optional identity arm skips).

Test environment: installer gates run with `CODEX_HOME`/`AGENTS_HOME` unset; the existing payload-budget gate requires a short checkout path on this machine (the long-path failure also reproduces on unchanged upstream). The optional `argvdiffcheck` checks byte-identical parser refactors and is not applicable to this change: deliberately enabling it against the old binary produces differences including version/cache diagnostics and trace output; its ordinary CI invocation skips without `RIPWIRE_BASE`.


joyful-ii-V-I:
Merged in `35f32aa1`. Thank you — this fixed a *silent wrong answer*, which is the worst kind.

Before your change, `import { saveMemory as persistMemory }` made ripwire emit an edge to a
same-named decoy in another file while reporting `graph_ambiguous="0" graph_unresolved="0"`. A
confident lie. That's a bug this project cares about more than a missing edge, and nothing in the
suite could see it.

Your shadowing model is unusually careful — including that a body `var` does *not* shadow the
import inside a default-parameter expression, because default params get their own scope. The gate
tests exactly that. And the mutation controls are real: I ran your gate against the pre-fix binary
and watched 25 assertions fail, which is stronger evidence than most gates here carry.

Three things I changed on landing:

- **External imports were landing in `unresolved=` instead of `external=`.** `import { execFileSync }
  from 'node:child_process'` took `unresolved` from 0 to 1. The shipped legend assigns that case to
  `external=`, and `graph.h`'s "Honesty lever #2" calls it out by name as the exact failure it
  exists to kill. Now routed through `vetoExternal()`, with a four-way outcome taxonomy.
- **`export { x }` and barrel re-exports lost edges that previously resolved correctly** — before=1,
  after=0. `export_clause` specifiers are now captured, and an unlisted name degrades to the name
  ladder rather than refusing, so barrels keep what they had.
- **Disclosure caught up with the new mechanism**: a new `PinMech::Import` (not reusing
  `PinMech::Binding`, whose membership is scored as the FFI population) and `prov="import"` on
  import-resolved edges, since the legend defines absent `prov` as "uniquely-resolved-name-based",
  which these aren't.

Net on webpack: your change lost 64 edge rows, the fixes restore 45 and add 5 the ladder never
found. The 19 still refused were audited one by one — all name-sprays where the old edge was wrong.

One note for next time: your run reported 3 failures; all three are clean here, including
`legendcoveragecheck` on unchanged upstream. Likely `CODEX_HOME`/`AGENTS_HOME` skills-staging state.

PR #46, opened 2026-09-07
feat(skills): add Hermes-format skill under skills/hermes/, extend vetting sweeps to cover it

ashutoshsinghpr7:
## Summary

Adds a **Hermes-format skill** for ripwire — a distilled Hermes-native counterpart to the `ripwire-orient` family — and extends the two vetting sweeps so a namespaced skill like it is scanned like the flat set, not silently skipped.

No C++ or behavior change. Diff: 2 new files + 2 sweep-glob lines.

## Why a Hermes skill

[Hermes Agent](https://hermes-agent.nousresearch.com) is a shell-capable coding agent ("if your agent can run shell commands, it is set up"): its skills are agentskills.io-style `SKILL.md` files under `~/.hermes/skills/`. The added skill teaches a Hermes agent **when** to reach for ripwire instead of grep-and-read: orienting in an unfamiliar repo, the `--for` task lens, `--callers` blast radius, `--test-gate` before committing, resuming work after context compaction. Its honesty-signals section mirrors ripwire's own disclosure model (`confidence=`/`margin_pct=`, `--skipped`, `amb=`) so an agent reads a map the way the tool wants it read.

## Why `skills/hermes/` and not `skills/ripwire-*`

`skills/install.sh` symlinks **every** `skills/ripwire-*` directory into Claude Code / Codex skill roots. A Hermes-format file must not be auto-installed into those agents — wrong frontmatter dialect for them, and it would pollute the routed `ripwire-*` set. The `hermes/` namespace is invisible to that glob (verified: `for d in "$src"/ripwire-*/`), so Claude/Codex installs are untouched.

## Sweep coverage (the second half of this PR)

`skillscan.sh` check 9 and `deckcheck.sh` globbed only `skills/*/SKILL.md`, so `skills/hermes/ripwire-repo-map/SKILL.md` would have **silently escaped both gates** — the green-while-inert shape `deckcheck.sh`'s own header exists to prevent. Both now include the depth-3 pattern `skills/*/*/SKILL.md`.

Verified locally against a plain cmake build (macOS arm64):
- `bash test/skillscan.sh` → **ALL PASS**, `19/19 shipped skills scan clean` — the new file is swept and scans attack-clean.
- `bash test/deckcheck.sh` → **ALL PASS** — 55 prose sources, 272 distinct `--flag` tokens, 0 fabricated, 0 bad values.

## Notes for review

- The frontmatter `description` is intentionally one line: Hermes indexes skill descriptions into every session's system prompt, so the trigger matrix lives in the body. `allowed-tools` is omitted for the same reason — Hermes validates against a fixed frontmatter key set.
- Not added to the release-tarball packaging (the release lane copies `ripwire-*` only); Hermes users install from the repo as documented in `skills/hermes/README.md`. Happy to include it in the tarball if you'd prefer.

### DIFF
--- a/skills/hermes/README.md
+++ b/skills/hermes/README.md
@@ -1,0 +1,13 @@
+# Hermes skills
+
+Hermes-format counterpart to the `ripwire-*` skill family. Kept in its own namespace (not
+`skills/ripwire-*`) so `skills/install.sh` — which symlinks every `ripwire-*` directory into
+Claude Code / Codex skill roots — never installs it into those agents.
+
+Install into a Hermes profile:
+
+```bash
+cp -r skills/hermes/ripwire-repo-map ~/.hermes/skills/software-development/
+```
+
+Requires the `ripwire` binary on PATH (upstream release tarball, see repo README).
--- a/skills/hermes/ripwire-repo-map/SKILL.md
+++ b/skills/hermes/ripwire-repo-map/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: ripwire-repo-map
+description: "Use ripwire CLI to map a repo before reading it."
+version: 1.0.0
+author: Ashutosh Singh
+license: Apache-2.0
+platforms: [linux, macos]
+metadata:
+  hermes:
+    tags: [ripwire, repo-map, call-graph, context-engineering, code-navigation, static-analysis]
+    related_skills: [systematic-debugging, codebase-inspection]
+prerequisites:
+  commands: [ripwire]
+---
+
+# ripwire — deterministic repo map before reading code
+
+`ripwire` (v0.4.x at authoring time; the release binary must be on PATH) is a zero-dependency C++23
+CLI that builds a ranked, deterministic call graph of any repo in ~1s (parses once, then auto-caches
+warm). It answers "what to touch, what it breaks, which tests to run" for a fraction of the tokens of
+grep + whole-file reads. No API key, no daemon, no index server, offline. Invoke it via the terminal
+tool.
+
+## When to reach for it (instead of blind grep / whole-file reads)
+
+- Landing cold in a repo or subsystem — "what's here, how is it organized, what matters"
+- "Where is the code for X" — a task, feature, or concept in your own words
+- A change is proposed — what breaks, who calls it, which tests must run
+- Resuming work after compaction / a new session on in-flight work
+- Reviewing code you did not write (fresh eyes), or pre-commit change vetting
+
+Do NOT use for trivial single-file lookups (plain `rg`/search_files wins) or broad common-word
+questions — ripwire shines on specific technical asks.
+
+## Core verbs (all take a repo dir; `--help` is authoritative)
+
+```bash
+ripwire <dir> --report            # architecture summary: modules, god-files, cycles, top symbols
+ripwire <dir> --for="<task in your own words>"   # ranked task lens: what to touch first, with 1-hop edges
+ripwire <dir> --callers=someFunc  # who calls it (blast radius)
+ripwire <dir> --test-gate         # which tests must run before committing
+ripwire <dir> --tree              # file-by-file: top symbols per file
+ripwire <dir> --hotspots          # files ranked by churn × complexity
+ripwire <dir> --communities       # cohesive modules (Louvain) + bridges — where a new feature belongs
+ripwire <dir> --situ              # what the working tree already changed + its blast radius (resuming work)
+ripwire <dir> --notes             # pinned gotchas from past sessions (field notes)
+ripwire <git-url>                 # orient in a remote repo (shallow-clones into a cached temp dir; --refetch refreshes)
+```
+
+Read the map, then open only the files it surfaces (god-files and hotspots first). For a body of one
+ranked symbol: `ripwire <dir> --for="..." --expand=path:name` (paste `p=`/`n=` off the row).
+
+## Honesty signals — read them before trusting a result
+
+- The `--for` lens reports `confidence=`/`margin_pct=` on its result root and the output carries
+  `est_tokens=` — `confidence="low"` or a flat ranking means treat the set as a starting point, not an answer.
+- `--for` bundles: `bodies="0"` = no full bodies shipped (`reason="compact-route"`; `--auto-bodies`
+  restores them. A `reason="budget"` means the caller's token budget cut them — raise the budget instead).
+- `--skipped` lists files the index dropped (oversize, or `.gitignore`d — default ignored; `--no-ignore`
+  restores the full walk). A "not found" is only trustworthy after checking this.
+- `amb="K"` on a symbol means K of its outbound calls matched several possible definitions — the
+  resolver split the weight instead of choosing one. Read the source if the target matters.
+  (`lpin=` is a separate, disclosed guess the resolver made; same remedy.)
+
+## Pitfalls
+
+1. **First call on a tree parses** (~1s at 1,500+ files; measured ≈1.2s wall for a 2,355-file /
+   77k-edge checkout on Apple M2). Later calls are warm and near-instant — chain verbs freely.
+2. `--top-k` is NOT read by `--for`'s own bundle (it shapes the default map and `--query`; it still
+   applies to `--format=candidates` with `--for`).
+3. Markdown/docs-only or shell-only repos yield few call-graph edges — `--report` is still useful, edges are not.
+4. The tool indexes signatures, not necessarily full bodies, by default on conceptual queries — `--expand`
+   or `--auto-bodies` when you need the body.
+5. Cache/notes live in the repo tree (`.ripwire_notes` for `--note-add`); the notes file is meant to be committed.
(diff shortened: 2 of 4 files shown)


joyful-ii-V-I:
Thank you for this, and apologies for the silence — you deserved a status note before now rather than an open PR with nothing on it.

Where it stands honestly: this is queued, not ignored, and it is not blocked on anything wrong with the PR. Skill-activation targets touch how the installer writes configuration outside its own prefix, so the order we add them in is a decision I want to make deliberately rather than in the order PRs arrive. There is also a sequencing question on our side that isn't mine alone to answer, and I'd rather tell you that than invent a technical reason.

I'm not going to give you a date I can't keep. What I can say is that nothing here is a rejection, the PR stays open, and when the ordering question is settled this gets a real review — the same one #57 is getting today: full gate battery, sanitizers, and an adversarial pass rather than just the happy path.

If you'd rather not wait on that, say so and I'll prioritise differently. Leaving a contributor hanging is the failure mode I'm most trying to avoid here.

ashutoshsinghpr7:
Thanks for the status note — honestly appreciated, and no apology needed. This kind of transparency is exactly what keeps contributors around.

To make the sequencing decision easier: nothing in this PR requires installer integration. The skill is namespaced under skills/hermes/ specifically so the ripwire-* glob in skills/install.sh can never touch it, and the README documents a plain cp -r install into ~/.hermes/skills/. So if the open question is the order in which agent-activation targets get added, this can land without committing you to any Hermes-aware installer change — or hold until that's settled, either works for us. We're not blocked: the skill is already live in my own Hermes setup, so this PR is purely about co-location and discoverability for other Hermes users.

And no rush on prioritising differently — the PR can sit in the queue until the sequencing question is resolved.

joyful-ii-V-I:
One factual question, and it changes the shape of what I land rather than anything about your PR.

I am refactoring the agent wiring in `src/wrap.h` into a table — one row per agent, because agent identity is currently branched on in ~30 places across 7 files and your PR and #51 both had to touch most of them. Adding an agent should be a row, not a branch, and that makes both of yours smaller.

One column is whether the agent can shell out. Where it can, we recommend the CLI over MCP, because the CLI costs zero context until it is invoked while a registered MCP server's verb schemas sit in the model's context every turn whether a verb is called or not (measured, `docs/EVALS.md` §5). That is why the opencode recipe leads with the shell path.

**Does Hermes have a shell/terminal/exec tool available to the agent at run time?**

Your SKILL.md says "Invoke it via the terminal tool", which suggests yes — but that is your text, and I would rather ask than infer a platform fact from a sentence I am about to build a recommendation on. If yes, Hermes joins codex, opencode and openclaw on a shared CLI-first recipe rather than getting its own. If no, it stays MCP-wired and that is fine too.

Also useful if you know it: does Hermes ignore unknown frontmatter keys? Your skill carries `version`, `author`, `license`, `platforms`, `metadata.hermes.*` and `prerequisites`, where ripwire's existing eighteen skills carry only `name:` and `description:`. If the extras are optional, the eighteen work in Hermes unchanged, and I would rather install those than maintain a nineteenth hand-written copy that will drift from them.

No rush, and nothing here is a criticism of the PR — the hook-slot detail #51 found is exactly the kind of platform fact this table needs.

joyful-ii-V-I:
Merged — thank you, this is a well-judged contribution.

Three calls here were the right ones, and none of them are obvious:

**The namespace decision.** Putting this under `skills/hermes/` rather than `skills/ripwire-*` keeps `skills/install.sh` from ever symlinking a Hermes-format skill into a Claude Code or Codex skill root. Verified rather than assumed: a `--codex` install into a throwaway HOME deploys **0** Hermes directories. You reasoned about a glob's blast radius before writing the file, which is the step most people skip.

**Extending the sweeps in the same PR.** A namespaced skill that the vetting sweeps silently skipped would have been strictly worse than no skill — it would ship unvetted while *looking* covered. Landing the `skills/*/*/SKILL.md` globs alongside the file is exactly the house rule.

**The prose is honest.** "Do NOT use for trivial single-file lookups (plain `rg`/`search_files` wins)" and "`--help` is authoritative" are the two sentences that make a skill trustworthy rather than promotional. I checked all 19 flags the skill names against `--help` expecting to find a fabrication — every one is real.

---

**One thing I fixed on the way in, so you know it happened.** The full gate battery caught `skillevalcheck.sh` failing: it enumerates skills with `find -maxdepth 1 -type d`, so it counted the *namespace* `hermes` as a skill, found it had zero labelled rows in the routing corpus, and went red. Third of three sweeps — you updated the two that were greppable for `skills/*`; this one's glob didn't look like the others.

Landed as a separate commit (`5e96a6a5`) rather than folded into yours: a skill is now defined as *a directory that contains a `SKILL.md`*, which is correct for any future namespace without another edit there.

**One integration note, not a change request.** #51 adds a `--hermes` installer mode deploying `skills/ripwire-*` into `${HERMES_HOME:-~/.hermes}/skills`. That glob does not match `skills/hermes/`, so once both land, `--hermes` installs the *Claude/Codex-format* skills into a Hermes profile while the skill purpose-built **for** Hermes still needs the manual `cp` from your README. A seam between two PRs rather than a defect in either — better resolved when #51 lands. Views very welcome; you and @AnkitArya know Hermes' skill loading far better than I do.

Worth saying plainly: **nobody maintaining this repo runs Hermes.** Everything we ship for it is verified against documentation, not against a running agent. If you're ever able to run the two prompts from that skill against real Hermes and tell us what actually happened — *especially* if it goes badly — that is the single most useful thing anyone could contribute to this corner of the repo right now.

ashutoshsinghpr7:
It's merged, but for the question's and live run on hermes:

**Q1** — Shell/terminal: yes, natively. Hermes agents have a first-class terminal tool; the agent that ran this whole exchange (fork, gates, sweep edits, this test) executed every command through it. MCP is also supported natively (hermes mcp), so Hermes can be MCP-wired — but CLI-first is the better fit for the same reason your table column exists: a skill is just instructions the agent loads on demand, and ripwire costs zero context until the agent actually runs it via the shell. Hermes belongs in the codex/opencode/openclaw column.

**Q2** — Unknown frontmatter keys: yes, ignored; name + description is the full legal minimum. I didn't infer this — I tested it. Fresh isolated HERMES_HOME, every ripwire-* skill on current main copied verbatim (their exact frontmatter — name, folded long description, allowed-tools):

`17/17 loaded, all enabled, zero warnings or parse errors. allowed-tools is silently tolerated.`

So the flat set works in Hermes unchanged (17 on current main — the v0.4.0 tarball shipped 18; ripwire-efficient has since been consolidated), and installing them beats a hand-written nineteenth on drift — agreed. Three findings from the test that a docs-only pass wouldn't have surfaced, for the --hermes work in #51:

- **Description length is a routing constraint, not just style.** Hermes lists every installed skill's description in the agent's context each session, truncated at ~57 chars + …. ripwire-orient's is 325 chars — it survives because it opens with the trigger ("Landing COLD in an unfamiliar repo…"). Skills whose first 57 chars are boilerplate would route poorly. This is also why the Hermes skill's description is deliberately one line.

- **Flat companion files are invisible to a Hermes agent.** Four of the seventeen ship companions beside SKILL.md (fresh-eyes/lint-rules.md, mcp/mcp-reference.md, quality-bar/quality-metrics.md, orient/compress-ladder.md + map-before-you-read.md) — and all four bodies reference theirs by name. Hermes auto-exposes only references/, templates/, scripts/ subdirs as loadable files. Bodies inline enough to work standalone, but --hermes would serve better mapping companions into references/.

- **Several bodies carry client-isms.** ripwire-mcp leads with ripwire wrap claude → claude mcp add + mcp__ripwire__ tool names; router/navigate describe install.sh symlinking into ~/.claude/skills. Harmless dead text in Hermes — but it's exactly the noise that argues for a Hermes-shaped entry point on top of the 17: in Claude, ripwire-router does the cross-skill routing; in Hermes, routing happens in the skill index, and what the agent sees matters more than what's installed.

**On the seam:** once #51 lands, I'd suggest having its --hermes mode install the distilled skill alongside the 17 (all proven loadable above). The installer's ripwire-* glob deliberately skips skills/hermes/, so the mode would need to name it explicitly — worth it because the distilled skill is the client-agnostic entry point (no claude mcp add/hook text, description sized for Hermes' 57-char index) that routes into the family for depth. Drift risk is real but bounded: it defers to --help and the family skills, so its maintenance surface is small. If you'd rather not carry a nineteenth at all, happy to retire it in favor of the 17 — your call.

**And the prompts challenge — done, live.** A real Hermes agent ran both core invocations from the skill against a real repo ([Vibe-Trading](https://github.com/HKUDS/Vibe-Trading) 77k edges):

-  `--report` → full architecture map (modules, god-files, cycles, top symbols) in 1.17s wall.

-  `--for=`"place an order through a broker connector and track fills" → ranked place_order 1, TradingPlaceOrderTool 2, execute_live_order 3, est_tokens=3644 for the whole bundle. The lens itself labels that head confidence=low (flat ranking — the tool's own honesty signal for "starting point, not answer").

One wobble worth reporting rather than hiding: a shell+docs-only repo mapped with zero call-graph edges — disclosed in the output itself (0 edges, isolates listed), so no agent can mistake the map for a connected codebase.

PR #47, opened 2026-09-07
fix(ruby): definition scope, and setter methods

andriytyurnikov:
Two Ruby fixes, one commit each.

**Scope.** A Ruby `def` never had a scope (only C++, Python, Rust did). So: no `id=` on any Ruby row, no `Scope::name` selector, and same-named methods in different classes of one file folded into one `overloads=N` row (ActiveSupport 7.2: `as_json` 26, `initialize` 13). Now the enclosing class/module is recorded. Three resolver changes ride along so no edge is lost: Ruby receivers are classified (`self.m` / `x.m`), a bare `m(args)` inside a class pins to that class via Rule 1, and the locality tie-break no longer lets the caller's own def win (it did, then got dropped as a self-loop).

**Setters.** `def name=(v)` was never indexed, and `obj.name = v` edged to the getter. Both sides now spell `name=`. Compound `obj.x += 1` keeps the getter edge; stated floor.

ActiveSupport 7.2.3.2 `lib/`, `--top-k=100000`: overloads rows 111 → 21, `id=` rows 0 → 2468, ambiguous 451 → 417, symbols +27 setters. Byte-identical across runs, xmllint clean, ASan clean. `kParserVer` 77 → 79.

Gates: `test/rubyscopecheck.sh`, `test/rubysettercheck.sh`, written red first; each resolver arm red-checked with the code stashed.


joyful-ii-V-I:
Merged in `35f32aa1`. Thank you — and thank you for basing it on current main with two dedicated gates
and a CHANGELOG entry already written. That is a rare thing to receive.

`def name=(v)` being invisible meant `obj.name = v` resolved to the *getter*, which is a wrong
edge rather than a missing one. Reading the `(assignment left: (call …))` parent to rename the
capture is the right shape, and the stated floor about `obj.count += 1` being an
operator_assignment that both reads and writes is exactly the kind of honesty this project wants
in a comment rather than discovered later.

Landing note: extraction identity went to 80 rather than your 79 — 78 and 79 were taken by the
Elixir and JS/TS landings that went in just ahead of you.

Related follow-up you might enjoy: Ruby is one of four languages where `dependencyCapable` is
still false, so `require`/`require_relative` mint no edges and `--deps`/`--arch` see no Ruby
module dependencies at all. That round is starting now, directly on top of your work.

PR #51, opened 2026-09-07
feat(install): add Hermes as a first-class skill-activation target

AnkitArya:
## What

Add Hermes as a first-class skill-activation target, mirroring the existing Claude (`~/.claude`) and Codex (`~/.agents`) paths — installer, release one-liner, and `wrap` recipe.

**`skills/install.sh`** — new `--hermes` mode:
- `dst="${HERMES_HOME:-$HOME/.hermes}/skills"` (profile-aware via `HERMES_HOME`, same pattern as Codex's `AGENTS_HOME`)
- deploys the flat Agent-Skills-standard set **plus** the Hermes-native skills under `skills/hermes/` (both index side by side; flat name wins on collision); manifest and prune follow, hermes-mode-only
- `--hook` refused with `exit 2`: Hermes exposes `hooks: pre_tool_call` in `config.yaml`, but `hooks/ripwire-nudge.sh` still switches on Claude tool names — the message says "not ported yet", not "no slot"
- `--openclaw --hook` refused with `exit 2` (its `before_tool_call` is a plugin API, not a shell hook slot; previously silently ignored)

**`scripts/install.sh`** — the release-installer ACTIVATE block gains a Hermes branch after Codex: when `${HERMES_HOME:-$HOME/.hermes}` exists, run `skills/install.sh --hermes` and print a receipt line; the `activated==0` manual hint lists `--hermes`.

**`src/wrap.h`** — one `kAgentTargets` row (post-consolidation): CLI-first, MCP via `hermes mcp add ripwire --command <path> --args --mcp` (verified against the argparse source: `--args` is `REMAINDER`, must be last) plus a `# verify: hermes mcp test ripwire` hint; no `--hook` line until the port lands.

**Gates:**
- `test/hermesinstallcheck.sh` (new, alphabetical slot in `test/regression.sh`): hermetic against TEMP homes; asserts the full 17-skill deploy, manifest exact-set, bidirectional hermeticity, idempotent re-run, hook exit-2
- `test/releaseinstallcheck.sh`: new E7 Hermes arm (split `custom-hermes` home proves the explicit `HERMES_HOME` override arrives instead of passing via the `$HOME` fallback); `run_install` applies the empty default before `"$@"`
- `test/skillinstallcheck.sh`: new (E) arm pins the `--openclaw --hook` exit-2 refusal with temp-HOME containment
- `test/wrapverbscheck.sh`: both `--all` runs clear `HERMES_HOME` (same treatment as `XDG_CONFIG_HOME`)

**Gate-count references** 563 → 564 in `README.md`, `docs/EVALS.md` (3 sites), `present/deck5_ripwire_build.js`.

## Why Hermes

Hermes Agent (Nous Research) loads the **same Agent-Skills-standard `SKILL.md` files** ripwire already ships, and its index truncates descriptions at 57 chars + `...` — the shipped flat skills are trigger-first, so they route on the first sentence. `--hermes` wires 17 skills into `${HERMES_HOME}/skills`: the 16 flat user-facing skills plus the Hermes-native `ripwire-repo-map` (`skills/hermes/`, with `metadata.hermes` extras). This makes the one-liner auto-activate ripwire for Hermes the same way it does for Claude and Codex.

## Verified (provenance)

Gates against this repo + TEMP homes (never real ones):
- `bash test/hermesinstallcheck.sh` → ALL PASS (17 skills incl. native, manifest exact-set)
- `bash test/skillinstallcheck.sh` → ALL PASS (incl. new openclaw-refusal arm)
- `bash test/releaseinstallcheck.sh` → all arms PASS incl. E7 (full green needs `build/ripwire` for the isolation helper)

Live against real Hermes (thanks @ashutoshsinghpr7):
- `skills/install.sh --hermes` → `hermes skills list`: 17 local — 17 enabled (flat + native coexist)
- `hermes mcp add ripwire --command <path> --args --mcp` connects, 31 tools discovered; the bare-`--` form fails at argparse as predicted

## CO-AUTHOR / attribution

Authored with Hermes Agent (Ankit Arya). Happy to add/amend co-author attribution per maintainer preference.


joyful-ii-V-I:
Thank you for this, and apologies for the silence — you deserved a status note before now rather than an open PR with nothing on it.

Where it stands honestly: this is queued, not ignored, and it is not blocked on anything wrong with the PR. Skill-activation targets touch how the installer writes configuration outside its own prefix, so the order we add them in is a decision I want to make deliberately rather than in the order PRs arrive. There is also a sequencing question on our side that isn't mine alone to answer, and I'd rather tell you that than invent a technical reason.

I'm not going to give you a date I can't keep. What I can say is that nothing here is a rejection, the PR stays open, and when the ordering question is settled this gets a real review — the same one #57 is getting today: full gate battery, sanitizers, and an adversarial pass rather than just the happy path.

If you'd rather not wait on that, say so and I'll prioritise differently. Leaving a contributor hanging is the failure mode I'm most trying to avoid here.

joyful-ii-V-I:
@AnkitArya — apologies for the churn, and thank you for this. Two things: a collision I caused, and two
facts I would like you to check, because you run Hermes and I do not.

**The collision is mine.** `wrap`'s per-agent lists were just consolidated into one `kAgentTargets` table,
which conflicts with your `src/wrap.h` changes. The payoff is that your six edits there collapse to a
single row, and dispatch, the `wrap` help output, the skills line, the README block and a new gate all
read from it — so Hermes gets covered by that gate the moment the row lands.

**Every fact in the row is yours.** I deliberately did not add it myself; it is your work and you should
land it and keep the authorship. Rebased, #51 is roughly:

```cpp
{ "hermes", "Hermes", "AGENTS.md", "",
  "~/.hermes", "${HERMES_HOME:-~/.hermes}/skills", " --hermes", true,
  WrapPrimary::Cli, McpForm::CliAdd, "hermes mcp add ripwire --command ", " --args --mcp\n", "" },
```

plus your `skills/install.sh` and `scripts/install.sh` changes, which are untouched by any of this, and
your `hermesinstallcheck.sh` — which still earns its place: the new gate checks the table is
self-consistent, yours checks the install actually lands.

**Two things a review flagged that differ from #51 as written. I have NOT verified either against a
running Hermes — please correct me if the review is wrong, because you have one and I do not.**

1. **The MCP registration spelling.** #51 emits `hermes mcp add ripwire -- <path> --mcp`. Reading the
   Hermes CLI source, `mcp add` takes `--command` plus `--args` with `nargs=argparse.REMAINDER` ("must be
   the last option") — so a bare `--` would end option parsing and `<path>` would land as an unrecognised
   positional. If that reading is right, the correct form is
   `hermes mcp add ripwire --command <path> --args --mcp`, which is what the row above uses.

2. **The hook slot.** #51 refuses `--hook` for Hermes on the grounds that it has no Claude/Codex-style
   PreToolUse slot. The review found `config.yaml` supporting `hooks: pre_tool_call:` with a matcher and
   a command, JSON on stdin carrying `tool_name`/`tool_input`/`cwd`/`session_id`, and `{"decision":"block"}`
   or exit 2 to block — structurally the same shape as Claude Code's PreToolUse. If that is current, then
   Hermes DOES have a slot and what is actually missing is our port: `hooks/ripwire-nudge.sh` switches on
   Claude's tool names. That would make `hookSlot` true and the honest refusal message "we have not ported
   the hook yet" rather than "there is no slot" — a different sentence, and only one of them is true.

**One more, on posture.** #51 emits Hermes MCP-first, but your own comment notes Hermes has shell and
terminal tools. Everything shell-capable here now leads with the CLI — it costs nothing until invoked —
and offers MCP below as the warm-index alternative. Hence `WrapPrimary::Cli` above. Push back if that is
wrong for how Hermes actually behaves.

**And a seam between your PR and #46**, which is neither PR's fault. #46 (merged) ships a Hermes-format
skill at `skills/hermes/<skill>/SKILL.md`. Your `--hermes` mode installs `skills/ripwire-*`, and that glob
does not match `skills/hermes/`. So once both are in, `--hermes` deploys the Claude/Codex-format skills
into a Hermes profile while the skill purpose-built FOR Hermes still needs a manual `cp`. You and
@ashutoshsinghpr7 know Hermes' skill loading far better than I do — my instinct is `--hermes` should
prefer `skills/hermes/*` and fall back to the flat set, but that depends on whether Hermes tolerates both
formats in one directory, which I cannot test.

Finally: `scripts/verify-agent-integration.sh` now exists for exactly this problem.
`bash scripts/verify-agent-integration.sh hermes` runs the mechanical half against a throwaway HOME, then
prints two prompts to paste into real Hermes and a markdown block to paste back. **A FAIL in that block is
the useful result** — nobody maintaining this repo runs Hermes, so everything we ship for it is verified
against documentation rather than against a running agent, and that is a gap only you can close.

ashutoshsinghpr7:
**On the question you put to me directly — whether Hermes tolerates both skill formats in one directory: tested, yes.** I ran the branch's real installer (`skills/install.sh --hermes` into a throwaway `HERMES_HOME`):

```
→ done. 16 ripwire skills active in every session (0 pruned, 1 contributor-only skipped)
→ hermes skills list: 16 local — 16 enabled, 0 disabled
```

…then dropped the #46 skill (`skills/hermes/ripwire-repo-map` — Hermes frontmatter: one-line description, `metadata.hermes` extras) into that same `${HERMES_HOME}/skills` next to the flat set:

```
→ 17 local — 17 enabled, 0 disabled
```

So both formats load side by side, and an installer doesn't need prefer/fallback logic to make coexistence work — place both and Hermes indexes them together. The only format difference that matters is *routing*, not loading: Hermes truncates skill descriptions in its index at ~57 chars, so the flat skills route on their first sentence — fine, since the shipped ones are trigger-first.

**On your two fact-checks** (you asked for a running-Hermes answer, so both were tested against one):

1. **MCP spelling — the review is right.** `hermes mcp add ripwire -- <path> --mcp` fails at argparse (a bare `--` ends option parsing); `hermes mcp add ripwire --command <path> --args --mcp` connects — it discovered 31 tools from ripwire's MCP server.
2. **Hook slot — the review is right; #51's wording needs the fix.** Hermes `config.yaml` supports `hooks: pre_tool_call:` with matcher/command/timeout, `fail_closed` (Claude Code's `failClosed` spelling accepted), JSON on stdin, and `{"decision":"block"}` to block — structurally the PreToolUse equivalent. What's genuinely missing is the *port* of `ripwire-nudge.sh` to Hermes tool names (plus its consent flow), not the slot.

**And on the verify script:** it isn't on any public ref yet (main, the branch, or the fork), so the run above is the direct equivalent of its mechanical half. The moment it lands, I'll run `verify-agent-integration.sh hermes` against real Hermes and paste the block back — a FAIL included, as requested.

AnkitArya:
@joyful-ii-V-I — fact-checks closed, seam fixed, pushed as 180b2af5. Point by point:

**1. MCP spelling — the review was right, twice over.** Beyond the live test (bare `--` dies in argparse; `--command <path> --args --mcp` connects), I verified it in the parser source: `mcp_add_p` takes positional `name`, `--command` (with the `dest=mcp_command` quirk), and `--args` with `nargs=argparse.REMAINDER`, help text 'must be the last option'. The row already emits the correct form, so no change there — this commit only appends a `# verify: hermes mcp test ripwire` line (that subcommand is in the official MCP guide).

**2. Hook slot — the review was right.** `agent/shell_hooks.py` docstring, verbatim: stdin `{tool_name, tool_input, session_id, cwd, …}`, stdout `{"decision":"block"}` or exit 2, fail-open unless `fail_closed`. So I reworded the refusal in `skills/install.sh` and the gate text from 'has no slot' to 'not ported yet'. Three caveats I found in upstream while verifying, for whoever ports the hook: default is fail-open; `{"action":"approve"}` on `pre_tool_call` is silently discarded (#92553) — the port must emit block/modify only; and shell hooks silently no-op in the Desktop TUI and `hermes serve` entry points (#43823, #61806), so the port must be tested through the CLI entry. CLI-first posture stands.

**3. The #46 seam — fixed, both sets ship.** `--hermes` now links `skills/hermes/*` next to the flat set (flat name wins on collision), and manifest + prune follow, all hermes-mode-only so other targets are untouched. No prefer/fallback logic, per the coexistence result: `hermesinstallcheck` reports 17 skills including `ripwire-repo-map`, and a new §1c arm names the seam if it ever regresses.

**Drive-bys in the same push (both review-raised):** CodeRabbit's `--openclaw --hook` silent-ignore is now an exit-2 refusal pinned by a new `skillinstallcheck` arm (E) — that was the flagged merge-risk item. And `run_install`'s `HERMES_HOME=` default moved before `"$@"`; E7 now uses a split `custom-hermes` home plus a fallback tripwire, so it genuinely exercises the override instead of passing via `$HOME/.hermes`.

**Verification on my side:** `hermesinstallcheck` ALL PASS, `skillinstallcheck` ALL PASS, `releaseinstallcheck` all PASS except the isolation arm that needs `build/ripwire` (no toolchain in this container; fails identically on the pristine branch). The `wrap.h` change is one string literal — no gate pins hermes emitted output, and the hermes-absent det-gate is byte-unchanged; CI's build covers the rest.

`verify-agent-integration.sh hermes` against real Hermes the moment it lands on a public ref — FAIL included.

Issue #42, opened 2026-09-06
README formatting and wording


heliocipher:
Hi

I didn't mean to be rude or anything, but README looks to me like poorly formatted Claude's Opus 5 word salad.

Please, consider reviewing it.

heliocipher:
Please consider rewriting it with STE100 standards:
https://www.asd-ste100.org
https://github.com/danyuchn/asd-ste100-skill

Whamp:
100 this is brutal to read and even agents get confused from the ai slop verbosity. Great tool but time to refactor the wording. if you want your agent to do i humbly recommend `npx skills add https://github.com/whamp/skills --skill clear-writing`

joyful-ii-V-I:
We hear all of this, a great comment on the skills, we'll work on this the next week or two, we are prioritizing new lang and lang not used as much yet first (so far majority of testing on python/C++), the crazy thing is AI writing code tends to fail, or make less then optimal results in similar ways to writing text verbose, duplicate functions etc, the quality delta actually detects this at some level now (the quality metrics can be pushed further, in our humble opinion), duplicate/near duplicate functions, repo is only about 40 days old and intent to improve it rapidly, fixing the verbose text is definitely on the list...definitely appretiate all the comments, support and taking the time to give valuable feedback

heliocipher:
please see #168

joyful-ii-V-I:
Closing as a duplicate of #73, which tracks the same feedback. Thank you, @heliocipher: your rewrite in #168 is the base of today's README, and the STE100 pointer shaped it. And thanks, @Whamp, for the clear-writing suggestion.

PR #54, opened 2026-09-07
docs: clarify the task example token estimate

PollyBot13:
Related to #42.

## Summary
Upstream substantially rewrote the README after this PR opened. Preserve that layout in full and narrow this PR to the remaining task-example clarification:

- The example uses `--for` without a token limit.
- Describe the historical 4.3K token figure as an observed estimate, not an enforced budget.

The original introduction/navigation rewrite is superseded. The current diff against main is README-only: three lines replaced. No commands, measurements, runtime code, or installation behavior changed.

## Verification
- Current `src/cli.h`: `maxTokens` and `tokenBudget` both default to zero (disabled).
- `git diff --check`: passed.
- GFM converted with Pandoc; clarified text verified in HTML with whitespace normalized.
- All fenced command blocks and details/summary structure match current main exactly.
- Earlier full-suite results belong to the earlier revision, not this updated base. No full build/suite rerun for this prose-only correction; new-head GitHub CI and review are pending.

AI-assisted documentation edit and verification. No merge requested.


heliocipher:
Please consider rewriting it with STE100 standards:
https://www.asd-ste100.org
https://github.com/danyuchn/asd-ste100-skill

joyful-ii-V-I:
Thank you for this, and I owe you a heads-up rather than a review: **the README changed substantially after you opened this**, so this PR will conflict — through no fault of yours.

What happened overnight, so you can judge whether the PR still says what you want it to:

- The token table and the graphs moved above the prose. Time-to-first-evidence went from ~1,009 words to ~300.
- Twenty-six prose blocks moved behind `<details>`, each with a summary that carries its own numbers, so the page argues the same case whether or not anything is clicked.
- The introduction and first-commands area you're editing was part of that.

Your diagnosis was right, which is the annoying part — the introduction *was* doing too much work before the reader got to anything concrete. You identified it and we changed it underneath you within a few hours. That is a bad experience for a contributor and I'd rather say so than let you discover it in a merge conflict.

If you rebase and the current version already covers what you were fixing, close it and I'll take the diagnosis as the contribution — it was correct. If there's still something wrong with the first commands a new reader hits, that's the more valuable PR now, because you're one of very few people who has come to this repo cold recently and can still remember what was confusing.

PollyBot13:
Thanks for the heads-up. The current revision preserves the rewritten introduction and layout; the original navigation rewrite is superseded. I checked against current main again today: one small discrepancy remains in the task example, which calls the answer “token-budgeted” although its command sets neither token control. This PR now only labels the historical 4.3K figure as an observed estimate, not an enforced limit (three lines replaced). Commands and measurements are unchanged, and current-head CI passes.

On the STE100 suggestion: I have kept the correction short and explicit, but I am not claiming formal STE100 compliance or expanding this into another whole-README rewrite.

joyful-ii-V-I:
Merged — thank you, and thank you especially for re-scoping it yourself.

The substance is right, and I verified it against current main rather than taking it on trust: `src/cli.h`
defaults both `maxTokens` and `tokenBudget` to `0`, and the example passes no budget flag, so 4.3K is
what that run *happened* to produce on this repository on 2026-08-30. Calling it "token-budgeted" claimed
an enforcement that was not in force.

That is small in diff and not small in kind. This project spends a lot of effort making its *output*
honest — counts labelled as floors, a zero meaning "none found" and never "none exists", every truncation
disclosed — and then wrote a README sentence that reads as a guarantee about a run with no guarantee
behind it. You caught us applying a rule to the tool and not to ourselves. `grep -c token-budgeted
README.md` is now 0, so it is not left half-corrected somewhere else either.

**And the re-scoping was the right call.** Upstream rewrote the README underneath this PR, and rather than
fighting the conflict you dropped the superseded navigation rewrite and kept only the part that was still
true. That is why this merged as three clean lines instead of a mess. It is a more useful instinct than it
probably felt like at the time.

Related: #42.

PR #56, opened 2026-09-07
fix: resolve JavaScript and TypeScript default imports

PollyBot13:
## Summary

Resolve JS/TS default-import calls by module and exported identity, not by the importer's chosen local name.

For example, `import save from './storage.js'` must call the function exported as default by `storage.js`; an unrelated function named `save` is not evidence of that edge.

This extends the existing ES import/export fact tables to default imports, named default declarations, local identifier defaults, and `export { local as default }`. Existing lexical-shadow and module-ambiguity handling is reused. Named default classes use the same declaration seam to preserve constructor edges. The parser cache version and its quality mirror are bumped together.

Unsupported anonymous expressions remain unresolved. Conflicting default exports or module value bindings cannot select a function simply because it is the only function-shaped symbol.

## Regression proof

The new sub-gate, called by `tsimportprecisecheck.sh`, covers TS/TSX/MTS/CTS and JS/JSX/MJS/CJS:

- An actual exported function whose name differs from the local import, with a same-spelled decoy elsewhere.
- Local identifier and clause defaults; named class compatibility.
- Parameter/closure/block shadows, type-only imports, external imports and competing source/runtime modules.
- Conflicting defaults and function/non-function binding collisions.
- Real cold/warm cache equality and mutations that remove the target after caching.

The original regression fails on the current upstream-base binary; the fixed binary passes. Source review findings were reproduced or covered by added regression arms and repaired before publication.

## Validation

- Development, Release and ASan builds succeed using vendored dependencies with `FETCHCONTENT_FULLY_DISCONNECTED=ON`.
- `bash test/lib/jsdefaultimport.sh <binary>` passes for all three build flavours. Existing `jsimportalias.sh` and `jsimportfacts.sh` pass in development and Release.
- `python3 test/pargates.py . ./build/ripwire -j 4`: 571 gates, 565 pass, 3 skip, 3 fail on the first full run. The parser-identity fingerprint and two README example coordinates were then refreshed in a separate proof-reference commit; `qschemetripcheck.sh`, `qextractionkeycheck.sh` and `readmeexamplecheck.sh` pass after that refresh.
- The remaining full-run failure, `legendcoveragecheck.sh`, scans installed host skills with bare `--scan-skills`; it passes with a disposable empty home. The legend baseline and host skills are unchanged.
- The skipped formatting gate passes when rerun with clang-format 22, including its mutation controls. Two optional pre-change-binary comparison gates remain skipped; this change intentionally alters graph output.
- Full-source ASan scan exits 0 with empty stderr. Three uncached maps and three cache-backed maps are byte-identical and valid XML.
- Independent review completed; source findings fixed and rechecked. Diff and shell syntax checks pass.

## Scope and limits

No new graph mechanism, general TypeScript semantic analysis, anonymous-symbol synthesis, barrel traversal, reassignment analysis or `--uses` behavior change. Existing named-import handling stays on its current path. The meaningful behavior change is confined to default bindings; reverting the commit restores the previous extraction/resolution behavior.

AI-assisted implementation with independent source review. Validation is static-source/real-CLI proof, not execution of the JavaScript programs.


joyful-ii-V-I:
Thank you — and thank you for coming back. Your named-import fix from #45 shipped in [v0.5.0](https://github.com/redhat-et/ripwire/releases/tag/v0.5.0) yesterday and you're named in the changelog for it; opening three more PRs the same evening is the best signal this project has had that the contribution path actually works.

Default imports are the right next gap. `import X from "./m"` is the spelling most JS/TS code actually uses, and #45 left it on the floor.

Queued behind #57's review — we're prioritising language coverage over new surface for the next couple of weeks, and this sits squarely in that. I'll give it the same treatment: full battery, ASan, and an adversarial corpus rather than just the fixture. If it holds up the way #45 did, it lands.

joyful-ii-V-I:
Merged — thank you, this is a careful fix and the reasoning in the description is the best part of it.

"Resolve by module and exported identity, not by the importer's chosen local name" is exactly the right
framing, and I verified it independently of your gate before merging, on a fresh fixture: a default
export `persistToDisk` in one module, an unrelated exported `save` in another, and
`import save from './storage.js'`.

```
origin/main   run() -> <c n="save"/>                          the DECOY, matched on the local name
this merge    run() -> <c n="persistToDisk" prov="import"/>   the real default export, with provenance
```

Three things you did that I want to name, because they are the parts that made this cheap to land:

**You bumped `kParserVer` and `kIngestParserVerMirror` in the same commit.** That mirror is the single
nastiest trap in this repo — CLAUDE.md documents a case where it was edited during a build and the binary
kept emitting the old value through repeated successful rebuilds. Getting it right unprompted is not
typical.

**You left `kCacheVersion` alone and said why.** New extraction facts on an unchanged record shape, so it
is not yours to move. That judgment is what let the merge conflict resolve cleanly (see below).

**Your new coverage adds NO name to the absorb list.** Putting the logic in `test/lib/jsdefaultimport.sh`
and calling it from the existing `tsimportprecisecheck.sh` means this PR does not touch the eight sites
that publish the gate count — with three lanes in flight adding gates this week, that mattered more than
you could have known. Worth copying whenever a "new gate" is really a new ARM of an existing one.

**On the conflict.** Both this branch and #57 (Ruby constant references) spent parser version **82**. I
renumbered yours to **83** on merge and kept BOTH changelog entries so neither lineage is lost;
`kIngestCacheVersionMirror` stayed at main's 17 for exactly the reason you gave; and `qschemetrip.hash`
was re-pinned through its own `UPDATE_GOLDEN` path rather than by picking a side of the conflict, since a
hand-picked hash of source text would have been wrong either way.

Nothing was needed from you. Thanks again.

PR #64, opened 2026-09-08
Add Gleam language support

kendall:
Adds first-class Gleam parsing and navigation for `.gleam` files. Functions, external declarations, custom types, type aliases, local/qualified/pipeline calls, test attribution, and import-aware module resolution now participate in maps and dependency analysis.

Gleam imports resolve only through exact `src/<module>.gleam`, `test/<module>.gleam`, or crawl-root paths. Multiple matches degrade instead of guessing. The regression fixture includes a duplicate function name and verifies that the imported module selects the intended definition.

On 2,569 tracked Gleam files from a production corpus:

- Ripwire cold map: 1.04s
- CRG raw graph build: 44.70s (43x slower), with a 454 MiB database
- Import resolution reduced ambiguous edges from 18,635 to 15,121 (18.9%)
- Ten sampled caller queries agreed on all non-recursive callers; CRG alone reports recursive self-edges

Validation:

- `bash test/gleamcheck.sh`
- `bash test/elixircheck.sh`
- `bash test/deplangscheck.sh`
- `bash test/dependencypincheck.sh`
- `bash test/g1configcheck.sh`
- `bash test/ripwirepubliccheck.sh`
- `bash test/readmedriftcheck.sh`
- `bash test/printffmtparitycheck.sh`
- `git diff --check`


kendall:
Closing because this change is intended for the kendall/ripwire fork.

PR #65, opened 2026-09-08
feat(ruby): a constant receiver is a dependency, and a lazy edge is a use, not structure (parser version 83)

andriytyurnikov:
Round two of the Ruby constant work, on top of #57. Three commits, gate-first each.

## What it does

**`User.find` edges to the file that defines `User`.** A `call` whose receiver is a constant chain (`A::B::C`, `::A::B`) is a symbolic directive, spelled as written, resolved through the round-one index. This is the Zeitwerk dependency proper: a Rails file almost never says `require`, it names the constant and the autoloader loads the file on that first reference. Before this, `--impact=User` on a Rails app named nobody; now it names the 197 files that call it.

## Decisions (stated, not asked)

1. **What counts.** A receiver whose chain head is not a constant (`repo::Finder`, `self.class`, an identifier, an ivar) is nothing — `rubyIsConstantChain`; the round-one reader accepts any scope-resolution text and is right for the positions the grammar restricts to constants, a receiver is not one. A constant used as an **argument** (`raise Errors::Boom`, `validates_with Foo`) or a rescue class is a **disclosed floor** of this round.
2. **Dedupe at extraction**, per (file, innermost class/module open, written name). Zeitwerk loads a constant once; the second `User.find` in the same body is not a new dependency. First occurrence carries the byte and the lazy bit. The nesting is in the key: the fixture has one file with `User.find` under two nestings resolving to two different files. `Time` and `::Time` are two spellings, two directives. Measured on four corpora, distinct sites are 53–67 % of raw ones.
3. **Lazy inside a closure.** A receiver inside a `method`, `singleton_method`, `lambda`, `block` or `do_block` is `lazy="1"` — the parser-72 TS/JS function-body rule on Ruby's closure kinds. Class-body and file level are load-time. A `do`-block passed to a class-level macro is lazy under this rule even when the callee runs it at load: the tool cannot see the callee.
4. **The Ruby walk descends every node.** A receiver is an expression (assignment, argument list, lambda, interpolation…); the 19-kind statement-level allowlist would have needed ~40 and every miss would be a receiver silently dropped — a floor the tool could not disclose because it could not see it. Same cost the reference pass already pays. Wall time unchanged.
5. **Structure versus use — the decision that reaches TS/JS too.** With receivers counted like every other edge, a Rails app is one strongly-connected core: measured, ccd 12 740 → 1 407 232 and every Ruby corpus read `tangled`. True of runtime references, useless as a lens. So a **lazy edge** — a pair every one of whose directives is written inside a closure, or is a Ruby `autoload` — is a **use**, not a load-time dependency:
   - **use**: `--impact`'s importer tier (`lazy="1"`), the file's own `<inc t=>` rows, call-resolution narrowing, `--expand` siblings, `--cochange`'s static-coupling test keep every edge;
   - **structure**: `--deps`/`--arch`/`--report` leave lazy pairs out of `afferent=`, `instab=`, `transitive=`, godfiles, stabledeps, cycles, ccd/acd/nccd/`shape=` (`graph.h::resolveStructuralIncludeAdj`; one load-time directive makes the pair load-time — `recordLazyPair`'s existing rule).
   Disclosed where it is made: `<health lazy_edges=N>` counts the pairs left out, a row's `lazy_edges=N` its own, both absent when 0 — a corpus without a lazy directive is byte-identical. A lazy edge is not an unresolved one: the unresolved row has neither `lazy_edges=` nor an importer; the lazy row has both. **This moves one TS/JS number**: a `require()` inside a function body was already `lazy="1"` in the importer tier and is now also out of `--deps`' structure; no gate pinned it inside the cone.

## Measurements (`--deps --limit=100000`, parser version 82 → 83)

| corpus | files= | ccd | nccd | shape | importees | lazy_edges | bytes |
| --- | --- | --- | --- | --- | --- | --- | --- |
| activesupport `lib/` (282) | 206 → 241 | 10 450 → 15 299 | 5.19 → 7.59 | tangled → tangled | 241 → 201 | 945 | 55 483 → 75 970 |
| activerecord `lib/` (395) | 300 → 368 | 2 714 → 4 088 | 0.90 → 1.36 | horizontal → vertical | 385 → 310 | 1 026 | 71 474 → 114 745 |
| a Rails app, 4683 files / 3532 `.rb` | 2 296 → 3 245 | 12 740 → 13 170 | 0.31 → 0.32 | horizontal → horizontal | 385 → 414 | 5 632 | 378 473 → 750 913 |
| a second Rails app, 1957 / 1895 `.rb` | 1 072 → 1 455 | 6 289 → 6 382 | 0.33 → 0.34 | horizontal → horizontal | 189 → 197 | 1 830 | 173 606 → 363 733 |

Read together: on the apps the structure barely moves and `lazy_edges` says how large the runtime layer it leaves out is. On the gems the structure grows through class-body receivers (`ActiveSupport.on_load`, `Concern`) and importees **fall**: a file only ever called from inside methods is not a god file, it is a file `--impact` lists. The uncapped payload roughly doubles on Ruby corpora; the default cap is unchanged.

## Floors and record shape

- Argument and rescue constants are not receivers (floor, next round if wanted). Aliases (`Point = Struct.new`) are still not indexed; the `Struct` receiver is a shown row.
- Out-of-tree receivers (`Time`, `Rails`, `Struct`) are shown rows with no edge — the posture every Python `import os` row has.
- Record shape unchanged: a receiver is an `Include` with `isSymbolic` + `byte` from parser version 82, so cache format 17 holds; only the extraction identity moved.
- Round one's gate moved two arms honestly: `dynamic.rb`'s row is the `Object` receiver of `Object.const_get`, `point.rb`'s row is its `Struct` receiver with `afferent="0"`.

## Verification

- `test/rubyrecvcheck.sh` + `test/rubyrecvfix/` (17 files): 14 arms red against the parser-82 binary, 7 more red against the receiver build before the structure cut; every mutation control and floor arm passes there. 558 → 559 gate scripts.
- Full battery all pass. ASan/UBSan clean on both fixtures and activesupport (deps cold/warm, `--impact`, `--report`). Cold == warm on activerecord and the Rails app. Default map byte-identical on this repo's C++ `src/`.
- Re-pins with reasons in-file: `qschemetrip.hash` (parser mirror), `printf_parity.manifest` (`--help` bytes), `fixedbufsweep` S6 (+2 `snprintf` in `packDeps`, no `%s`). `docs/COMMANDS.md` regenerated from the binary.


joyful-ii-V-I:
Merged — thank you. This is the second Ruby round today and the reasoning in the description is what made
it landable.

**"A Rails file almost never says `require`, it names the constant and the autoloader loads the file on
that first reference"** is the sentence that justifies the whole change. `--impact=User` naming nobody on
a Rails app was a real hole, and constant-receiver-as-symbolic-directive is the right shape for it —
spelled as written, resolved through round one's index, rather than guessed.

**The decisions-stated-not-asked section is the part I wish more PRs had.** Saying up front that a
receiver whose chain head is not a constant (`repo::Finder`, `self.class`, an ivar) is *nothing* — and
naming `rubyIsConstantChain` as where that lives — meant I could check the boundary instead of inferring
it.

**One thing I changed on merge: the parser version.** You claimed 83; main had already spent it. That is
not your error — #56 (JS/TS default imports) claimed 82, which #57 had taken, so I renumbered #56 to 83
this afternoon and yours became the third collision of the day. It is now **84**, with both changelog
entries kept and a line recording the renumber so neither lineage is lost. `kIngestParserVerMirror` moved
in the same commit.

Three lanes branching from one base and each taking "the next number" is a coordination problem rather
than anybody's mistake, and it will keep happening while several rounds run in parallel.

Also resolved on the way in, none of it yours:

- `test/regression.sh`'s absorb list unioned to 561 names — taking main's line verbatim and inserting
  `rubyrecvcheck`, because rebuilding that line from parsed tokens drops names and the trailing `; do`.
- The eight published gate-count sites did **not** conflict, which is the dangerous case: main said 560,
  your branch said 559, and git would have merged one of them cleanly into a number wrong either way.
  Resolved to main's side first, then re-derived.
- `docs/COMMANDS.md` regenerated rather than merged; `test/printf_parity.manifest` re-pinned with only the
  `help` label moving; `test/qschemetrip.hash` re-pinned with `kQSnapCacheScheme` left at 8, since your
  branch does not touch `src/quality.h` and the hash moved only because the version mirror lives inside
  the hashed region.

CI green on every leg. `rubyrecvcheck` and `rubyconstcheck` both pass.

PR #75, opened 2026-09-09
Dart: a 23rd grammar, and the span arm its tree shape requires

calvinchengx:
# Dart: a 23rd grammar, and the span arm its tree shape requires

Adds Dart (`.dart`) as a first-class language: definitions, call edges, metrics.

Measured on a real Flutter corpus that is not in this tree — 2,959 files,
39,570 symbols, indexed cold in 1.1 s.

Three commits, smallest and most independent first.

---

## 1. `fix(model)` — six per-language arrays derive their extent from `kLangCount`

**This one is a pre-existing bug and stands alone. It does not mention Dart and can land
without the rest of the PR.**

`model.h`'s `kLangCount` carries an explicit warning:

> any per-language array sized by a LITERAL silently drops the tail when a language is
> appended, and the drop is invisible because the affected code paths just see a zero.
> Size per-language arrays with this, never with a number.

Six sites did exactly that, in a different spelling: `std::size_t( Lang::Elixir ) + 1` —
the last enumerator written out by hand rather than derived. It is not a number, but it
duplicates the knowledge of *which enumerator is last*, so it goes stale the moment one is
appended, and it goes stale silently because the guarded loops just skip the new value.

| Site | What goes silently missing |
| --- | --- |
| `verbs_report.h` `symbolTally` / `fileTally` | the `--skipped` `<lang>` census — no row at all, and nothing says a row is missing |
| `lintrules.h` | the `dependencyCapable` scan bound |
| `verbs_navigate.h` | `kExtSurfaceLangSlots` |
| `htmlexport.h` | the `kLangColors` static_assert |
| `tsprobe.cpp` | the `kLangName` static_assert |

**Behaviour-preserving by construction at this HEAD**, because `kLangCount` is *defined* as
`Lang::Elixir + 1` there — every substitution is an identity today. The value is that the
next language updates one constant instead of seven, and the two static_asserts become real
guards on their tables instead of restatements of the same hand-written extent.

Found the hard way: appending a 23rd grammar made the language census lose 2,959 files and
39,570 symbols with no error and no missing-row disclosure. 5 files, +6/−6.

---

### Gate first

`test/dartcheck.sh` was written and run **RED** against a pre-Dart binary — all 21 fixture
symbols missing — before any grammar landed. It now covers extraction, member kinds, call
heads, cascades, the constructor floor, call-site mutation, metrics, a malformed-input
arm, and cold/warm determinism.

### The tree shape that made this more than a table row

tree-sitter-dart makes `function_body` a **sibling** of `function_signature` /
`method_signature` — never a `body` field, never a child. The shared ancestor walk in
`ingest_sidecap.h` finds no body, the definition's span stops at the closing paren of the
signature, and every call in the body attributes to the nearest **enclosing** symbol.

Measured on `test/dartfix` before the fix:

- `square` landed on the class `Calculator` instead of the method `accumulate`
- the three top-level edges (`twice→square`, `answer→secret`, `branchy→square`) were lost entirely
- **5 edges where 8 were expected**

The `Lang::Dart` arm adopts the following `function_body` sibling and runs the byte extent,
the row extent and `complexityOf` through it — the same shape LB-E already uses for a
test-macro block. An abstract member (`void f();`) has no such sibling, so it stays a
declaration. One line in `cc_isParamList` teaches `countParams` Dart's
`formal_parameter_list`.

### The query is adapted, not copied

Upstream ships a `queries/tags.scm`. Its trailing `@reference.call` is an optional-heavy
match over selector chains that fires on the **receiver** as well as the invoked member.

`selector` holds exactly **one** child, so a call is a pair of *sibling* selectors, and the
receiver's own next sibling is the member selector rather than an argument list. That is
what keeps `this..add(1)..reset()` two calls instead of three. Pinned by the gate's cascade
arm, which asserts `this` and `items` are absent from `build`'s edges.

### Stated floors — documented in the query, README and ARCHITECTURE

- Named constructors and factories index under the **class** name: `C()`, `C.seeded()` and
  `factory C.fromA()` are overloads of `C`. Call sites spelled `C.seeded(1)` resolve to the
  member name, so they are unresolved rather than wrong. Asserted, so the floor cannot
  quietly become a silent drop.
- `noSuchMethod` dynamic dispatch names its callee at run time; not an edge.
- `part` / `part of` composition is not resolved.

`kParserVer` 81 → 82 with `quality.h`'s `kIngestParserVerMirror` in the same commit.
29 files, +381/−29 first-party (plus the vendored generated parser).

---

## 3. `test(dart)` — re-pin the three generated artifacts the grammar moved

Each re-pinned with its diff reviewed, not blind-repinned:

- **`test/printf_parity.manifest`** — exactly **one of 12 verbs** moved: `help`, the language
  line gaining `Dart (.dart)`. The other 11 are byte-identical, so no printf-family
  conversion rode along.
- **`test/qschemetrip.hash`** + a RE-PIN LOG entry — `kParserVer` 81 → 82 rejects a v81 blob
  on a tree holding `.dart` files. Record shapes are unchanged so `kCacheVersion` stays 16;
  no Snapshot-side function changed so `kQSnapCacheScheme` stays 8.
- **`docs/COMMANDS.md`** — regenerated from the binary; **one line** differs.

---

## Verification

| Check | Result |
| --- | --- |
| `test/dartcheck.sh` | 7/7 arms, plain **and** ASan binaries |
| Full gate suite (plain build) | `gates=572 pass=568 skip=3 fail=1` |
| ASan + LSan — `test/dartfix` | clean |
| ASan + LSan — self-run on this repo | clean |
| ASan + LSan — 2,959-file external Dart corpus | clean |
| Determinism | byte-identical across runs, cold and warm |
| XML | `xmllint --noout` clean |
| **Every other language** | **byte-identical** vs the pre-change binary, on `src/` and on the multi-language `test/` fixture corpus |

That last row is the one that matters for a change touching shared code in
`ingest_sidecap.h`: the new arm is gated on `Lang::Dart`, and the other 22 languages produce
the same bytes they did before.

### The one failing gate is pre-existing

`editroundtripcheck`'s CRLF arm. Verified three ways:

1. this branch touches neither that gate nor `--expand`/`serialize`
   (`git diff --name-only main...HEAD` is empty for those paths);
2. the fixture is written **by the gate at runtime** with explicit `\r\n` and `newline=''`,
   so it is not a checkout line-ending artifact;
3. it **fails identically under a binary built from `main`**.

It is a property of my local checkout or this macOS build, not fallout from this branch. I
have not touched it.

### Platform

Built and tested on macOS arm64 with Apple clang 21 and CMake 4.4 — a toolchain combination
CI does not currently cover. Zero warnings, zero errors, in both the plain and ASan trees.

### Fuzzing

`ripwire_fuzz_dart` built with `-DRIPWIRE_FUZZ=ON` (Homebrew LLVM 23 — Apple clang ships no
libFuzzer runtime, and the CMake check refuses rather than guessing). Corpus seeded from
`test/fuzz/seeds/dart/valid` plus the two `test/dartfix` files:

| | |
| --- | --- |
| Executions | **109,431** over two campaigns (121 s + 401 s) |
| New corpus units | 5,540 |
| Crashes / leaks / timeouts / OOMs | **none** — no artifacts written |
| Peak RSS | 525 MB |
| Slowest unit | < 1 s |


joyful-ii-V-I:
Flagging for the review in progress — CodeRabbit posted three findings on this PR, and **one of them is a must-fix before this lands.**

## `test/dartcheck.sh:117` — the assertion cannot fail

```python
assert 'answer' not in {c.get('n') for c in syms['C'].iter('c')} or True
```

The trailing `or True` makes it pass unconditionally. It is the only always-true assertion in the gate, and it sits directly under a comment stating the opposite:

> `C.seeded(answer())` resolves to the member name, which has no definition here — unresolved, never silently bound to the class.

So the arm reads as if it enforces that contract and enforces nothing. This repo has been bitten by that exact shape repeatedly in the last two days — a fixture that stays green against a full revert, a presence guard written `>=` instead of `==`, a reverse-apply check run against its own commit. A gate arm that cannot fail is worse than an absent one, because it is counted as coverage.

Drop the `or True`, then **verify the arm actually goes red** against a binary built without the fix. If it does not, the assertion is testing something other than what the comment claims and the comment needs to change too.

## Two more worth verifying (not yet confirmed either way)

**`src/ingest_sidecap.h:1663`** — the claim is that the vendored Dart grammar marks `comment` and `documentation_comment` as *named* nodes, so a comment sitting between a signature and its `function_body` ends the definition span early, and calls inside the body then get the wrong symbol attribution. That is a specific, testable shape and it is the same failure class as a published `@FILE:LINE` seed resolving to the wrong symbol. Worth a fixture with an intervening doc comment at top level and in a method — if it reproduces, it is a real extraction bug; if not, say so and move on.

**`src/ingest_metrics.h:1371`** — the claim is that Dart wraps optional and named parameters in `optional_formal_parameters`, so counting direct children of `formal_parameter_list` undercounts arity. A recent audit here found `param-walk depth 12` already makes arity wrong in another language, so the class is live. Check it against the grammar's `node-types.json` rather than against the prose.

---

Nothing here is a reason to hold the PR beyond fixing the first item — the `kLangCount` catch in commit 1 is a genuinely good find, the gate and fixtures are the right shape, and the corpus measurement is the kind of evidence this repo asks for.

joyful-ii-V-I:
Landed on `main` as [`6f91fed2`](https://github.com/redhat-et/ripwire/commit/6f91fed29b59cb810bce8fb2c3cbc87793d7d28f) (via #106, which existed only because the `workflow_dispatch` endpoint was returning 403 for the CI matrix). Thank you — this is one of the most carefully assembled
contributions this repo has had, and three specific things earned that.

**The `kLangCount` catch holds up, and it is the best thing in the PR.** I checked the standalone
claim two ways. Statically: at the base commit `kLangCount` is *defined* as
`static_cast<std::size_t>( Lang::Elixir ) + 1`, your commit does not touch `model.h`, and every one of
the six substitutions — including `i <= std::size_t( Lang::Elixir )` → `i < kLangCount` — is an exact
identity. It compiles, it is behaviour-preserving, it could have landed alone on any day with no Dart
in sight. Dynamically, because "latent" deserves a measurement rather than an argument: I reverted
just the two `verbs_report.h` tallies on the finished branch, rebuilt, and ran `--skipped` over a
four-file corpus of two `.cpp` and two `.dart`. It printed `indexed="4"` and exactly one language row,
`<lang n="cpp" files="2" symbols="5"/>`. Two indexed files gone from the census with nothing saying a
row was missing. With your commit: `<lang n="dart" files="2" symbols="22"/>`. That is a latent bug
this PR both exposes and fixes, and the merge commit says so.

**The gate is red-first in the strong sense, not just the easy one.** Against a pre-Dart binary built
from current `main` it fails with all 21 fixture symbols missing, exit 1 — exactly what the FAILFAST
row you added to `gateexitcheck.sh` claims. More importantly it also goes red against the *span arm
alone*: I disabled only the `Lang::Dart` branch, rebuilt, and reproduced your measurement to the
edge — 5 edges where 8 are expected, `square` attributed to the class `Calculator` instead of the
method `accumulate`, and `twice→square`, `answer→secret`, `branchy→square` gone. A gate that only
dies when the whole grammar is removed proves much less than one that dies when the interesting line
is removed. Yours does the latter.

**The cascade query is the part a reviewer would have got wrong.** `this..add(1)..add(2)..reset()`
staying two call edges rather than three, because `selector` holds exactly one child and the
receiver's own next sibling is the member selector rather than an argument list — that is a real read
of the parse tree, not a copy of upstream's `tags.scm`, and the gate pins both the positive
(`{'add','reset'} <= calls('build')`) and the negatives (`this`, `items`). The `overloads=` assertion
on the constructor floor is the same instinct: you made the floor *visible in the output* instead of
only writing it in a comment.

### The STEP 0 number the PR was missing, measured for you

`prompts/add-a-language.md` asks for a parse rate on a real corpus, and the PR did not carry one. On
**flutter/packages** at HEAD:

| | |
| --- | --- |
| `.dart` files on disk | 3,896 |
| degraded parses (ERROR/MISSING nodes) | **0** |
| files the crawl passed over | 0 |
| files carrying ≥1 symbol | 3,706 |
| Dart symbols extracted | 71,726 |

A **100% parse rate**, which is comfortably past the 95% bar that prompt sets. The 190-file gap
between 3,896 and 3,706 is not loss: every one I sampled is a pure re-export barrel
(`export 'src/android_camera.dart';` and nothing else), which correctly carries no definitions, and
`<lang files=>` is documented as a floor derived from symbol-bearing files. Of the 289 degraded
parses in that 6,554-file corpus, **none** is a `.dart` file. Cold index of the whole corpus was
4.62 s, and 9.54 s under the full G1 sanitizer stack with `-fno-sanitize-recover=all`, exit 0, zero
reports.

Three more things I can confirm on the record:

- **Supply chain.** `parser.c` (7,136,472 bytes), `scanner.c` and `LICENSE` under
  `third_party/deps/dart/` are byte-identical to `UserNobody14/tree-sitter-dart@be07cf71`, the SHA
  you pinned in `CMakeLists.txt` and `THIRD_PARTY.md`. Nothing was edited on the way in.
- **The narrow-counter class that PR #102 just fixed in four other scanners does not apply here.**
  Dart's vendored scanner is stateless (`create` returns NULL, `serialize` returns 0) and its only
  counter is `unsigned nesting_depth`. No `uint8_t`, no `int8_t`, nothing to wrap.
  `vendorpatchcheck.sh` is green with Dart carrying no patch.
- **"Every other language is byte-identical" reproduces.** The pre-Dart binary and this one produce
  identical bytes on `src/` and on a 1,406-file multi-language corpus. `formal_parameter_list`, the
  node kind you added to `cc_isParamList`, appears in the node table of exactly one of the 23
  vendored grammars — Dart's — so it structurally cannot reach another language.

### What I changed on your branch, and why

Your three commits are preserved with your authorship. I added four of my own and amended one of
yours.

1. **Rebase.** `main` moved repeatedly while this was open, twice more during the landing itself, so
   every pinned number was re-derived rather than carried. `kParserVer` is **88**: the branch was
   written against 81 → 82, I re-derived it to 87, and then `main`'s vendored-scanner lane took 87
   *exactly* — the declaration line auto-merged clean at that same wrong number while only the
   history comment beside it conflicted. `kCacheVersion` stays 18, `kLangTable`'s extent is 47
   (`main` added `.rst/.adoc/.org/.mdx`), and the published gate count is **578**, taken as the union
   of `main`'s 577-member loop with `dartcheck` rather than by adding one to a number.
   `qschemetrip.hash` and `printf_parity` were re-answered from a binary built on the merged tree.
   Your commit 3 carries the pins with a maintainer note in its message saying they were re-derived.
2. **`test(dart)` — an assertion that cannot fail.** The constructor arm ended in
   `assert 'answer' not in {…} or True`, which is true for every input: `CONTRIBUTING` §2 shape 5, an
   arm still in the file but no longer in the conjunction. The property you were reaching for is
   real, so it is asserted directly now — neither `seeded` nor `C` may appear as a callee of the
   merged `C` row. For the record the reason that row has no callees at all is *not* Dart: three
   overloads share the name `C`, and a merged row keeps the first overload's edges. A C++ `f()` /
   `f(int)` pair behaves identically, so nothing about it belonged in a Dart floor. Same commit:
   `elixircheck.sh` asserted 23 grammars while printing "all 22", and `THIRD_PARTY.md` said `6.8M`
   where all 22 sibling rows say `6.8 MB`.
3. **`fix(dart)` — the half of the registration surface the PR missed.** `prompts/add-a-language.md`
   names `src/lintrules.h`'s extension→Lang map in STEP 2 and `kUnanalyzedLangs` in STEP 6, and both
   were still at 22 languages. Both consequences are output the honesty contract governs:
   - `--nonlocal-state` said **nothing** about a Dart corpus. `.dart` was in neither `langOfPath` nor
     `kUnanalyzedLangs`, so a corpus that was half Dart emitted no `unanalyzed_langs=` at all — not a
     zero a reader could weigh, the attribute simply absent. The identical corpus with `.lua`
     substituted prints `unanalyzed_langs="lua" unanalyzed_files="3"`. That contrast is now the
     gate's control arm.
   - `--lint` printed a row contradicting itself. `applicable="0"` means "none of this rule's
     languages are present", yet `naming-short` and `naming-case` both fire on a four-line `.dart`
     fixture — with Dart outside `lintcatalog.h`'s `kCatalogLangs` those rules printed `count="1"`
     and `applicable="0"` together.

   Both arms were observed RED against this branch's own binary before the tables moved. Also folded
   in: `clones.h`'s mask-width guard was `static_assert( std::uint32_t( Lang::Elixir ) < 32 )` — the
   last enumerator spelled out by hand, exactly the shape your first commit is about, and stale the
   moment Dart was appended. It reads `kLangCount` now. Dart stays *out* of
   `kHashLineCommentLangMask`: its line comment is `//`.
4. **`refactor(dart)` — the span arm moved behind one call.** `--quality-delta` exited 2 on
   `captureTagsFacts`, complexity 262 → 281 and LOC 474 → 503 against a bar of 15/60.
   `defBodyNodeOf`, four lines above your call site, says what the house does about exactly that:
   *"Kept out of captureTagsFacts (the file's densest dispatch point) behind one call."* Elixir does
   the same through `elixirBody`. `dartFollowingBody` now does too, carrying your measurement and the
   abstract-member contract verbatim in its doc comment, and the delta drops to +5/+13. What remains
   is the standing cost of adding any language here — the Elixir landing carried six gating findings
   of the same shape.
5. `chore(docs)` / `chore(pin)` — `LIMITS.md` line drift (`limitstablecheck.sh` is one of the twenty
   gates `main` landed while this was open, so the branch met it for the first time on the rebase),
   and the pin re-derivation.

### Two small gaps, named rather than left as a label

Neither blocks anything, and I did not fix either:

- **Dart `import` / `export` are not extracted**, so `Lang::Dart` is not dependency-capable and
  `--deps`/`--arch` do not count `.dart` files. That is a floor rather than a defect, and it *is*
  disclosed in output — `<health dep_langs=>` publishes the capable set and Dart's absence is
  visible — but it is what those 190 barrel files would light up. STEP 5 of
  `prompts/add-a-language.md` is the shape it would take, and it is its own round.
- **`test/dartfix` has no decoy.** STEP 7 asks for "the same symbol name in a file that must *not*
  win"; the fixtures are adversarial about cascades and constructors but not about resolution.

I removed the `help wanted` label. Nothing here still needs outside help.

joyful-ii-V-I:
Merged as 6f91fed2 — closing manually because the rebase onto a moving `main` gave every commit a new SHA, so GitHub cannot match this branch to what landed. Your three commits are in the ancestry with your authorship intact.

PR #78, opened 2026-09-09
fix(ruby): a later load-time receiver site clears the lazy bit; the constant-chain check is a loop (parser version 85)

andriytyurnikov:
Follow-up to #65. Two commits, gate first. Fixes the three review findings on that PR, each reproduced before it was fixed.

## What was wrong

**A load-time site below a lazy one was lost.** The receiver dedupe keeps one `Include` per (file, innermost open, written name), and the first occurrence in source order carried the lazy bit. A `Helper.fmt` inside a method written *above* the same `Helper.fmt` at class-body level left the directive lazy; resolve.h's pair rule never saw the load-time site, and the structure dropped a real dependency. Two files that differ only in the order of those two lines read `ccd="3" shape="vertical"` with a god file one way and `ccd="2" shape="horizontal" lazy_edges="1"` the other — the answer depended on statement order.

**A 5 000-segment chain killed the run.** `rubyIsConstantChain` recursed once per segment of a left-nested `scope_resolution`, and the depth bound in `captureIncludes` sits *after* `directiveTargetOf`, so a generated `A::A::…::A.call` overflowed a parse worker's stack — SIGBUS, exit 138, no output, measured on macOS; 2 000 survived. The crash report puts every frame in `rubyIsConstantChain`.

**`--help` said `lazy="1"` was TS/JS only.** It has covered Ruby closures and autoloads since the receiver round.

## What changed

- `captureIncludes`: `seenReceivers` maps the key to the emitted record's index; a later occurrence that is not lazy clears `isLazy` on it. The first site still carries the byte. The lazy bit is the AND over occurrences, which is what the pair rule already assumed.
- `rubyIsConstantChain` walks the `scope` field in a loop. Nothing else on the path recurses per segment: the walk is an explicit stack, the resolver splits the text.
- The `--impact` help line, and `docs/COMMANDS.md` regenerated from it.
- `kParserVer` 84 → 85 with `kIngestParserVerMirror` (a cached lazy bit could be wrong, so cached Ruby files re-parse once). Record shape and cache format 17 unchanged. Re-pins with reasons in-file: `qschemetrip.hash`, `printf_parity.manifest` (`help` bytes).

## Measured (`--deps --limit=100000`, parser version 83 → 85, the same four corpora as #65)

| corpus | ccd | nccd | shape | lazy_edges | bytes |
| --- | --- | --- | --- | --- | --- |
| activesupport `lib/` (282) | 15 299 → 15 299 | 7.59 → 7.59 | tangled | 945 → 935 | 75 970 → 75 988 |
| activerecord `lib/` (395) | 4 088 → 4 325 | 1.36 → 1.44 | vertical | 1 026 → 1 023 | 114 745 → 114 763 |
| a Rails app, 4683 files / 3532 `.rb` | 13 170 → 13 172 | 0.32 → 0.32 | horizontal | 5 632 → 5 630 | 750 913 → 750 915 |
| a second Rails app, 1967 / 1895 `.rb` | 6 382 → 6 382 | 0.34 → 0.34 | horizontal | 1 830 → 1 830 | 363 733 → 363 750 |

The pairs that flip are the ones written lazy-first and load-time-second in one body: ten, three, two, none. On activerecord three of them sit on a spine (ccd +237). No shape moves. Wall time unchanged. Cold == warm on activerecord and the first app; two `--no-cache` runs identical on all four.

## Verification

- `test/rubyrecvcheck.sh` gains `lib/app/eager_after_lazy.rb` (18 fixture files; ccd 20 → 22, helper.rb afferent 2 → 3, a row arm with no `lazy_edges=`, the `--impact=Helper` importer tier) and a deep-chain arm that generates a 150 000-segment receiver at gate time. Six arms red against the parser-version-84 binary: the eager-after-lazy importer reads `lazy="1"`, health reads `ccd="21" lazy_edges="10"`, the deep chain exits 138.
- Full battery: 574 pass, 0 fail, 3 skips (base binary / clang-format 22 absent locally). ASan/UBSan clean on both Ruby fixtures, the deep chain, the ordering pair and a cache round-trip.
- No new gate script; the count stays at 563.

One note: the CHANGELOG's receiver entry on main still says "parser version 83" in its heading after the renumber to 84. The new entry references that rather than editing it.


joyful-ii-V-I:
Landed on `main` as cf38ba75 (via #91) — thank you @andriytyurnikov.

GitHub could not auto-close this one because the branch was rebased onto a main that had moved, so the commits carry new SHAs. All three are still yours:

- `6c8c132b` fix(ruby): a later load-time receiver site clears the lazy bit; the constant-chain check is a loop
- `a640639c` fix(legend): --impact's lazy= legend names the closure, not a language
- `3e1459f7` chore(rebase): re-pin impact and help from the merged tree over #89

I verified each of the three findings independently before landing rather than taking the description — the order-dependence reproduces (two files differing only in the order of two lines gave `ccd="2" lazy_edges="1"` vs `ccd="3"`), the constant-chain recursion faults at 5,000 segments and the loop rewrite is byte-identical across the fixture, a 15-shape receiver matrix and `--deps` over the whole tree, and running your gate against a pre-fix binary gives exactly the six red arms your commit message claims. #91 has the full write-up, including what I changed on the branch and why (`kParserVer` 85→86 with the mirror, two `printf_parity.manifest` rows, and your `COMMANDS.md:938` hand-fix reverted because that line is a dated capture rather than generated).

Gate-first with a reproduction before each fix, and a crash gate sized for the platform that would actually catch it — that is exactly the shape this repo asks for.

joyful-ii-V-I:
A follow-up you might enjoy, @andriytyurnikov — and another thank-you.

Your fix landed in the Ruby importer path: the `importers=` reach and the `lazy=` bit that rides on it. Separately, we ran a profiling round this week that put `PROFILE_SCOPE` through `buildGraph` across six corpora, and the phase computing exactly that reach — `buildGraph/2b`, the transitive include closure — turned out to be **the single largest phase in the whole run on `rails/rails`**.

Ruby's require/autoload graph is dense enough to make that closure an all-pairs reachability: Σ|closure| = **3,994,331** ids over 3,916 files, max closure 1,420. And **38.7 ms of its ~62 ms was one `std::sort`** — 3,916 runs averaging 1,020 elements, sorted only so a membership test could be a `binary_search`. (The `std::unique` after it removed 0 duplicates in all 3,916 calls; an epoch stamp already guaranteed uniqueness.)

Converting that sort to a radix sort, interleaved A/B against the same warm cache:

| on `rails` | before | after | |
| --- | --- | --- | --- |
| `buildGraph/2b` | 65.2 ms | **35.0 ms** | −46%, n=7 interleaved, reps do not overlap |
| `buildGraph` total | 223.0 ms | **186.7 ms** | −16% |
| `--callers=main`, whole run | | | **−7.8% / −8.5%** median, two independent n=21 A/Bs |
| default map, whole run | | | −7.6% / −7.3% median |

Controls flat — `go`, `django`, `rust-analyzer`, a private C++ tree and this repo all within ±1.7%, no regression in either direction.

Two honest caveats. This is on a branch, not yet on `main`, so it is a measurement rather than something you can pull today. And the two pieces of work were independent — the perf round wasn't prompted by your PR — they simply converged on the same path in the same week. The nice part is what that means together: the Ruby answer you made **correct** is also, on a Rails-sized tree, about 8% cheaper to compute end to end.

Thanks again for the submission. Three findings, each reproduced before it was fixed, a determinism bug caught by reasoning about statement order, and a crash gate sized for the platform that would actually catch it — that is a genuinely high bar, and it made the tool better.

PR #93, opened 2026-09-10
feat(infra): vendor timsort as the third fastSort algorithm — with the measurement that refuses it attached

joyful-ii-V-I:
The 2026-09-10 round measured timsort against `std::sort`, radix, pdqsort, an `is_sorted` guard and
`std::stable_sort` on the real captured id sets of three call sites across seven corpora, and
recommended vendoring nothing. **It is vendored anyway, on toolbox-parity grounds and not on
performance** — the portable `src/infra/` layer is maintained in a private C++ tree whose `fastSort.h`
documents three algorithms, and the third should be present here with its refutation attached rather
than absent with no explanation. **No call site is routed to it.**

### The version discrepancy, resolved before anything was vendored

Two claims were on the table and could not both describe one file: that upstream has no caller-owned
workspace, and that the private tree's copy — reading `GFX_TIMSORT_VERSION_MAJOR 3 / MINOR 0 / PATCH 0`
— has one. Upstream was fetched and every tag read:

- `timsort_workspace` exists in **none** of upstream's thirteen tags, nor on its default branch.
- Upstream shipped **v3.0.1 without bumping `GFX_TIMSORT_VERSION_PATCH`** — it reads `0` at the v3.0.0
  tag and `0` at the v3.0.1 tag. The macro identifies neither release, which is how `3/0/0` came to be
  read as v3.0.0. By content (single-template-parameter `TimSort`, `std::iter_difference_t`, the fourth
  `rotateRight` call site) the private tree's copy is **v3.0.1**.

So the workspace is a **local patch on v3.0.1**. What lands here is public upstream v3.0.1 verbatim with
that patch re-derived against the public tag — byte-identical to the copy that prompted the question.
MIT notice intact, header-only, no host dependency (G3).

The patch is the only reason to prefer this file over the release: after one `reserve_for()`, a sort
through the workspace overload performs **zero heap allocations** (G2, the property radixSort.h's
caller-owned scratch buys). A routine bump to a newer upstream release deletes that silently — it still
compiles, still sorts, still passes any test that only checks output. The header records the pin, the
patch and that warning.

### The gate, written first

`test/timsortcheck.sh` ran RED before the header existed (arm A). The zero-allocation claim is
falsifiable three ways rather than asserted:

| arm | asserts |
| --- | --- |
| E | 0 allocations across 200 workspace sorts of 4,096 records after one `reserve_for` |
| F | the **owning** entry point allocates on the same input (2,400) — a dead counter cannot read as success |
| G | rebinding the workspace ctor to its own vectors (what a naive re-vendor produces) turns arm E red — it does, 2,400 |

Arm G caught itself being vacuous on first run: a quoted `#include` resolves against the including
file's own directory before any `-I`, so a scratch copy of `timsort.hpp` alone was invisible and the
control reported a perfect zero. Both layer headers now go into the scratch dir and `src/infra` is kept
off that build's include path.

Arms B/C/D: self-contained compile against the layer's include path alone with warnings-as-errors; a
differential vs `std::stable_sort` over 8 shapes x 15 sizes x 5 seeds on index-tagged records, so
**stability** is checked and not just ordering; byte-identical output across runs. Arms H/I/J keep the
prose honest — the provenance record must be in the header, the facade's "no call site" claim is checked
against the tree rather than trusted, and every ratio the facade quotes must also appear in
`bench/PROFILE.md`.

Harness verified clean under `-fsanitize=address,undefined -fno-sanitize-recover=all`, with the
zero-allocation result holding there too.

### The facade

`fastSort.h`'s header comment is now the single place the three-way choice is documented, in the order
the questions decide: **magnitude** (outside one site every measured sort costs at most 0.5 ms per run),
then **descents per element** (0.35x for a three-line `is_sorted` guard where timsort manages 0.58x;
1.71x, and 6.0x slower than radix, where the input is scattered — and the covariate is descents, *not*
n: two sites three orders of magnitude apart in volume sit within 2x in every arm's ratio), then
**payload width**.

`radixSort.h`'s "beats timsort 3-6x on random float keys" was unfalsifiable here — the routine it named
was not vendored at the time — and is replaced by the two-sided measured statement now that it is.

### Housekeeping

- `test/binoverridecheck.sh` gains the pinned exemption the new gate needs (it compiles harnesses and
  never invokes the binary).
- `test/ripwirepubliccheck.sh`'s email allowlist gains `src/infra/timsort.hpp` **by exact path** — the
  one vendored file deliberately outside `third_party/`, because the portable layer is copied wholesale
  and this sorter travels with it; its MIT notice names its upstream authors and the licence requires
  keeping it. `src/` at large stays covered.
- `--quality-delta` gating rows acked with reasoning: the complexity/verbosity/params/duplication rows
  are upstream's own shape (restructuring them would destroy the ability to re-derive the file from a
  public tag plus one described patch), and the `stable` vs `unstable` 22-token twin is what a facade
  *is*.
- Gate count **571 -> 572** across all eight published sites, derived from this tip's own loop.


joyful-ii-V-I:
Coordinator note — I pushed one commit here and cancelled the in-flight run. Count only; **the placement question is untouched and still yours to decide.**

**Why it needed intervening on.** This branch's tip said **572** and `main` also says **572**, and both were right about themselves: main's is 571 + `rustanccheck` (#96, landed while this lane was green), this branch's is 571 + `timsortcheck`. The union is **573**.

The check this lane was going to merge on — *"my published count equals the count derived from my own tip's loop"* — **passes here and is still wrong.** That form can only catch a lane disagreeing with itself, never a lane disagreeing with a main that moved underneath it. Comparing against main is what catches it, and that is the check I have been running before every landing tonight.

It would also have landed silently. Only `test/regression.sh` conflicted, because both sides edited the same `for _g in` line. `README.md`, `docs/EVALS.md` and `present/deck5_ripwire_build.js` **auto-merged clean at 572** — both sides had written the identical token, so the merge had nothing to flag and all three would have shown as untouched in the diff. This is the fifth time tonight that exact shape has come up; the union of the two loops has been the answer every time.

`bench/PROFILE.md` also conflicted: main's loop-hoist ledger and this lane's timsort ledger were appended at the same point. Both kept, in landing order.

**On the `src/infra/` vs `third_party/` question** — you were right to refuse a placement change you had not reviewed, and right that a background task's reasoning is not the owner's approval. That push was mine, and my own brief caused it: I wrote *"vendor into `src/infra/` … in the same shape pdqsort is vendored"*, and `pdqsort.hpp` is in `third_party/`. Those two halves point opposite ways and you resolved the ambiguity reasonably.

I have not re-applied it. It is preserved on `proposal/timsort-third-party-2026-09-10` and the owner decides in the morning. For that decision, the case is: `pdqsort.hpp` is already in `third_party/`; `ripwirepubliccheck`'s own comment names `third_party/**` as the home for vendored code carrying upstream author emails, which is why an exact-path allowlist was needed here; and `infraportcheck` rule A line 14 says outright *"A sibling in the layer is fine; so is a `third_party/` header"*, with arm A defining the vendorable set as *"layer + third_party/"* — so a `third_party/` header already travels with the portable layer, which was the main argument for `src/infra/`. `third_party` is on the include path, so `#include "timsort.hpp"` is unchanged either way.

Your finding about the version macro is the most valuable thing in this lane and should survive wherever the file ends up: upstream shipped **v3.0.1 without bumping `GFX_TIMSORT_VERSION_PATCH`**, so a `3/0/0` header identifies neither release, and the workspace exists at **no** upstream tag. Re-deriving the patch from the public tag and getting a byte-identical result is exactly the right way to have established that.

joyful-ii-V-I:
Coordinator note #2 — this lane is now **stacked on `lane/calls-rank-order-2026-09-10`**, and the gate count is **575**. Content and placement are untouched; this is base and count only.

**Why stacked rather than rebased onto main.** Three lanes went green within an hour and every one of them adds a gate, so only one could hold 573. Rebasing each onto main in turn costs two more rebases and two more 26-job matrices, and each landing re-triggers the next — a treadmill. Stacking fixes the order once:

```
optremarks (573, landed)  →  calls-rank (574)  →  this lane (575)
```

Each tip contains its predecessors, so no further rebase is owed and nothing needs re-running after each merge.

**The union, fourth time tonight it has been the answer:**

| | loop | adds |
| --- | --- | --- |
| main side | 574 | `callsrankordercheck`, `optremarkshotcheck` |
| this lane | 573 | `timsortcheck` |
| **union** | **575** | |

Only `test/regression.sh` conflicted. `README.md`, `docs/EVALS.md` and `present/deck5_ripwire_build.js` auto-merged clean at a stale number — both sides had written the identical token, so the merge had nothing to flag and all three would have shown as untouched in the diff. All eight published sites re-derived from the merged loop; the `8,574` byte figures, `1,574 files` and the 9544 table are deliberately left alone because they are not gate counts. `manifestcheck` is green at 575, and its own mutation control confirms it compares against the runner's loop rather than itself.

**To be explicit about the earlier friction:** you were right both times. A background task's reasoning is not the owner's approval, and you were right to reset a placement change you had not reviewed. The push was mine, as coordinator, and my own brief caused it — I wrote *"vendor into `src/infra/` … in the same shape pdqsort is vendored"* when `pdqsort.hpp` is in `third_party/`. Those halves point opposite ways.

Your verification of the first count bump was the right response to it: you checked rather than trusted, and reached the same conclusion independently. That is exactly what should happen when an unexpected commit appears on a lane you own.

**Still open for the owner, not for me:** `src/infra/` vs `third_party/`, preserved on `proposal/timsort-third-party-2026-09-10`. And your version-macro finding should survive wherever the file lands — upstream shipped **v3.0.1 without bumping `GFX_TIMSORT_VERSION_PATCH`**, so a `3/0/0` header identifies neither release, and `timsort_workspace` exists at **no** upstream tag. Re-deriving the patch from the public tag and getting a byte-identical result is how that was established, and it is the single most valuable thing in this lane.

PR #101, opened 2026-09-10
fix(config): respect CLAUDE_CONFIG_DIR instead of hardcoding ~/.claude

s0undt3ch:
## Problem

Every path that resolved Claude Code's config directory was hardcoded to `$HOME/.claude`, ignoring the `CLAUDE_CONFIG_DIR` environment variable that Claude Code itself honours when relocating its config.

## Fix

Six sites updated across the shell scripts and C++ source:

| File | What changed |
|------|-------------|
| `install.sh` | Detection check and receipt message |
| `scripts/install.sh` | Same two sites in the release installer |
| `skills/install.sh` | `settings.json` path (--hook) and the `claude)` destination branch |
| `src/main.cpp` | `--scan-skills` default skill-home enumeration |
| `src/wrap.h` | `agentDetector()` — new `claude` branch (mirrors the `opencode` branch) checks `CLAUDE_CONFIG_DIR` before falling back to `$HOME/.claude`; `skillsRoot` display string updated to show the env-var form |
| `src/cli.h` | `--scan-skills` help text updated to show `${CLAUDE_CONFIG_DIR:-~/.claude}` |

Follows the same pattern already used for `CODEX_HOME` and `AGENTS_HOME` in the same files.

---

https://claude.ai/code/session_01TXSxdjhzx7tgQpBpjsiCWy

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -43,9 +43,9 @@ cmake --install build-install --prefix "$prefix" --component ripwire
 # meaning: Codex skills plus its advisory hooks.
 skillsInstaller="$prefix/share/ripwire/skills/install.sh"
 if [ -z "${RIPWIRE_NO_ACTIVATE:-}" ] && [ -f "$skillsInstaller" ]; then
-    if [ -d "$HOME/.claude" ]; then
+    if [ -d "${CLAUDE_CONFIG_DIR:-$HOME/.claude}" ]; then
         if bash "$skillsInstaller" >/dev/null 2>&1; then
-            echo "install.sh: activated the ripwire skills for Claude Code ($HOME/.claude/skills)"
+            echo "install.sh: activated the ripwire skills for Claude Code (${CLAUDE_CONFIG_DIR:-$HOME/.claude}/skills)"
         else
             echo "install.sh: could not activate the Claude Code skills; run: bash \"$skillsInstaller\"" >&2
         fi
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -237,9 +237,9 @@ if [ -d "$extractedDir/skills" ]; then
     # provision agent homes later. test/releaseinstallcheck.sh arms (E1)-(E6) pin all of it.
     activated=0
     if [ -z "${RIPWIRE_NO_ACTIVATE:-}" ]; then
-        if [ -d "$HOME/.claude" ]; then
+        if [ -d "${CLAUDE_CONFIG_DIR:-$HOME/.claude}" ]; then
             if bash "$skillsShareDir/install.sh" >/dev/null 2>&1; then
-                echo "install.sh: activated the ripwire skills for Claude Code ($HOME/.claude/skills)"
+                echo "install.sh: activated the ripwire skills for Claude Code (${CLAUDE_CONFIG_DIR:-$HOME/.claude}/skills)"
                 activated=$(( activated + 1 ))
             else
                 echo "install.sh: could not activate the Claude Code skills; run: bash \"$skillsShareDir/install.sh\"" >&2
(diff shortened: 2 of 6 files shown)


joyful-ii-V-I:
Landed on `main` as 92578461 — thank you, this was a real defect and framing it against the existing
`CODEX_HOME` / `AGENTS_HOME` handling was the right way in. Your commit is preserved as `4ee920e5`.

**Your six sites were correct, but not the whole census — and that is the useful part of this review
for you.** I scanned every executable path in the tree that *builds* a Claude config path (the three
shell installers, `hooks/`, and all of `src/`) and found a seventh: `claudeInspect()` in
`src/codexdoctor.h:293`, which constructs **both** `<claudeHome>/skills` and
`<claudeHome>/settings.json` from `$HOME/.claude`.

That is the worst possible place for the fix to stop, so I measured it on your branch as submitted.
With `CLAUDE_CONFIG_DIR` set, `skills/install.sh` deploys into the relocated home and reports success
— and then `ripwire --doctor --agent=claude` says:

```
claude-skills ok="0" manifest="0" declared="0" live="0"
              hint="run bash skills/install.sh --claude to restore exact manifest parity"
claude-hooks  ok="0" configured="0" nudge_refs="0" route_hook="0"
```

A hint telling the user to re-run the installer that had just worked, with every symptom pointing at
the one component that was right. Fixed with the same `envOr()` shape `CODEX_HOME`/`AGENTS_HOME` use
two functions below it.

**Three things I verified that you may have wondered about, all of which came out in your favour:**

- `.claude` also appears in `src/ingest.h` and `src/mcpindex.h`. Those are the crawler's
  directory-**name** denylist — a basename not to descend into while walking someone's repo, never a
  path being built — so leaving them alone was right. They are now pinned as the only two exemptions.
- Your empty-string choice is correct and is now gated: `${VAR:-...}` in shell and `p && *p` /
  `envOr` in C++ all treat `CLAUDE_CONFIG_DIR=""` as unset, so the two languages agree.
- Putting the env-var form in `skillsRoot` while leaving the `homeDir` column literal matches the
  `${AGENTS_HOME:-}` / `${HERMES_HOME:-}` rows exactly (and `opencode`'s `homeDir` is literal despite
  XDG relocation). `configDirTpl` is never displayed, so nothing user-facing disagrees.

**Two things I changed on your branch**, each in its own commit so the diff is readable:

1. `65d98b76` — `agentDetector()`. Hand-writing the second exception took it from complexity 13 to 19
   against this repo's bar of 15, and `--quality-delta` exits 2 on it. That is not a new rule: the
   function's own comment records the same refusal at 19 → 24 for the version before it. Your `claude`
   branch and the existing `opencode` branch are the same shape, so both now go through one
   `relocatableRootDetector( envVar, fallbackRoot, suffix, alternate )`. Behaviour is identical either
   side.
2. `0a41435a` — the gate, plus the consequences your `src/cli.h` help edit set off: registration in
   `test/regression.sh`, the `help_all` row in `test/printf_parity.manifest`, a `docs/COMMANDS.md`
   regen, and the README installer line brought into line with the two below it.

The gate is `test/claudeconfigdircheck.sh`, RED-then-GREEN both ways: **12 arms fail on unmodified
main, 5 fail on your branch as submitted** (arm A names `src/codexdoctor.h` by line; all three
`--doctor` arms go red), **16 pass on the landed tip**. Arm A is a census over executable code rather
than a list of the seven known sites, so an eighth is red on the commit that adds it.

One thing I deliberately did not touch: `skills/ripwire-security-scan/SKILL.md` suggests
`ls ~/.claude/mcp.json` as somewhere to look for MCP configs. That is scanner guidance rather than a
path this tool resolves, and it is not where Claude Code keeps MCP config anyway — a separate
question from this one.

The rebase gave new SHAs (main moved eleven times while this was in flight), so GitHub will not
auto-close this. Closing manually — the work is in `92578461`.

joyful-ii-V-I:
Merged as 92578461 (via the landing lane #105, rebased onto current main). Closing manually because the rebase changed the SHAs.

PR #104, opened 2026-09-10
test(gates): the published gate count becomes a build product — two lanes writing the same N+1 auto-merge CLEAN against a loop of N+2

joyful-ii-V-I:
The published gate count was hand-written at eight sites across three files. Two lanes that
each add one gate both write N+1; git auto-merges the **identical** text clean, and the tree
then publishes N+1 against a loop of N+2 — with every existing check green, because "my count
equals my own loop" holds on each branch, "my loop equals main's loop" holds after the merge,
and the member *sets* differ at the same number. It collided seven times in one night on
2026-09-10 and serialised every gate-adding lane.

This makes the number a build product, the way `docs/COMMANDS.md` and `docs/LIMITS.md`
already are.

**`docs/gatecount_build.py`** derives the count from the single `for _g in …; do` loop in
`test/regression.sh` and rewrites all eight sites. `--check` exits 1 on drift, `--root`
retargets a synthetic tree, `--sites` prints the site list.

**The marker.** Each site carries a comment the generator owns — `<!-- gatecount -->` in
markdown (invisible when rendered on GitHub) and `// gatecount` in the deck's JavaScript. A
count claim on an *unmarked* line is a hand-written count and the generator refuses the tree
rather than leave it behind; a marker that has lost its claim is refused too.

**`test/gatecountcheck.sh`**, written before the generator and watched red:

| arm | what it proves |
| --- | --- |
| A | the generator's count matches an independent derivation of the loop |
| B | every marked site states what the loop names right now |
| C | can-go-red on a **synthetic** tree via `--root` (never a probe file in this one — that perturbs the crawl other gates measure); a second added gate reds it again |
| D | a stripped marker is refused in **both** `--check` and write mode |
| E | an unparseable loop is a refusal, not a fabricated number |
| F | the three hand-kept copies of the site list (generator, gate, `manifestcheck`) agree |

**`manifestcheck`'s arms are all kept, none weakened.** The generator is how the sites are
*written*; manifestcheck is what notices one written some other way. Arm (F) closes the gap
those two site lists left — nothing compared them before.

Also proven by hand: editing the deck's `stat(s, "578", …)` spelling by hand reds arm (B) and
naming the site; stripping a marker from `README.md` reds (A) and (B) with both claim lines
listed; regenerating clears each.

**The merge recipe for a gate-adding lane is now:** union the `for _g in …` sets, run
`python3 docs/gatecount_build.py`, done. 577 → 578 in this PR was produced by that command,
not by hand.

Registered in the same commit: the loop, `binoverridecheck`'s pinned exemption list (the gate
binds no ripwire binary, so `(2b)`'s static tell needs no row), `docs/README.md`'s index, the
`CONTRIBUTING.md` submission checklist, and `docs/EVALS.md` §6/§8.

Touches no `src/` — the binary is unchanged, so `--for` / `--pack-task` / the flagless map are
byte-identical by construction.

Local: `gatecountcheck`, `manifestcheck`, `limitstablecheck`, `readmedriftcheck`,
`docscommandscheck`, `deckcheck`, `deckclaimcheck`, `ripwirepubliccheck`, `gateexitcheck`,
`gateabilitycheck`, `pargatescheck`, `loopconservationcheck`, `qschemetripcheck`,
`runhintcheck`, `testmacrocheck`, `affectedcheck` — all green. `binoverridecheck` (runs every
non-exempt gate against a broken stub) is left to CI.

## The motivating case, observed on a real rebase an hour after this PR opened

A lane rebased onto `dc48e1e9`. `test/regression.sh` was the ONLY conflict, and all eight published count sites auto-merged clean to 577 while the resolved loop named 578, with not one conflict marker among them, because main's string and the lane's string were both literally "577 gate scripts". The number was wrong in eight places and the only thing that could see it was `manifestcheck`.

The sites do not conflict *because* both sides agree on a stale value, so "resolve the conflicts and you are done" is exactly the reasoning that ships the wrong number. After this PR the recipe is: union the `for _g in` sets, run `python3 docs/gatecount_build.py`, done. A marker whose claim was deleted is refused (a claim deleted is a claim drifted).

## The one red run, for the record

Shard 2 went red on every platform: `forrankordercheck` arm (3) pins 4% byte bands on ten `--for` reference queries over the live repo, and the new generator's function named `rewrite` became a rank-1 name-exact hit for "compact legend rewrite" (+15%). Renamed to `rewrite_marked_sites`. That arm measures the corpus, not the tool, and two other queries already sit at +3.5% of the band from ordinary growth; filed as a follow-up.


joyful-ii-V-I:
Landed as main 1d11c8dc. Loop membership = the 578 members main had at 6f91fed2 (Dart) plus exactly `gatecountcheck` = 579, and the eight published sites read 579 because the generator wrote them. For every lane behind this: union your `for _g in` set with main's, run `python3 docs/gatecount_build.py`, commit the result. Nobody needs the number.

PR #109, opened 2026-09-10
fix(ci): a declared gate budget is a floor, not a ceiling

joyful-ii-V-I:
## The 900 s macOS timeout is not a hang — it is the smaller of two budgets

`--budget-scale` applies to `DEFAULT_TIMEOUT_SEC` only. That was deliberate: the `GATE_BUDGET_SEC`
entries were each derived from a CI measurement and were meant to stand as declared. Under CI's
`--budget-scale 4` it inverts the table's meaning — the gates it names as **heavy** become the only
gates in the job running on **less** time than an unnamed one.

| gate | in the table? | effective budget in CI |
| --- | --- | --- |
| `xmlwellformed.sh` — pipes one map through `xmllint` | no | 300 × 4 = **1200 s** |
| `crossdirincludecheck.sh` — builds a whole second ripwire from `git HEAD` | yes | **900 s** |

### Measured, on a CI run of `main`

Run `34479806177`, `macos-14` Release shard 2/2:

```
gates=297 pass=292 skip=4 fail=1 wall=3588.6s
=== crossdirincludecheck.sh (rc=124, 900.1s) ===
    L11: TIMEOUT after 900s (declared budget=900s)

slowest:
   900.1s  crossdirincludecheck.sh     <- killed at its cap
   585.8s  xmlwellformed.sh            <- 1200 s available
   345.9s  rootrelcheck.sh             <- 1200 s available
```

`xmlwellformed` at 586 s is the tell. Every gate on that runner ran 6–10× its idle-local wall, and only
the **undeclared** ones had a budget that had moved with it. `crossdirincludecheck`'s own arms all
passed — it was killed for being on a starved machine holding the smaller of the two budgets.

### The change

```python
limit = max( declared, DEFAULT_TIMEOUT_SEC * budget_scale )
```

At scale 1.0 — `regression.sh`, and a bare `pargates.py` run — every declared value still wins, so local
behaviour is byte-identical. No tripwire is loosened below what it is today on the machine it was
measured on. What changes is that the table can no longer buy a gate *less* time than saying nothing
would have.

**15 of the 18 declared entries were on the wrong side of that line**, not one. Only the three at 1200 s
were already safe.

### Gated as a population, not per entry

The arms already in `pargatescheck.sh` each check a declared **number** — and no number was wrong, which
is exactly why the suite could not see this. `git merge` sees text; this invariant is about the
population. The new arms:

- take the table **and** `run()`'s own budget prologue out of `pargates.py` **by `ast`** — never a
  reimplementation, so a copied formula cannot keep passing after the original moves;
- take the scale out of `ci.yml`, so raising `--budget-scale` re-evaluates the invariant instead of
  quietly re-opening the hole;
- assert at every scale CI really passes (plus 1.0) that no declared budget sits below the effective
  default, and that the floor never *lowers* a declared value;
- check `headbinlib.sh`'s 240 s waiter against the **minimum effective** declared budget rather than a
  hand-copied `900`, so the coupling those two files describe to each other can only widen;
- **mutation arm**: revert `run()` to `limit = declared` and the section must go red. The mutation is to
  the *code*, not the table — lowering a table entry proves nothing now, because the floor repairs one
  automatically. That is the point.

```
PASS  floor: ci.yml passes --budget-scale 4 (read from the workflow, not retyped here)
PASS  floor: at --budget-scale 1.0 all 18 declared budgets are >= the 300s an UNDECLARED gate gets
PASS  floor: at --budget-scale 4   all 18 declared budgets are >= the 1200s an UNDECLARED gate gets
PASS  floor: at both scales no declared budget is reduced below its declared value (a floor never lowers)
PASS  floor: headbinlib.sh waits 240s, strictly under the 1200s smallest effective declared budget
PASS  floor(mutation): reverting run() to the pre-floor 'limit = declared' is caught (15 violations)
```

`pargatescheck.sh` 40/40 PASS locally; `manifestcheck.sh` and `gatecountcheck.sh` green.

No gate added, **no gate count change**, no published number moves — so this cannot collide with any lane
in flight.

### DIFF
--- a/test/lib/headbinlib.sh
+++ b/test/lib/headbinlib.sh
@@ -60,6 +60,11 @@ ripwire_head_binary()
     # 240 s here against pargates' 900 s for the six head-binary gates (its `slow` set) leaves 660 s of
     # headroom. If either number moves, move it with the other one: pargates.py's `slow` comment names this
     # file, and this comment names pargates.py, so neither can drift alone unnoticed.
+    #
+    # Since 2026-09-10 pargates spends max(declared, DEFAULT x --budget-scale) on a declared gate, so the
+    # headroom can only GROW from 660 s under CI's scale -- it never shrinks. The invariant to preserve is
+    # still the strict inequality, not the arithmetic difference: this wait must expire with the gate's own
+    # assertions still able to run, so raise this number only alongside the 900 s it is measured against.
     _t=0
     while [ "$_t" -lt 240 ]; do
         [ -x "$_bin" ] && { printf '%s\n' "$_bin"; return 0; }
--- a/test/pargates.py
+++ b/test/pargates.py
@@ -26,7 +26,7 @@ only = None
 jsonout = None
 shard = None          # (k, n): run only the k-th of n deterministic slices of the gate list
 shard_plan = False    # print every slice's membership and predicted weight, run nothing
-budget_scale = 1.0    # multiply the DEFAULT per-gate budget (never the explicit overrides) -- CI passes >1
+budget_scale = 1.0    # scales the DEFAULT budget; a declared override acts as a FLOOR under it -- CI passes >1
 exclude_list = None   # a committed file naming gates this leg does not run (one per line, # comments)
 args = sys.argv[3:]
 for i, a in enumerate(args):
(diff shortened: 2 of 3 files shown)


joyful-ii-V-I:
Post-merge note, for whoever hits an exemption-shaped bug next.

The arithmetic here is the small half. The load-bearing half is that the new arms read **the scale out of `ci.yml`** and **the formula out of `run()` by `ast`**, rather than retyping either into the gate. So the exemption gets re-checked whenever the global it is exempt from moves, instead of aging quietly.

That is the difference between fixing a number and fixing the thing that let the number rot. `GATE_BUDGET_SEC` opting out of `--budget-scale` was *correct the day it was written* — each entry really was a CI measurement, and re-scaling it really would have blunted the tripwire. It became wrong the day `--budget-scale 4` arrived, and nothing connected the two changes. A fix that only moved 900 to 1200 would have been correct today and stale again at the next scale change.

Credit to the F3 session for the generalisation, which is better than anything in the original description:

> An exemption is a claim with a date on it.

Neither instance we found today carried one — this budget table, and a by-exact-path allowance elsewhere that went stale the same way. The gate cannot supply the date, but it can make the claim re-evaluate itself against the thing it is exempt from, which is the next best thing.

One caveat on this PR's own green, so nobody later reads more into it than it holds: the 27/27 run had a **healthy** runner — shard wall 2428.8 s against 3588.6 s on the run that went red. It is a green, not a controlled demonstration that the floor was what saved the gate. The evidence for the floor is the inversion itself (15 of 18 declared entries below the undeclared default) and the mutation arm, not this run.

PR #110, opened 2026-09-10
ci(macos): shard macOS 4 ways like every other leg

joyful-ii-V-I:
## macOS jobs were carrying twice the gates of ubuntu jobs

Every ubuntu leg splits **4 ways**. Both macOS legs split **2**. Nothing about the platform justifies it — it is a config asymmetry that was never revisited after sharding landed in #52.

The result: `macos-14 Release shard 2/2` is the last job to finish on essentially every run. It is the single check that #107 and #108 were both sitting at **26/27** waiting for.

## Measured, not assumed — and for free

`test/pargates.py --shard-plan` prints the plan and exits, so this needed no CI spend to verify. On `a8e652b0`:

| leg | split | gates/shard | **predicted per shard** |
| --- | --- | --- | --- |
| macOS Release | 2-way (today) | 296 / 298 | **4876 s** |
| macOS Release | 4-way (this PR) | 148 / 148 / 151 / 147 | **2438 s** |
| macOS plain | 2-way (today) | 294 / 296 | **4553 s** |
| macOS plain | 4-way (this PR) | 147 / 149 / 147 / 147 | **2276 s** |

That `load` is the real bucket sum from the greedy LPT pack in `shard_plan_for`, not a target — and gates absent from `.github/pargates-shard-weights.json` take the **median** rather than being dropped, so the four equal figures are measured balance across 594 gates, not an average.

Observed job wall on main `a8e652b0`:

```
32m29s  macos-14, plain,   shard 1/2      <- the tail
32m16s  macos-14, Release, shard 2/2      <- the tail
25m33s  ubuntu-24.04, Release, clang, shard 2/4   <- ubuntu's worst
```

Halving the macOS shard load should put macOS near **17m** and make ubuntu the critical path — roughly **7 minutes off every merge**.

## The skiplist needs no change

`--exclude-list` is applied **before** the shard split (`pargates.py` says so, and the plan above confirms it: the macOS plain legs rebalance across 4 buckets with the 4 skipped gates already removed).

## Cost, stated plainly

This adds **4 macOS jobs — 26 checks becomes 30.** macOS runners bill at 10× on GitHub-hosted, though this repo is public. It buys wall-clock on the critical path; it does not make any gate faster and it is not a correctness fix.

## Why no gate

The invariant worth gating is *"no leg's predicted shard load may exceed another's by more than a set factor"* — precisely the asymmetry that rotted here, and the same shape as #109's gate (read the workflow, never retype it).

I did not add it: a new gate moves the published gate count, and **#107 and #108 are both in flight carrying count changes of their own.** That collision costs more today than the gate is worth. It is written into the handoff instead, next to the `--only` refusal.

## Relationship to #109

Independent. #109 fixed the *budget* `crossdirincludecheck` runs under (900 s → 1200 s under CI's scale). This fixes the *load* that made macOS jobs slow in the first place — and halving gates per job also halves the `-j 3` contention that was inflating every gate's wall time on those runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,10 +160,14 @@ jobs:
       # far lower than Linux and 4 macOS shards would queue behind each other instead of running.
       matrix:
         include:
-          - { os: macos-14,     flavor: Release, cc: appleclang, shard: 1, shards: 2 }
-          - { os: macos-14,     flavor: Release, cc: appleclang, shard: 2, shards: 2 }
-          - { os: macos-14,     flavor: plain,   cc: appleclang, shard: 1, shards: 2, skiplist: .github/pargates-macos-plain-skip.txt }
-          - { os: macos-14,     flavor: plain,   cc: appleclang, shard: 2, shards: 2, skiplist: .github/pargates-macos-plain-skip.txt }
+          - { os: macos-14,     flavor: Release, cc: appleclang, shard: 1, shards: 4 }
+          - { os: macos-14,     flavor: Release, cc: appleclang, shard: 2, shards: 4 }
+          - { os: macos-14,     flavor: Release, cc: appleclang, shard: 3, shards: 4 }
+          - { os: macos-14,     flavor: Release, cc: appleclang, shard: 4, shards: 4 }
+          - { os: macos-14,     flavor: plain,   cc: appleclang, shard: 1, shards: 4, skiplist: .github/pargates-macos-plain-skip.txt }
+          - { os: macos-14,     flavor: plain,   cc: appleclang, shard: 2, shards: 4, skiplist: .github/pargates-macos-plain-skip.txt }
+          - { os: macos-14,     flavor: plain,   cc: appleclang, shard: 3, shards: 4, skiplist: .github/pargates-macos-plain-skip.txt }
+          - { os: macos-14,     flavor: plain,   cc: appleclang, shard: 4, shards: 4, skiplist: .github/pargates-macos-plain-skip.txt }
           - { os: ubuntu-24.04, flavor: Release, cc: gcc,   shard: 1, shards: 4 }
           - { os: ubuntu-24.04, flavor: Release, cc: gcc,   shard: 2, shards: 4 }
           - { os: ubuntu-24.04, flavor: Release, cc: gcc,   shard: 3, shards: 4 }


joyful-ii-V-I:
## This is load-bearing, not a wall-clock nicety — evidence from after it was opened

I opened this as a critical-path improvement. New data says it is the actual fix for the `crossdirincludecheck` timeout, and that **#109 was necessary but not sufficient**.

On PR #107, run `34495265793` attempt 1, the gate died **with #109's floor applied** — the message names the fix in its own words:

```
=== crossdirincludecheck.sh (rc=124, 1200.1s) ===
  TIMEOUT after 1200s (declared budget=1200s, declared 900s raised to
                       the scaled default 300s x --budget-scale 4)
```

### The gate is super-linear in contention, so no budget number fixes it

Using `xmlwellformed` as a runner-speed probe (fixed work: one map through `xmllint`):

| probe | `crossdirincludecheck` |
| --- | --- |
| 144.8 s | < 77.5 s |
| 506.0 s | < 402 s |
| 585.8 s | **> 900 s — killed** |
| 663.8 s | **> 1200 s — killed** |

The probe moves **4.6×** across that range while the gate moves **> 15×**. Extrapolating linearly from the probe would have predicted ~355 s and been wrong by more than 3×.

That is the `cmake` configure + build inside the gate (`headbinlib.sh:_headbin_build`): a parallel build degrades far worse under CPU contention than a single `xmllint` pipe does. Credit to the F3 session for isolating the mechanism.

### Which is exactly what this PR reduces

Halving gates-per-macOS-job halves the `-j 3` contention those gates compete under. **Contention is the input the gate is super-linear in**, so halving it helps this gate disproportionately more than any budget raise — and unlike a budget raise, it does not have to be re-guessed the next time a runner draw is slow.

So the ordering is: #109 removed an inversion that was indefensible on its own terms (15 of 18 declared-heavy gates below the undeclared default). This removes the contention that makes the gate slow. Neither is the end state — **the durable fix is that a full `cmake` build should not sit inside a budgeted gate at all**, either cached across the six head-binary gates or moved to a job that expects to build. That is written up, not built.

### The concurrency risk I flagged in the description is refuted

I noted this takes macOS from 4 jobs to 8 and that 9 concurrent macOS jobs were untested — if they queued in waves, the benefit would shrink. Measured on this PR's own run:

```
macOS job start offsets: [8, 8, 8, 9, 9, 9, 10, 10, 10] s   (n=9, spread 2s)
```

All nine start within two seconds. No cap, no waves, no queueing. The concern does not apply.

joyful-ii-V-I:
Landing this from the HEAD-binary session (identity joyful-ii-V-I, lane `lane/headbin-out-of-gate-2026-09-10`). The owner asked for #110 to be landed before the HEAD-binary fix is measured, and approved this merge.

- CI: 30/30 green on `01b9fddd`. No CI-relevant path (`.github/`, `test/pargates.py`, `test/lib/`) has moved on main since this branch's base `a8e652b0`; `git diff --stat a8e652b0 7e04ec85` over those paths is empty.
- **CodeRabbit (ci.yml 159-160):** valid. The matrix comment still says macOS takes 2 shards, and so does the workflow header (lines 9-10). I'm fixing both in the follow-up PR that stages the HEAD comparison binary. That PR edits this same job, so it isn't worth another 30-job run here for a comment.
- **Cost this adds, recomputed with `shard_plan_for`:** head-binary gates per shard are now `[1,2,2,2]` on macOS Release and `[2,3,0,1]` on macOS plain. That takes macOS from 4 to 7 in-gate HEAD builds per run, and the whole matrix from 20 to 23. The follow-up moves that build out of the gates entirely.

PR #115, opened 2026-09-10
fix(since): git log is handed the resolved commit, never the --since string — and a leading '-' reaches no git argv

joyful-ii-V-I:
## Summary

Defense in depth on `--since`, not a reachable exploit today. `resolveSinceScope` proved a revision with `git rev-parse --verify --quiet '<val>^{commit}'`, kept the **raw** value in `SinceScope::revBoundary`, and `sinceLogArgs` spliced that into `git log '<val>..'` as a positional argv entry. `shSingleQuote` stops the shell, not git: git reads a leading `-` as an option whatever the quoting. Nothing in ripwire refused one; git's rev-parse happened to.

The same seam had two live **wrong answers at exit 0** on `b3b70d7a`:

| value | before |
| --- | --- |
| `--since='-17 days ago'` | passed `looksLikeDate`; git received `--before=-17 days ago` / `--since=-17 days ago`; window stamped, exit 0 |
| `--since='^HEAD~3'` | `rev-parse --verify` answers `^<sha>` at rc 0; `window="^HEAD~3" commits="0"`, exit 0 |
| `--since=<branch>` | git log received `'<branch>..'`, the caller's string, not the sha rev-parse had just resolved |

## Change

The fix applies the house rule `gitResolveCommitSha` already states (hand git the resolved sha, refuse a leading `-`, trust only a bare-sha answer) by **calling** it:

- `isBareCommitSha` and `gitResolveCommitSha` move from `src/quality.h` down to `src/gitmine.h`, the same move `popenTrimmed` made on 2026-09-04. `quality.h`'s using-declarations keep every `quality::…` call site (crossref.h, gitIsAncestor, materializeCommitTree) compiling unchanged.
- `resolveSinceScope` refuses a value beginning with `-` before either step; the date step would otherwise hand it to `rev-list`/`log`. A revision now resolves through `gitResolveCommitSha`.
- `sinceLogArgs` hands git `<baselineSha>..`, with `VERIFY( isBareCommitSha( … ) )` on the invariant.
- `SinceScope::revBoundary` is removed.

**Consumers of `revBoundary`:** one, `sinceLogArgs`. Every display surface (`window=` on `--hotspots`/`--cochange`/`--rank-by=churn[-decay]`, the `--hotspots` header comment, `<since rev=>` on `--slice`) prints `cfg.since`, never the struct field. So no raw string needed keeping for display, and output is byte-identical (arm S1).

## Gate (written first) — `test/sincecheck.sh`

| arm | asserts | on `b3b70d7a` | on this branch |
| --- | --- | --- | --- |
| S0 | fixture: a branch and an annotated **tag object** both peel to HEAD~3; names are sha-length | PASS | PASS |
| S1 | a branch and a tag give bytes identical to the sha run on every `kSinceHosts` row, once the echoed value is normalised. Sha-length names mean `est_tokens=` cannot move. | PASS ×10 (fence) | PASS ×10 |
| S2 | through a PATH shim logging git's argv, every window host's log range is `'<sha>..'`, never `'<ref>..'`; plus a liveness control | **FAIL ×4** | PASS ×4 |
| S3 | `-17 days ago` and `--output=<tmp>/PWNED` refuse on every host (exit 1, empty stdout, value named) and appear in **no** git argv | **FAIL ×10** | PASS ×10 |
| S4 | `^HEAD~3` refuses on every host; presence arm confirms git still answers it with a non-sha | **FAIL ×5** | PASS ×5 |

The gate went from exit 1 (57 PASS / 19 FAIL) to exit 0 (76 PASS / 0 FAIL).

**A pre-existing hole in the same gate:** the N4b table-driven rows ran inside `printf '%s\n' "$SINCE_ROWS" | while …`. That loop is a subshell, so a FAIL row set `fail=1` where the exit code never saw it (CONTRIBUTING §2 shape 6). A bash script confirms it: `fail=0` after the pipe loop, `fail=1` after an fd-3 loop. All table loops now read rows with `while … <&3; done 3<<< "$SINCE_ROWS"`. The N4b rows are all PASS on both binaries, so nothing was hiding; the red run exiting 1 is the observed-red proof for the new loop shape.

## `--quality-delta` (vs HEAD)

It exits 2, then 0 after `--quality-ack --ack-only=gating` (+3 rows in `.ripwire_quality_acks`):

- **short-horizon-churn ×2** (`SinceScope`, `resolveSinceScope`; churn=self): the two symbols this change is about, last rewritten by N4 on 2026-09-04.
- **duplication** `gitmine.h::isBareCommitSha | prcontext.h::isCommitSha` (78 tokens): the twin predicate already existed; the move re-keyed it (`was=0`). It is left for its own lane, because `prcontext.h`'s copy is lowercase-only, so consolidating is a deliberate semantic choice on the `--pr-context` path.
- Left unacked, non-gating: **api-surface new-symbol ×2** (minor), which is the moved pair.

## Not in this PR (follow-up)

`src/mergescout.h:resolveCommittish` is a third resolver copy with neither guard. Measured on `b3b70d7a`, `--merge-scout=^HEAD~1` sends `^608656…` into a later git call and prints an empty arm at exit 0, where a bad ref should refuse. `src/prcontext.h:resolveBaseRefSha` has the answer check but no leading-`-` refusal. Both should route through the shared `gitResolveCommitSha` in their own lane, with their own red-first arms.

## Verification (local, targeted — CI carries the full suite)

- Dev build (plain, no build type): clean, 0 warnings.
- `test/sincecheck.sh`: **exit 1 on `b3b70d7a`** (19 FAIL: S2 ×4, S3 ×10, S4 ×5) → **exit 0 on this branch** (76 PASS).
- 30 targeted gates, all exit 0 (235 s at -P 4), chosen as every gate passing `--since` plus every gate reaching the moved helpers or named by `--test-gate`:
  `hotspotsincecheck sincewindowcheck sincecochangecheck churndecaycheck slicediffcheck sliceflowsenscheck modifierguardcheck emptyvaluerefusecheck a9disclosurecheck compactlegendcheck estchargecheck localscountcheck preproccondcheck qualitystalecheck w3fixlegendcheck churnjoincheck qrevtokencheck prrefsafecheck crossrefcheck crossrefdegradecheck mergescoutcheck dmmcheck baselinecheck qextractionkeycheck qschemetripcheck binoverridecheck cochangesurprisecheck fixedbufsweep nestprofilecheck rankbycheck`
- Not run locally, left to CI: the four C++-harness gates `--test-gate` names at 4–5 include hops (`includeprecisecheck`, `rustimportprecisecheck`, `clonebandcheck`, `type3clonecheck`). The main build already compiles every TU that includes the two headers.
- `./build/ripwire . --quality-delta`: exit 0 after the gating ack (2 minor rows, both the moved pair).
- G1: a fresh `-DRIPWIRE_ASAN=ON` build (0 warnings) runs `test/sincecheck.sh` (76 PASS) and `test/qrevtokencheck.sh` (25 PASS) at exit 0. It uses CI's macOS sanitizer options (`halt_on_error=1:abort_on_error=1`, `log_path=`), and no sanitizer log was written. CI's asan job does not run `sincecheck`, hence the local run.


joyful-ii-V-I:
Re the Docstring Coverage pre-merge warning (52% of the functions this diff touches): not adopting. ripwire's C++ has no docstring convention. CONTRIBUTING §3 style puts a `//` block comment above a function, and each C++ function or struct this diff adds or changes carries one: `isBareCommitSha`, `gitResolveCommitSha`, `resolveSinceScope`, `sinceLogArgs`, `SinceScope`. The check does not count those. What's left is the gate's one-line shell helpers in `test/sincecheck.sh` (`sinceHostArgv`, `padToSha`, `normSince`), which sit under the section's block comment. Same disposition as #118. No other findings were raised; the review reported no actionable comments.

PR #116, opened 2026-09-10
fix(merge-scout): refs resolve through gitResolveCommitSha — `^REF` refuses instead of an empty arm at exit 0

joyful-ii-V-I:
## The defect

`--merge-scout` resolved each ref with its own copy of the rev-parse probe (`mergescout.h` `resolveCommittish`) and kept rev-parse's **raw** stdout. That copy had neither half of the house rule `quality::gitResolveCommitSha` enforces.

**Measured on 605cd6c8**, through a PATH shim that logs every git argv entry:

| ref | rev-parse answers | what happened | should be |
| --- | --- | --- | --- |
| `^HEAD~1` | `^e55a20…` at **rc 0** | counted as resolved; `git merge-base` received the argv entry `^e55a20…`; stdout `<merge-scout arms="1" …>` with an empty `ok="0"` arm, stderr `no merge-base for ref (unrelated history?)`, **exit 0** | refusal, exit 1, naming the ref |
| `--output=<tmp>/PWNED` | nothing, rc 1 | refused, but only because git's own rev-parse rejected it: the token still reached git as `[--output=…^{commit}]` | refused before git is asked |

The first row is a wrong answer at exit 0: an input that should get a loud bad-ref refusal (like every other unresolvable ref) got a plausible-looking empty arm instead. The second row is defense in depth. I probed 17 option-shaped spellings on this git and none resolves at rc 0 today.

## The fix

`resolveCommittish` is deleted. `resolveAllRefs` calls `quality::gitResolveCommitSha`, which refuses a leading `-` and trusts only a bare 40/64-hex object name. The stale "Mirrors gitmine.h's resolveSinceScope rev probe" comment goes with it.

The call is spelled `quality::gitResolveCommitSha` on purpose, because the `harden/since-resolved-sha` lane moves the function. At that lane's `ba21f5ef` the body lives in `gitmine.h` and `quality.h:1018` carries `using rw::gitResolveCommitSha;`. `git merge-tree` of this branch against `ba21f5ef` is clean (the lanes share no file). I have not compiled the merged tree.

`prcontext.h` had two comments naming the deleted function. They now name the shared resolver. The one above `resolveBaseRefSha` says plainly that prcontext's copy still leaves the `-` half to git; converging it is out of this PR's scope.

## Gate first: `test/mergescoutcheck.sh`

Before the fix, 8 FAILs on 605cd6c8, all in the new arms (no old arm changed state). After the fix: 61 PASS / 0 FAIL.

[...]

PR #117, opened 2026-09-10
fix(pr-context): BASEREF resolves through the shared gitResolveCommitSha — a leading '-' reaches no git argv

joyful-ii-V-I:
## What this is

Defense in depth, following the `--merge-scout` resolver hardening (#116), which makes `mergescout.h`'s `resolveAllRefs` call `quality::gitResolveCommitSha` directly.

`src/prcontext.h`'s `resolveBaseRefSha` and its private `isCommitSha` were a second hand-rolled copy of the probe that `quality::gitResolveCommitSha` owns: `rev-parse --verify --quiet '<ref>^{commit}'`, with the answer kept only if it is 40/64 hex. The copy held the answer-shape half of that house rule. It did not hold the other half: refuse a ref beginning with `-` **before git is asked**.

`--pr-context=--output=FILE` still reached git as the argv entry `--output=FILE^{commit}`. Only git's own `rev-parse` rejection stopped it. Exit 1 and an untouched victim file were already true before this change, and still are. What changes is who refuses: ripwire now does, and git is never asked.

## The change

- `resolveDiffAnchor` calls `quality::gitResolveCommitSha( root, std::string( baseRef ) )`. `resolveBaseRefSha` and `isCommitSha` are deleted; nothing else used them.
- **Comments.** The P0.1 block and the call site now describe the shared resolver, not the copy.
  - `DiffAnchor::refSha`'s comment called the sha "the ONLY spelling of the ref that reaches a git argv". That was false: the raw ref reaches the resolve probe.
  - It now says the sha is the only spelling any git call **after** that probe receives. The gate asserts exactly that.
- **One predicate difference, unreachable.** `isBareCommitSha` accepts `A-F` as well as `a-f`. `rev-parse` never prints uppercase, and neither alphabet can begin with `-`.

After this change, `grep -rn 'rev-parse --verify' src/` on main finds two remaining copies that take a user ref. This PR leaves both alone:
- `mergescout.h` `resolveCommittish`, which #116 removes
- `gitmine.h` `resolveSinceScope`, which #115 routes through the shared resolver

## Gate first: `test/prrefsafecheck.sh`

A PATH shim logs every git argv entry as `[entry]`, one git call per line. Table rows are read with `while read … <&3; done 3<<< "$ROWS"`.

| row | asserts | on 7e04ec85 (base) | this branch |
| --- | --- | --- | --- |
| shim liveness | a valid ref's `[rev-parse][--verify][--quiet][mainline^{commit}]` probe is logged | PASS | PASS |
| `dash`: `--output=` and `--upload-pack=` single-root, `--output=` multi-root | exit 1; names the ref; no stdout; log non-empty for that very run; payload in no git argv; nothing written | **3 FAIL**: payload in the rev-parse argv | PASS |
| `nonbare`: `^HEAD`, `^mainline` (presence guard: rev-parse answers each at rc 0 with `^<sha>`) | refuses; no `^<sha>` in any argv | PASS | PASS |
| `valid`: `mainline`, single- and multi-root | byte-identical to the unshimmed run; rc 0; the resolved sha reaches git; the raw ref appears only inside its resolve probe | PASS | PASS |

The `nonbare` rows are green on both binaries on purpose. They fence the answer-shape half across the move to the shared predicate.

My first draft of the multi-root `valid` row failed on the base binary for a fixture reason: REPO2 had no `mainline` branch. That run also showed the "raw ref only in its probe" arm passing vacuously on a refused run. The fixture now creates the branch in REPO2, and that arm requires rc 0 plus the resolved sha in the log.

## Verified locally

- **Gates:** `prrefsafecheck` is RED on the base binary as tabled and ALL PASS on the fix. Also ALL PASS:
  - `prcontextcheck`, `pranchorcheck`, `prmaskanchorcheck`, `qrevtokencheck`
  - the tests `--test-gate` names: `fixedbufsweep`, `testrowruncheck`
  - the gate-shape audits: `gateexitcheck`, `manifestcheck`
- **Before/after binaries:** the 7e04ec85 binary and this one give identical stdout, stderr and exit code over this repo (`--no-cache`). Refs tried: `origin/main`, `HEAD~3`, `b3b70d7a`, `nosuchrefzzz`, `^HEAD`, `-p`, `--output=<ghost>`. The ghost path stays absent.
- **`--quality-delta`:** gating=0. One minor row: `short-horizon-churn` on `resolveDiffAnchor`, `churn="ambient"`.
- **`--edit-check=resolveDiffAnchor`:** `unchanged`, 0 incompatible callers.
- **`--test-gate`:** exit 4. That is its contract whenever the untested blast radius is non-empty (here 49 dispatcher symbols). Its three named tests were run, as listed above.
- **Full suite:** left to CI.

## Overlap with #115 and #116 (measured with `git merge-tree`, not inferred)

- **#115** (`harden/since-resolved-sha`, ba21f5ef) moves `gitResolveCommitSha` to `gitmine.h` and keeps `using rw::gitResolveCommitSha;` in `quality.h`.
  - `git merge-tree --write-tree ba21f5ef 9c853f67` is clean.
  - The merged tree contains this call site, that using-declaration and the `gitmine.h` definition, so the `quality::` spelling survives the move.
- **#116** (`harden/mergescout-resolved-sha`, bda65848) **conflicts** with this PR in `src/prcontext.h`, in two hunks:
  1. **The P0.1 comment.** #116 rewords it to name `quality::gitResolveCommitSha`; this PR rewrites it.
  2. **The `isCommitSha` / `resolveBaseRefSha` block.** #116 keeps both functions and updates their comment to name exactly the gap this PR closes; this PR deletes them.
  - **Resolution, whichever lands second:** take this PR's side of both hunks. The call site sits outside the conflict and already calls the shared resolver, so nothing is left referencing the deleted functions.
  - No other file conflicts.


joyful-ii-V-I:
Re the Docstring Coverage pre-merge warning (50% across the 4 functions this diff touches): not adopting it. ripwire's C++ uses `//` comments, not docstrings, and neither CONTRIBUTING.md nor CLAUDE.md sets a docstring convention. This is the same disposition as #118 and #115.

What this diff touches:
- **`resolveDiffAnchor`** (`src/prcontext.h:287`): its behaviour is described by the `//` block that opens at line 252 ("Pick the revision this diff is ANCHORED at…"). That block includes the P0.1 paragraph this PR rewrites. It sits above `struct DiffAnchor`, the type the function returns, not directly above the function.
- **`isCommitSha` and `resolveBaseRefSha`:** deleted, not added.
- **`prctx`** in `test/prrefsafecheck.sh`: has a `#` comment directly above it giving its arguments and outputs.

There were no other findings. Both reviews reported no actionable comments: the first on 9c853f67, the second on 9c853f67..b2fc7983 (the `docs/LIMITS.md` regen).

PR #118, opened 2026-09-10
ci(headbin): build the HEAD comparison binary once per job, before the suite — never inside a gate budget

joyful-ii-V-I:
## The defect

`test/lib/headbinlib.sh` built a second `ripwire` from git HEAD **inside** whichever head-binary gate started first. That put a parallel `cmake` build inside a wall-clock budget, next to `-j 3` other gates. A build like that slows down super-linearly under CPU contention. Here are five draws of `macos-14 Release shard 2/2`, using `xmlwellformed` as the runner-speed probe (fixed work, no build):

| probe | `crossdirincludecheck` | verdict |
| --- | --- | --- |
| 144.8 s | < 77.5 s | pass |
| 352.1 s | not in top-8 | pass |
| 506.0 s | < 402 s | pass |
| 585.8 s | > 900 s | killed at the old budget |
| 663.8 s | > 1200 s | killed with #109's floor applied |

PR #108 run `34514091873` drew it again today: probe 678.6 s, `crossdirincludecheck` killed at 1200.1 s, 296/301 passed. #108 was merged over it.

The probe slowed 4.6× and the gate more than 15×. The budget has already been raised twice and the gate still died. The next slow draw will need a bigger number than this draw can tell you, so **this PR moves the build out of the gates instead of raising the budget again.**

## What changes

- **`ci.yml`:** a new step, *Stage the HEAD comparison binary*, runs after Build and before the gate suite. It builds through `headbinlib.sh` itself, so it uses the same recipe and the same sha-keyed cache as a local run. It then checks the binary's provenance and exports `RIPWIRE_HEADBIN`. It gets the idle runner the Build step gets, and it is outside every gate budget.
- **`headbinlib.sh`, staged mode (`RIPWIRE_HEADBIN` set):** it never builds and never waits. It returns the path only if the file is executable and `--version` says `built_from=<prefix of HEAD's sha>` without `+dirty`. Anything else gets one `headbinlib:` line on stderr and **rc 3**.
- **`headbin_refusal`:** rc 3 **fails** the gate. `pargates.py` counts a SKIP printed past a gate's first 400 bytes as a pass, so the old `|| skip` would have hidden a broken staging pipeline. Any other rc keeps the historic skip. All six callers use it.
- **Unset (a developer's `regression.sh` or local `pargates.py`):** same path as before, with the cache, elected builder, 240 s waiter and private-build fallback. Two small additions: the build now passes `--target ripwire`, which saves the 18 s the probe and doctest binaries took after a 90 s ripwire-only build on the dev machine, and `RIPWIRE_HEADBIN_BUILD_LOG` captures the build output that CI prints on failure.
- **Budgets:** no numbers change. The 240 s waiter / 900 s coupling still guards the unstaged path, and pargatescheck's floor arm still passes (240 < 1200).
- **CodeRabbit's finding on #110** (ci.yml 159-160 and the header still said macOS takes 2 shards) is fixed in `c9ec3727`.

## The gate, written first: `test/headbinstagecheck.sh` (commit `6908b7ec`)

Before the fix it printed 19 FAIL rows. One of them was the missing-binary arm, which hit its 120 s watchdog because the old library waits on a held lock. After the fix it prints 26 PASS rows. It reads real sources throughout, and every section has a mutation control that must go red:

- **(A) CI.** It parses every workflow. Any job that runs `test/pargates.py`, or a head-binary gate by name, must stage, verify and export `RIPWIRE_HEADBIN` in an **earlier, separate** step after checkout. Mutations: the step deleted, and the step moved after the suite.
- **(B) Library.** It runs the real `headbinlib.sh` on a synthetic git corpus with a logging `cmake` shim first on PATH, so "no build was attempted" is something the gate observes. Seven refusal cases (missing, empty, not executable, other sha, `+dirty`, no stamp, `unknown`) must each return 3 with nothing built. The missing-binary case also must not wait while a builder lock is held. HEAD-stamped binaries are accepted, at 9 and 12 characters. Unset, the library still builds through the shim, fills the log, and hits its cache on the second call. Mutation: with the staged dispatch line removed, the library must build.
- **(C) Callers.** Every call site is derived from `test/*.sh` and must route its failure through `headbin_refusal`. `headbin_refusal` is exercised for real: rc 3 calls `no()` and rc 1 calls `skip()`. One real caller (`pyimportprecisecheck`) runs end to end against a staged binary that doesn't exist, and must produce exactly one FAIL naming `RIPWIRE_HEADBIN` while its 16 fixture arms still PASS. Mutation: that caller with the refusal reverted to a bare skip must be flagged.
- **(D) Runner.** `pargates.py` hands `RIPWIRE_HEADBIN` through to its gates (`gates=1`, not an empty selection).

## Why a step per job, not an artifact job: measured

There were three ways to get the build out of the gates: (1) build once per `(os, flavor, cc)` in an earlier job and hand shards the artifact; (2) hoist the build into a step before `pargates.py` in each job; (3) move the comparison arms into their own job. Here is what decided it:

- **Idle build cost.** The Build step on an idle runner (runs `34507397561` and `34514091873`) took 140–261 s on macOS and 81–277 s on Ubuntu. That is roughly what the new step costs.
- **An artifact job doesn't shorten the critical path.** With `needs:`, every shard waits for the builder job: queue, setup, checkout, configure, build, then upload, which adds up to at least one build. Option 2 adds one build to each shard instead. So option 2's critical path is at most option 1's. Option 1 only saves runner-minutes.
- **An artifact can't use this recipe.** `headbinlib.sh` builds with `-DRIPWIRE_NATIVE=ON`, which bakes in the host CPU's instruction set, and ci.yml says never to do that for a binary that runs somewhere else. Hosted runners vary, so option 1 would have to fork the recipe to a portable build, and the CI HEAD binary would stop matching the local one. The recipe doesn't vary by flavor or cc anyway: `CC`/`CXX` are set only on the Configure step, so headbinlib uses each runner's default compiler. The real key would be per OS, not per `(os, flavor, cc)`.
- **macOS concurrency.** An artifact job would take a macOS runner slot and hold up all 8 macOS shards behind it.
- **Option 3** splits the suite. It is the biggest change and gets the same effect.

**What's left inside a gate once the build is staged** (dev machine, cache warm): `pyimportprecise` 2 s, `crossdirinclude` 2 s, `rustimportprecise` 4 s, `preproccond` 4 s, `nestedimport` 6 s, `tsimportprecise` 11 s. On CI these gates showed `crossdirincludecheck` at 366.6 s (#110, macOS Release 1/4), `rustimportprecisecheck` at 214.3 s (#110, macOS plain 1/4) and `nestedimportcheck` at 318.5 s (#110, Ubuntu plain clang 3/4). Almost all of that was the build.

## Cost, stated plainly

- **Builds per run go from 22 to 24.** On `05f4b892`, 22 shard jobs contain a head-binary gate and build inside it (16 Linux, 4 macOS Release, 2 macOS plain). The step runs in all 24 release shard jobs. Shard membership moves whenever a gate is added; on this tip 23 shards host a head-binary gate and macOS plain 4/4 hosts none, so that one build buys nothing. It is about 3 min on a shard that isn't the critical path. Making the step conditional would need a runtime shard-membership query, which is new harness surface. Not done here.
- **Each shard's wall gains an uncontended build up front and loses the contended one inside the gate phase.** The CI run on this PR measures the net; I'll post it as a comment, normalised to the `xmlwellformed` probe per the house rule, not to job wall.
- **Provenance is checked by sha, not by recipe.** A clean `build/ripwire` at HEAD would also pass `ripwire_headbin_verify`. That's deliberate: output is a pure function of the sha (the determinism contract), so it compares identically. Staged mode refuses a stale binary (another sha), a `+dirty` build, and a missing one.

## Verification (local, dev machine, load average 22–51 from other sessions)

- **Full suite**, on the pre-rebase commit with `RIPWIRE_HEADBIN` staged the way CI stages it: `gates=596 pass=593 skip=3 fail=0 tree_writes=0`, 955 s at `-j 6`. The 3 skips are environment-only and were already there (`argvdiffcheck`, `editchecknotecheck`, `g1freshcheck`). The six callers ran 0.9–15.1 s inside the suite, and `binoverridecheck` 167 s.
- **At this tip (`c9ec3727`, rebased onto `05f4b892`):**
  - I pulled the staging step out of `ci.yml` with a YAML parser and ran it verbatim: rc 0 in 103 s from a cold cache, and it exported `RIPWIRE_HEADBIN`.
  - The seven gates individually, with `RIPWIRE_BIN` set and the binary staged, all pass. `binoverridecheck`: all 556 non-exempt gates went red against the sentinel, 0 false-greens.
  - `headbinstagecheck`, `pargatescheck`, `manifestcheck`, `gatecountcheck` and `gateexitcheck` all pass.
- **Unstaged path.** The six callers also pass with `RIPWIRE_HEADBIN` unset.
- **Negative control.** `crossdirincludecheck`, given a staged binary from the *previous* sha, exits 1 with exactly one FAIL row: `headbinlib: … reports built_from=b3b70d7af, but HEAD is 6029b4bd…`.
- **Gate list.** I compared member sets, not counts: 587 names here against 586 on main, and the only difference is `headbinstagecheck`. `docs/gatecount_build.py` regenerated the count.


joyful-ii-V-I:
## CI on `c9ec3727`, run `34531186264`: 31/31 green. What it shows and what it can't

**Probe first.** `xmlwellformed` ran at 140.6–183.8 s on the five legs that run it (shard 3/4 of each; macOS plain excludes it). That's a moderate draw, not one of the 585–679 s draws that killed the gate, so this run can't show slow-draw survival from timing alone. Two things it does show don't depend on the draw.

**1. No gate built the binary.** None of the six gates that call `ripwire_head_binary` appears in any shard's `slowest:` top 8, so each one finished faster than its shard's 8th-slowest gate. In every shard that bound is below the same runner's *idle* staging build, and a build slowed by contention can't beat an idle one:

| leg, shards 1/2/3/4 | staging step (idle build) | 8th-slowest gate = upper bound on each caller |
| --- | --- | --- |
| macOS Release | 157 / 180 / 145 / 174 s | 93.6 / 80.8 / 67.2 / 71.3 s |
| macOS plain | 149 / 169 / 139 / 243 s | 84.2 / 69.8 / 60.6 / (no caller) |
| ubuntu Release clang | 136 / 192 / 189 / 194 s | 55.9 / 70.3 / 67.8 / 79.0 s |
| ubuntu Release gcc | 199 / 190 / 206 / 171 s | 83.7 / 62.6 / 72.6 / 53.8 s |
| ubuntu plain clang | 139 / 195 / 146 / 195 s | 59.9 / 66.5 / 49.4 / 82.3 s |
| ubuntu plain gcc | 158 / 196 / 190 / 192 s | 72.3 / 64.3 / 72.4 / 70.9 s |

**2. Like for like: same shard, same neighbour, normalised to the probe.** Shard 3/4 holds `nestedimportcheck` next to `xmlwellformed` both in #110's run `34507397561` (before this PR) and in this one:

| leg | before: probe / `nestedimportcheck` | gate ÷ probe | after: probe / `nestedimportcheck` | gate ÷ probe |
| macOS Release | 153.5 / 248.5 s | 1.62 | 161.8 / < 67.2 s | < 0.42 |
| ubuntu Release clang | 125.4 / 217.3 s | 1.73 | 178.3 / < 67.8 s | < 0.39 |
| ubuntu Release gcc | 174.7 / 309.6 s | 1.77 | 169.1 / < 72.6 s | < 0.43 |
| ubuntu plain clang | 204.8 / 318.5 s | 1.56 | 140.6 / < 49.4 s | < 0.36 |
| ubuntu plain gcc | 194.9 / 307.8 s | 1.58 | 183.8 / < 72.4 s | < 0.40 |

Before, the gate cost 1.6–1.8 probe-units, mostly its build. Now it's under 0.43, and that's a loose upper bound. Locally, with the cache warm, the same gate runs in 6 s.

**The staging step itself** took 136–243 s across the 24 shards; the Build step on the same runners took 107–270 s. Every staged binary reports `built_from=a55996deb`, this run's merge commit. The recipe is one per OS, as the PR body says. Read from inside the staging step, all 8 macOS shards built with AppleClang 16 (`std::print`) and all 16 ubuntu shards with GNU 13.3 (`std::format+fputs`), whatever compiler the leg itself uses.

**Skips:** only the environment skips every leg already has (`argvdiffcheck`, `editchecknotecheck`, `formatgatecheck`, `g1freshcheck`, `portablebuildcheck` on Linux, `ripwirepubliccheck` arm 2b, `w3fixlegendcheck` N=3). No shard log contains a `headbinlib:` refusal, a `pre-change build failed` skip, or a staged-binary FAIL.

**`binoverridecheck`** took 309.5 s on macOS Release 2/4 and 193–214 s on the ubuntu 2/4 shards. It re-runs the six callers against a sentinel, and they no longer build inside its nested 300 s budgets either. That shard has no probe, so there's no ratio to give.

**The cost the PR body predicted shows up here:** macOS plain 4/4 staged a binary that no gate in that shard used (243 s, on a non-critical shard).

**CodeRabbit:** no actionable comments. On the pre-merge *Docstring coverage 54%* warning: the touched functions are shell and Python in `test/`. Each new shell function carries the house one-line contract header (`# ripwire_headbin_verify BIN SHA  →  …`), which that heuristic doesn't count, and `test/` has no docstring convention to follow. No change.

**Overlap:** #120 (clang-tidy) also edits `ci.yml`; `git merge-tree` of this tip against #120's head is clean.

PR #123, opened 2026-09-10
fix(limits): pin a cap by what it IS, not the line it sits on — a comment above a cap no longer reds CI

joyful-ii-V-I:
## Why

`docs/LIMITS.md` pinned every cap by `file:LINE`, and `test/limitstablecheck.sh` arm (B) compares the committed document byte-for-byte with the generator. So any `src/` edit that only added or removed lines above a cap (a rewritten comment, a deleted helper) went red with no claim in the document changed. On 2026-09-10, #115, #116 and #117 were all red on it. Each of #117's six failed jobs was `fail=1 limitstablecheck.sh`, caused only by `kPrDefaultBudgetTokens` moving from line 452 to 431 of `src/prcontext.h` after an unrelated deletion. Every src PR cut from one main pins the same lines, so each landing re-staled the next.

## What changes

- **`docs/limits_build.py`**: the per-file tables lose the `line` column, and the parameter table's `site` is the file, not `file:line`. The document now says a row is pinned by file, name, value, class and note, and says how to reach a declaration (`grep -n <name> <file>`, or `ripwire . --grep=<name>`, which was checked and does return the declaration line). If one file declares the same name, value and note more than once, it renders **one row marked `×N`**. It is never silently dropped, because a multiplicity is part of the cap set. No such pair exists on main.
- **`--check` is still a full-text compare**, so it stays strict on everything still rendered: value, note, class, each file's `*_capped` disclosure vocabulary, and the cap set itself.
- **`docs/LIMITS.md`** is regenerated. With the line cell normalised out of main's copy, the only difference is one added paragraph, and row order is unchanged.
- **`bench/capsweep/capsweep.py`**: comment only. `render()`'s docstring said line numbers "belong to docs/LIMITS.md". TUNING.md dropped them for this same reason, and the docstring now says both documents do.

## Options weighed

| option | verdict |
| --- | --- |
| keep the line, exclude it from `--check` | rejected. The committed doc would show line numbers that go silently wrong between regenerations, in a document whose claim is "generated and gated". Parallel src PRs would also still conflict on LIMITS.md over nothing. |
| render a stable anchor | rejected. Nothing finer than file + name survives an edit above a cap: a GitHub blob fragment *is* a line number, and an ordinal shifts when a cap is added above. |
| **drop the line** | **chosen**. Each row already sits under its file's heading, and the name is one search away. |

## Gate first: observed RED on the unchanged generator

These are new arms in `test/limitstablecheck.sh`. Arm (I) mutates a scratch copy of the real `src/`, never `src/` itself.

| arm | mutation | before | after |
| --- | --- | --- | --- |
| (I) | 3 lines inserted above `kQueryBudget` (155 → 158) | **RED** | green |
| (I) | a blank line deleted above it (155 → 154) | **RED** | green |
| (I) | all 159 `src/` files shifted down 2 lines | **RED** | green |
| (I) | `kQueryBudget` 50000 → 50001 | RED (STALE) | RED (STALE) |
| (I) | `kQueryBudget`'s declaration deleted | RED (STALE) | RED (STALE) |
| (I) | `loops_capped` renamed away in its file | RED (STALE) | RED (STALE) |
| (I) | a class flipped in a sidecar copy | RED (STALE) | RED (STALE) |
| (J) | `kSynthRowCap` = 7, 7, 9 in one file | **RED** (3 rows) | one `×2` row + one row; deleting one 7 is RED |

How the arms guard against a silent pass:

- Every mutation is checked to have taken (the declaration's line or value is read back from disk) before `--check` is read.
- The target is chosen from the copy by rule (declared once, unclassified, one rendered row, a blank line above it), not hard-coded.
- A red must say `STALE`, so a refusal for some other reason cannot pass for strictness.
- The restored copy must be green again.

Arm (G)'s row regex also loses the line cell.

## Verification (local, macOS, plain build)

- `limitstablecheck`: 16 PASS / 0 FAIL. `gatecountcheck`, `manifestcheck`, `capsweepcheck` and `binoverridecheck` all exit 0. There is no `src/` change, so no binary behaviour moves.
- `--quality-delta` exits 2 on two `short-horizon-churn churn="self"` rows, one per edited `render()`. That is this change's own footprint and is not acked. It also reports one minor verbosity row: `limits_build.py` `render` grows from 82 to 87 LOC because of the new paragraph.
- `--test-gate` exits 4. Its tests to run are `limitstablecheck`, `binoverridecheck` and `capsweepcheck`, all green above. The 30 "untested" symbols are in `bench/*.py` files that reference none of the changed files (grep: 0 hits). They are name-resolved edges, not a real blast radius.

## Merge notes

- **`docs/LIMITS.md` conflicts with #115, #116, #117 and #120**, because each committed a line-pinned regeneration. Resolve by regenerating on the merged tree (`python3 docs/limits_build.py`), never by hand. After this lands, a line shift above a cap changes nothing in LIMITS.md, so this should be the last conflict of this kind.
- **A local, unpushed branch that widens the cap census (`lane/capregister-harness-2026-09-10`) edits the same two files.** A trial merge conflicts in exactly one line of `test/limitstablecheck.sh`, arm (G)'s row regex: keep both changes (no line cell, plus `BOUNDARY`). It also conflicts in `docs/LIMITS.md`, which should be regenerated. `docs/limits_build.py` merges cleanly. That branch's new arm is lettered (H), which is why these arms are (I) and (J). The composition was checked on that trial merge. With the one hunk resolved and LIMITS.md regenerated, `limitstablecheck` passes all 21 rows, including that branch's four (H) rows (206 caps, 7 parameters, 109 classified). The one same-file duplicate that its wider census finds renders as `| kRadixThreshold ×2 | 2048 |` in `src/infra/sortutil.h`.
- **Not fixed here:** the KEY-regex gap that hid `kHandoffSymbolsPerFile` from the register. That branch already fixes it with its own red-proven arm, and fixing it twice would only add a conflict.


joyful-ii-V-I:
CodeRabbit reviewed a8f2d24c (commits 4110fd98..a8f2d24c; a real review, not rate-limited) and found **no actionable comments**, with merge risk rated minimal.

**About its one pre-merge warning, Docstring Coverage 50%:** I'm not changing anything. #115, #117 and #118 answered the same warning the same way: this tree explains code with a block comment above it saying *why*, not with a docstring on every function.

- `docs/limits_build.py`:
  - The reason rows are no longer pinned by line sits in the comment block directly above `pinned()`, which also has a docstring.
  - `render()` had no docstring before this PR either. What it produces is the document itself, and the new paragraph explains the pinning there.
  - `times()` is a one-line `×N` formatter.
- `test/limitstablecheck.sh`: the embedded helpers are two-line read/write/report wrappers. Each arm is explained in the comment block above it, and `mutate()` carries a docstring.

Adding docstrings only to reach a percentage would add text that repeats the code.

Issue #69, opened 2026-09-08
Help wanted — verify ripwire's Hermes integration on a real Hermes


joyful-ii-V-I:
Hermes support is in flight (#51 by @AnkitArya, plus the merged Hermes-format skill from #46 by
@ashutoshsinghpr7). **Nobody maintaining this repo runs Hermes**, so what we ship for it is verified
against documentation and source-reading, not against a running agent.

If you have Hermes:

```bash
bash scripts/verify-agent-integration.sh hermes
```

Phase 1 runs the mechanical half inside a throwaway `HOME` — nothing touches your real config. Phase 2
prints two prompts to paste into Hermes and a markdown block to paste back. **A FAIL is the useful
result.**

## Two claims that are actively disputed, and one that is not

**Disputed — a review of the Hermes CLI source contradicts #51 as written, and we cannot settle it:**

1. **MCP registration.** #51 emits `hermes mcp add ripwire -- <path> --mcp`. Reading the CLI, `mcp add`
   appears to take `--command` plus `--args` with `nargs=argparse.REMAINDER` ("must be the last option"),
   which would make a bare `--` end option parsing and strand `<path>` as an unrecognised positional. If
   that reading is right the correct form is `hermes mcp add ripwire --command <path> --args --mcp`.
   **Which one actually registers a working server?**
2. **Hook slot.** #51 refuses `--hook` for Hermes on the grounds that it has no Claude/Codex-style
   PreToolUse slot. A review found `config.yaml` supporting `hooks: pre_tool_call:` with a matcher and a
   command, JSON on stdin carrying `tool_name`/`tool_input`/`cwd`/`session_id`, and `{"decision":"block"}`
   or exit 2 to block. If that is current, Hermes DOES have a slot and what is missing is our port —
   `hooks/ripwire-nudge.sh` switches on Claude's tool names. The honest refusal message would then be "we
   have not ported the hook yet", not "there is no slot".

**Not disputed, but unrun:** `AGENTS.md` as the context file, `${HERMES_HOME:-~/.hermes}/skills` as the
skills root, and that Hermes has a `terminal` tool and should therefore be CLI-first like Claude, Codex,
opencode and openclaw.

## A seam between the two Hermes PRs, if you want the harder question

#46 (merged) ships a Hermes-format skill at `skills/hermes/<skill>/SKILL.md`. #51's `--hermes` installs
`skills/ripwire-*`, and that glob does not match `skills/hermes/`. So once both land, `--hermes` deploys
[...]

Issue #68, opened 2026-09-08
Help wanted — verify ripwire's openclaw integration on a real openclaw


joyful-ii-V-I:
ripwire ships openclaw support as of dddd5171. **Nobody maintaining this repo runs openclaw**, so every
claim below is verified against docs.openclaw.ai and openclaw's own source, and **none of it has been run
against a live openclaw**. This issue exists so that gap is stated in public rather than discovered by
someone installing.

If you have openclaw, this costs one command:

```bash
bash scripts/verify-agent-integration.sh openclaw
```

Phase 1 is automatic and installs nothing into your real home — it runs the recipe, feeds the installer
the exact flag the recipe printed, and checks the skills land and every symlink resolves, all inside a
throwaway `HOME`. Phase 2 prints two prompts to paste into openclaw and a filled-in markdown block to
paste back here.

**A FAIL is the useful outcome.** It is the only way we learn that a recipe we publish does not work.

## The four claims least likely to survive contact with reality

1. **Context file.** We tell users to paste the wiring into `~/.openclaw/workspace/AGENTS.md` — openclaw's
   own workspace bootstrap file, NOT the repository's `AGENTS.md`. An earlier draft of the row said the
   repo's, which would have sent people to edit a file openclaw never reads. Is the workspace path right?
2. **Skills root and its condition.** We say skills go to `~/.agents/skills` — the same root Codex uses,
   which openclaw's docs call a "compatibility skill root" — and that openclaw reads it **only** when
   `OPENCLAW_STATE_DIR` is the default `~/.openclaw`. We print that condition alongside the path and a
   gate fails if we ever stop printing it. Is the condition stated correctly today?
3. **MCP registration spelling.** We emit
   `openclaw mcp add ripwire --command <path> --arg --mcp`. That is read off `src/cli/mcp-cli.ts` (Commander,
   `--arg <value>` with a required value), **not** run. If `openclaw mcp doctor ripwire --probe` disagrees,
   we are shipping a command that silently registers nothing.
4. **Hook slot.** We deliberately emit **no** hook line for openclaw, on the grounds that its
   `before_tool_call` is an in-process plugin API rather than a shell-command slot like Claude Code's or
   Codex's. If openclaw does have a shell hook slot, we are under-claiming and should port ours.

## What we do NOT need

Not asking anyone to advocate for ripwire, file a PR, or change anything in openclaw. Just: does the
recipe we print actually work? Paste the block, including failures.

csy20:
[...]

PR #131, opened 2026-09-11
docs: three accuracy fixes — the Markdown grammar's attribution row, Hermes's live verification, PageRank's threading

joyful-ii-V-I:
Three places where the docs said something the repository contradicts. Each was checked against its source before editing.

## 1. `THIRD_PARTY.md`: `deps/markdown` gets the attribution row it shipped without

`tree-sitter-markdown` is vendored under `third_party/deps/markdown`. Its sources and LICENSE are there, and `CMakeLists.txt:435` pins it at `f969cd3a` (v0.5.3, block grammar only). But the licence table had no row, so its MIT attribution (Matthias Deiml) appeared nowhere.

**Gate first.** `test/dependencypincheck.sh` gains arm (A'): every directory under `third_party/deps/` must have its `THIRD_PARTY.md` row.
- The list comes from the directory population, not a hand-kept list of names.
- On the tree before this change it was red with exactly one row missing, `markdown`. With the row added, all 9 arms pass.

**Size** is `du -sk`, the way the existing rows were measured (lua 392 KB, json 56 KB and toml 164 KB all match), so 2120 KB is written 2.1 MB.

**Not fixed here.** Arm (A)'s completeness loop still iterates a hand-written list that omits `lua` and `php`.

## 2. `INSTALL.md`: Hermes was run live when it landed; openclaw has not been

`INSTALL.md` (#121) said neither Hermes nor openclaw "has been verified against a real install yet". That is true for openclaw but not for Hermes: #51's own "Verified" section records a live run against a real Hermes by @ashutoshsinghpr7. The installer enabled the shipped skills, and `hermes mcp add` connected and discovered the MCP tools.

The paragraph now says:
- a contributor ran the Hermes installer and MCP registration live when support landed;
- the maintainers have not re-verified it since;
- openclaw has not been verified against a real install.

"Initial support" and the help-wanted issues #68 and #69 stay. `deckcheck` (which scans INSTALL.md) passes.

## 3. `docs/ARCHITECTURE.md`: PageRank's power iteration is single-threaded

The doc said the iteration is "parallelized over fixed contiguous row blocks". `src/pagerank.cpp` is serial: the iteration loop, the dangling-mass fold and the L1 residual run on one thread, and neither `pagerank.h` nor its caller in `graph.h` starts a thread.

The fixed blocks (`kReductionBlockSize = 1024`) exist for determinism, not parallelism. They fix the fold order, so the summation tree is a property of the source; that is the "fixed block partials in canonical order" rule in CLAUDE.md. The five gates that read ARCHITECTURE.md (`agentloopgradercheck`, `cacheidentitycheck`, `elixircheck`, `recallpassagecheck`, `resolverhonestycheck`) all pass.

Findings 1 and 2 came up while preparing the 0.6.0 release notes. Finding 3 came from a read-only probe of memgraph, used as a C++20 test corpus. No code changes, and no gate count or LIMITS.md changes.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -72,8 +72,10 @@ you changed it):
 | openclaw (initial support) | `bash ~/.local/share/ripwire/skills/install.sh --openclaw` | `~/.agents/skills` |
 | Anything else | `bash ~/.local/share/ripwire/skills/install.sh <dir>` | `<dir>` |
 
-Hermes and openclaw support is initial. CI checks what the installers write on disk, but neither has been
-verified against a real Hermes or openclaw install yet. If you use one, the help-wanted issues
+Hermes and openclaw support is initial. CI checks what the installers write on disk. For Hermes, a contributor
+also ran the installer and the MCP registration against a real Hermes install when support landed
+([#51](https://github.com/redhat-et/ripwire/pull/51)), but the maintainers have not re-verified it since. openclaw
+has not been verified against a real install yet. If you use one, the help-wanted issues
 [#69 (Hermes)](https://github.com/redhat-et/ripwire/issues/69) and
 [#68 (openclaw)](https://github.com/redhat-et/ripwire/issues/68) ask for exactly that check.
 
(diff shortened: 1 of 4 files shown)


joyful-ii-V-I:
@coderabbitai review

PR #139, opened 2026-09-11
feat(ruby): a constant argument and a rescue class are dependencies; lazy_edges counts distinct pairs (parser version 89)

andriytyurnikov:
Round three of the Ruby constant work, on the same corpus-own index as #57 (superclass, mixins, autoload) and #65 (constant receivers). #65 pinned constant **arguments** and **rescue classes** as its disclosed floor; this lifts them, gate-first.

## The rule

Ruby's rule is that *evaluating* a constant is what makes the autoloader load its file, and a receiver is only one of the places a constant is evaluated.

- **A constant argument is a directive.** A constant chain that is a direct positional child of an `argument_list`, or the value of a keyword pair written directly in that list: `raise Errors::Boom`, `validates_with Validator`, `delegate :name, to: Helper`, `record.is_a?(User)`, `super(Validator)`, `yield User`. The `argument_list` is the grammar's one node for the arguments of a call (with or without parens), a `super` and a `yield`, so one read covers all three. The lists of `include`/`extend`/`prepend`/`autoload` stay round one's, one record per statement (pinned: `include Helper` + `extend Validator, User` = 3 records, none doubled). An argument is lazy inside a closure and load-time at class-body or file level, exactly like a receiver.
- **A rescue class is a directive, and it is lazy always.** Ruby evaluates a rescue list only while matching an exception — `class X; begin; 1; rescue Nope; end; end` is silent, and the same `begin` with a `raise` inside names `Nope` in a NameError (ruby 4.0.6). So a class-body rescue is a use, not a load-time dependency, and stays out of the ccd/godfiles structure like every other lazy pair.
- **One dedupe key.** Arguments and rescue classes share #65's (file, innermost open, written name) record with receivers; the parser-86 AND rule still decides the lazy bit — a rescue above a class-body receiver of the same name is one load-time directive (mixed.rb).
- **Disclosed floor**, pinned to yield nothing (floor.rb): `when` patterns (eager in Ruby — the next round's first candidate), array/hash-literal elements, splats, assignment RHS, interpolation, binary operands.

## A pre-existing defect the fixture exposed

`<health lazy_edges=>` and a row's `lazy_edges=` are documented as **distinct** (file, target) pairs, but `resolveStructuralIncludeAdj` walked the un-deduped adjacency — which is in directive order, not sorted (`buildPreciseIncludeAdj` sorts only when `dedup=true`) — and counted a pair once per *run* of equal ids. `Errors::Boom`, `User`, `Errors::Bust` in one method resolve to errors.rb, user.rb, errors.rb and read as 3 for 2 pairs. That is my code from #65; #65's own fixture never interleaved two spellings of one file with another target, so its pins were right by shape rather than by the count. The dropped ids are now sorted and uniqued. Split out with a third build (parser 89 with the old count): 1 546 / 1 147 / 6 398 / 2 549 old against 1 381 / 1 015 / 6 342 / 2 519 distinct on the four corpora below, with every other `--deps` byte identical between the two counts on activerecord.

## Measured

`--deps --limit=100000 --no-cache`, parser 88 → 89. Gems are Rails 7.2.3.2; the apps are the same two Rails apps as #57/#65, aggregates only.

| corpus | ccd | nccd | shape | load-time importees | lazy_edges | bytes |
| --- | --- | --- | --- | --- | --- | --- |
| activesupport `lib/` (282) | 15 299 → 15 775 | 7.59 → 7.83 | tangled | 201 → 202 | 935 → 1 381 | 75 988 → 85 530 |
| activerecord `lib/` (395) | 4 325 → 4 657 | 1.44 → 1.55 | vertical | 310 → 310 | 1 023 → 1 015 | 114 763 → 129 011 |
| a Rails app, 4683 files / 3532 `.rb` | 13 172 → 17 798 | 0.32 → 0.43 | horizontal | 414 → 1 015 | 5 630 → 6 342 | 750 915 → 855 559 |
| a second Rails app, 2002 / 1895 `.rb` | 6 543 → 7 577 | 0.34 → 0.39 | horizontal | 200 → 557 | 1 878 → 2 519 | 370 991 → 471 865 |

The importee column is the finding: a class-body `validates_with X` / `delegate … to: X` / `rescue_from X` is where a Rails file names what it loads. No shape moved. `--impact` on two real error classes of the second app: importers 0 → 9 and 0 → 7. The `lazy_edges` column carries both mechanisms above, and activerecord is where the over-count fix outweighs the new pairs. This repo's own (Ruby-free) default map is byte-identical before and after.

## What moved, and why

- `kParserVer` 88 → 89 with the mirror; `Include`'s shape and cache format 18 unchanged. Re-pins with reasons in-file: `test/qschemetrip.hash`, `test/printf_parity.manifest` (`impact` + `help_all`: the `lazy=` help and legend now name the two new closure kinds). `docs/COMMANDS.md` regenerated.
- `test/rubyrecvcheck.sh`: the floor arm inverts (`Boom` now has report.rb as a lazy importer); report.rb 5 → 6 directives, `lazy_edges` 2 → 3, health 9 → 10. `test/rubyrequirecheck.sh`: main.rb's `rescue LoadError` is a shown, out-of-tree row, 12 → 13, cone unchanged.
- Gate count 598 → 599 via `docs/gatecount_build.py`.

## Verification

- `test/rubyargcheck.sh` written first: 22 arms red against the parser-88 binary, presence guard and every control green. Two commits: the red gate, then the feature.
- Green: rubyargcheck, rubyrecvcheck, rubyconstcheck, rubyrequirecheck, rubyscopecheck, rubysettercheck, rubymetricscheck, javarubycheck, deplangscheck, qextractionkeycheck, qschemetripcheck, printffmtparitycheck, gatecountcheck.
- ASan/UBSan (`-fno-sanitize-recover=all`): the four Ruby gates green under `asan/ripwire`, zero sanitizer bytes on stderr; a direct run over the fixture rc=0, empty stderr.
- Determinism ×3, warm == cold and `xmllint` on activerecord; own-tree default map byte-identical to the parser-88 binary.
- **The full battery has not completed on my machine** — killed for memory twice beside other applications. I am rerunning it sequentially now and will post the result here as a comment.

Reviewer pushback I expect, and my answer: (a) a rescue class as always-lazy rather than closure-gated — that is Ruby's evaluation order, measured, and it is what keeps a class-body `begin/rescue` from becoming a structural edge; (b) `when` left as a floor even though Ruby evaluates it eagerly — it is a third node shape (`when` → `pattern`), stated in the gate header as the next round's first candidate rather than folded in unmeasured.


andriytyurnikov:
On CodeRabbit's `docs/COMMANDS.md` finding (lines 942 and 1728): both lines are inside the fenced sample-output blocks, and their source is the recorded capture `docs/captures/COMMANDS_showcase_2026-09-11.md` (line 1002 for the `--impact` legend), not the help-derived text. The help-derived lines — 934 for `--impact`, 1720 for `--deps` — were regenerated from the binary and already name constant arguments and rescue classes. A capture is a dated recording that gets refreshed as a round by `test/showcase_capture.py`; hand-editing one is exactly what was reverted on #91, so I have left it alone here. Not a code change; noting it so the finding has a stated answer rather than a silent skip.

The full battery is still running sequentially on my machine; verdict to follow in this thread.

andriytyurnikov:
Full battery, as promised. `test/regression.sh` run sequentially on the committed tree at `00430204` (the parallel runner was killed for memory twice on this machine, so this is the authoritative list, one gate at a time):

    632 gates: 631 PASS, 1 FAIL — versioncheck.sh
    rubyargcheck.sh                          PASS (in the absorb loop)

The one red was the binary's own git stamp: it had been built at `89cda44c+dirty`, one commit before, and `versioncheck` asserts `--version` carries HEAD's hash. Rebuilt at `00430204` (no source change) and `versioncheck.sh` is green, so the battery is 632/632 on this head. Everything else in the earlier verification record stands: ASan/UBSan clean on the four Ruby gates, determinism ×3, warm == cold, xmllint on activerecord, own-tree default map byte-identical to the parser-88 binary.

joyful-ii-V-I:
Thanks @andriytyurnikov, round three reads beautifully. You wrote `rubyargcheck` red first, measured Ruby's own rescue evaluation order, and caught your own `lazy_edges` run-count with the new fixture. We confirmed the count fix alone on rails and discourse too: every other `--deps` byte is identical.

Three things before it lands:

1. **The call graph moves.** Argument and rescue edges also feed the resolver's include narrow, which reads them as import evidence for bare method calls. On discourse, 1,017 call sites newly bind or narrow. We sampled 20 and 19 are wrong, for example `%w[table tables].include?` → `RequestTracker::RateLimiters::Stack#include?` and `record.update!` → `DiscourseDev::Config#update!`. Some correct edges are lost as well. A minimal repro: `Dev::Config` and `Ledger` each define `update!`, in different directories, and a third file runs `notify(Dev::Config)` and then `record.update!(state: 1)`. On main, `--callers` reports 0 with `declined_calls="2"`. On this branch it reports 2 callers, and both are false. Could these edges stay out of that narrow, so only the dependency verbs see them? The `--deps` and `--impact` growth this PR is for would be unchanged; only the resolver would stop reading the new kinds. If you read it differently, rounds one and two disclosed their `edges=`/`ambiguous=` movement, and this entry should too.
2. **`--deps` legend.** CodeRabbit's second point holds in the source. `packDeps` still defines lazy as a closure or an autoload, so it needs "or rescue class".
3. **Merge.** main added `declinecheck`, so both conflicts are one-liners and the gate count becomes 600. We're happy to push that merge if you like.

0.6.0 is close and we would like this in it. If it helps, we can push all three onto your branch as new commits for you to review: the narrow change (most likely a bit on the include record marking where it came from, so the resolver can skip those, plus a gate arm from the repro above), the legend line and the merge. Or we can wait for yours. Whichever you prefer — just say.

andriytyurnikov:
Taking all three myself — the narrow is my miss (I fenced the default map on Ruby-free trees and never measured the call graph on a Ruby corpus), so the gate arm from your repro comes first, then the include-record bit, the packDeps line, and the merge. I'll disclose the call-graph movement on the four corpora in the CHANGELOG entry either way. Back here when the battery is green on the merged tree.

joyful-ii-V-I:
No rush on the three you're taking — this is just a heads-up about `main` so it doesn't surprise you.

We're landing a round of fixes from a review sweep today, so `main` is moving more than usual and this
branch has picked up two conflicts: `CHANGELOG.md` (an append clash) and `test/printf_parity.manifest`
(a regenerable golden). Your Ruby code merges clean — it's only those two.

We'd rather not have you chasing a moving target, so whichever you prefer:

1. **We resolve those two files for you.** If "Allow edits from maintainers" is on, we push the resolution
   to your branch and nothing else — no rebase, no force-push, no touching your commits.
2. **You merge `main` once more when you're ready**, and we hold the rest of our round behind you so it
   does not happen twice.

Either way we are queuing this ahead of our own PRs, so you are not rebasing around our work. One thing we
will handle on our side: your `kParserVer` 93 collides with a bump in one of our branches, and since yours
has been waiting longer, ours moves to 94.

joyful-ii-V-I:
Took option 1 — `main` is merged in and pushed to your branch as a fast-forward (`f4b81134..76d6b33a`).
No rebase, no force, every commit of yours untouched.

Three things, and one of them is still red:

**The two conflicts.** `CHANGELOG.md` was an append clash; both sides kept, your round-three section under
`[Unreleased]` with the released `[0.6.0]` below it. `test/printf_parity.manifest` was more interesting —
**neither side was correct.** Your `help_all` hash covers your `--impact`/`--deps` help changes, main's
covers the twelve flag rows that #170 moved out of a six-space indent, and the merged binary produces
neither. Re-pinned from the merged build with `UPDATE_GOLDEN_EXPECT=help_all`: exactly one label moved, 41
unchanged.

**`test/rubyargcheck.sh` needed the new reporter contract.** #142 landed on main after you wrote this
branch: `ok()` is now a contract that survives a failed write, and `cond && ok … || no …` is gone tree-wide,
because a failed PASS write made that idiom print FAIL for an arm that passed. `gateexitcheck` enforces it,
so your file went red the moment main merged in — through no fault of yours. Converted to the same shape
`rubyrecvcheck` already uses; same arms, same population, rc 0.

**Still red: `showcasecapturecheck`,** and this one is ours to clear, not yours. Your `src/graph.h` changes
shift line numbers, so the committed capture's pinned seed `src/graph.h:3359` no longer resolves to
`rankGraphTeleport` — the capture genuinely needs regenerating for this branch. Regenerating it here is
blocked by a stale branch of ours leaking into the capture's public-export scrub, which correctly refuses
to write rather than ship an unscrubbed document. We'll prune that and regenerate; nothing for you to do.

`kParserVer` 93 and `test/qschemetrip.hash` carry through unchanged, and `gatecount_build --check` /
`limits_build --check` are both rc 0 on the merged tree. Full local suite is green apart from the capture
above.

joyful-ii-V-I:
@coderabbitai review

joyful-ii-V-I:
@coderabbitai review

joyful-ii-V-I:
@coderabbitai review

joyful-ii-V-I:
Merged. Thank you, @andriytyurnikov.

This is the third round of your Ruby constant work in main. It closes the two gaps #65 left as its disclosed floor: a constant argument and a rescue class are now dependencies.

It was easy to take for the same reasons as the first two rounds:
- The rule comes from Ruby's own evaluation order.
- `rubyargcheck` was red before the feature existed.
- The new floor arms pin, in writing, what still yields nothing, so the next round starts from a known boundary rather than a guess.

Your fixture also caught a defect that predates your work: `lazy_edges=` was counting repeated adjacency entries instead of distinct pairs.

While the PR waited, main moved under it several times, so we merged main in as maintainer commits on your branch:
- The gate list is the union of both sides.
- The gate count and `test/printf_parity.manifest` were regenerated from the merged build.
- The showcase capture was re-recorded, because your `graph.h` change moved a line the capture pins.

The parser version is 93, as promised; #172 takes 94 next.

It is merged, not squashed, so your commits keep your name in the history, and you are credited in the 0.6.1 release notes.

andriytyurnikov:
Closing the loop on CodeRabbit's two outside-diff findings on `src/graph.h` here, both verified on main at `1cf3086e`:

1. **Declined-candidate CSR grows with calls × candidates** (2026-09-11). Closed by #208: one interned copy per distinct candidate list, a uint32 offset ceiling behind `DEGRADED_PATH_ALERT`, `verifyOffsetCsr` beside `verifyCsr`, and `test/declinedlistcheck.sh`. The bound is by construction: a distinct list is a subset of the definitions sharing one name, so the store grows with symbols and narrowing outcomes, never with calls. `declinedlistcheck` 30/30 and `declinecheck` 162/162 green on main before I started; nothing left to do.

2. **The proven-file check retains an anonymous-namespace `helper` for `api.h:helper`** (2026-09-12, after merge). Real, and arm (B) of `decltodefcheck` did not cover it: (B)'s TUs do not include the header, and rule 2 proves the FILE, so a TU that includes `api.h` for its own reasons and defines an internal-linkage overload of `helper` was gathered by name and served — the header-qualified count read 3 where the defining `.cpp`'s read 1. One correction to the framing: a same-signature anonymous `helper` in an including TU makes every unqualified call ambiguous, so the shape that compiles is an overload or a namespace-scope `static`. Fixed in #216: an extracted `internalLinkage` bit (parser version 96, cache 22), kept out of another file's widening and COUNTED in `unproven_defs=` so the residue the (E2) arms assert stays disclosed. Red-first arm (B2) there, with the fixture and the numbers.

Issue #62, opened 2026-09-08
C++: disabled #if 0 code counted as a live call site in --uses/--callers


mariadb-KyleHutchinson:
## Summary

Call-graph extraction counts a call site inside a disabled `#if 0` / `#endif` block as a live call, with no indication it's unreachable. Presumably a consequence of tree-sitter-based extraction having no preprocessor awareness, but it means `--uses`/`--callers` can report dead code as a real caller with no disclosure.

## Environment

- ripwire 0.5.0 (Release, AppleClang 15.0.0.15000309, built_from=bacfa3b7b)
- macOS, arm64
- Test corpus: [mariadb-columnstore-engine](https://github.com/mariadb-corporation/mariadb-columnstore-engine) at commit `db1594e12bb103d345e6359a9ad35624cab4b866`

## Repro

```
git clone https://github.com/mariadb-corporation/mariadb-columnstore-engine
cd mariadb-columnstore-engine
git checkout db1594e12bb103d345e6359a9ad35624cab4b866
ripwire . --uses=storage-manager/src/S3Storage.cpp:putObject
```

## Actual

Output includes:
```xml
<u role="call" p="storage-manager/src/S3Storage.cpp:678" in_id="storage-manager/src/S3Storage.cpp::S3Storage::copyObject"/>
```

`storage-manager/src/S3Storage.cpp` lines 669-679:
```cpp
#if 0
    // no s3-s3 copy yet.  get & put for now.
    ...
    return putObject(data, len, destKey);
#endif
```

Line 678 sits inside this `#if 0` block — it is never compiled, and can never execute. `--uses` reports it identically to a live `role="call"` edge, with no marker distinguishing it from the two genuinely live callers found elsewhere in the same output.

## Expected

Either exclude edges whose source line falls inside an `#if 0`/`#ifdef 0` (or other statically-false preprocessor-disabled) region, or mark them distinctly (e.g. a `disabled="1"` attribute) so a consumer doesn't treat them as real blast radius / callers.

## Why it matters

This is exactly the situation `--impact`/`--callers`/`--uses` are meant to answer for a reviewer before merging a change: "who actually calls this." A disabled-code false positive dilutes that answer with a caller that can never fire, and nothing in the output discloses it.

joyful-ii-V-I:
> **Status:** fixed on `fix/confident-zero-2026-09-08`, which has not landed on `main` yet — so the
> code below is not visible there for the moment. Leaving this open until it merges so you can verify
[...]

PR #172, opened 2026-09-12
fix(ingest): `#if 0` stopped serving calls and went on serving every other role

joyful-ii-V-I:
Issue #62 stopped serving CALL sites from preprocessor-dead ranges. It left every other role serving them.

A `role="write"` inside `#if 0` is a write that cannot compile, and `--uses` counted it:

```
# c.cpp: liveWrite does `o.field = 1;`  ·  deadWrite does `o.field = 2;` inside `#if 0`
<uses of="Owner.field" count="4" … counts_floor="1">     <- truth is 2
```

A count carrying `counts_floor="1"` must never exceed the truth, so this is the contract's own failure
direction, not a cosmetic one.

## Six roles, red first

`test/ppdeadrolescheck.sh` was written and run against the unmodified `main` binary before any code changed:

| role | base | after |
| --- | --- | --- |
| call | PASS — #62 already correct | PASS |
| macro | PASS — a retag of `Call`, so it inherited #62's drop | PASS |
| read / write | **FAIL** | PASS |
| `--uses` floor | **FAIL** `count="4"`, truth 2 | PASS `count="2"` |
| import, `using ns::x;` | PASS | PASS |
| import, `#include` | **FAIL** | PASS |
| extends | **FAIL** | PASS |
| type | **FAIL** | PASS |
| `#else` of `#if 1` / of `#if 0` | **FAIL** both | PASS |
| dead definitions | **FAIL** ×7 | PASS |
| var→type bindings | **FAIL** | PASS |

Two of those were not on the list this started from. **`#include` is a second, independent emitter of
`role="import"`**, living in `captureIncludes` rather than the tags arm, which #62 never touched. And the
P2-D `Foo x;` bindings: a `Bar x;` inside `#if 0` above a live `Foo x;` gave `x.m()` an `amb="1"` and two
`prov="split"` edges, where deleting the block outright resolves it to one. That arm is gated against the
dead-block-deleted file as its control.

## Dead definitions are excluded too, and the cost is measured

A definition inside `#if 0` was still indexed and still a resolution target, so `amb=`, `prov="split"` and
`overloads=` were being minted by code that cannot compile — degrading the trust calibration the tool sells
("read the source if which-target matters"). Excluding them was the more invasive of the two options, so it
had to earn it:

- **llvm-project** (8,861 C-family files, 381,811 symbols): **−14 symbols, +3 edges, −16 declined**.
  `ambiguous`, `unresolved`, `est_tokens`, `extent_suspect_syms` and the unindexed roll-up are all
  unchanged. Every dropped row is a real `#if 0` definition — `LoopSimplify.cpp`'s `verifyLoop`, and the
  five in `Descriptor.cpp` that LLVM itself comments as "not needed… leaving it here for now", whose names
  were minting `overloads="2"` against a live macro.
- **ripwire's own map is byte-identical to base.**
- **User CPU 59.1 s vs 59.4 s** on llvm, interleaved — no regression.

The honest cost: `--grep` is unchanged (text inside `#if 0` is still findable, hit rows byte-identical), but
`--expand=deadType` now refuses with `matched no symbol (did you mean 'deadDep'?)` — a refusal carrying a
suggestion rather than a silent zero, and the old answer there was poor anyway, the dead def's span having
swallowed most of the file.

## One rule, asked once

This does not add five more `continue`s beside the existing one. The question is asked once per file and
answered once per record, as a post-pass over the window each capture driver appended, keyed on the record's
own site byte. A second spelling of "which lines of this file exist" is worse than the usual echo-site
drift, because the copies can disagree about the file itself. The call and import rows that drop are
byte-for-byte the ones that dropped before — a reference is sited at the same `startByte` either way — so
the half that was already correct is undisturbed. Ranges are walked once per file in the parse-pool worker
and ride `PendingParsedFile` to the deferred tags pass, so no C-family header pays for them twice.

## Versions

`kParserVer` 93 → 94 and its `quality.h` mirror. **`kCacheVersion` stays 21**: it is a record-*shape* lever,
and no record gains or loses a field here — only which records are extracted. `loadCache` returns empty on a
"version-or-parserVer mismatch", so the parser bump alone rejects every v0.6.0 blob.
`qschemetrip.hash` re-pinned; `qextractionkeycheck` needed no edit (it reads the live constants:
`scheme=10 cacheVer=21 parserVer=94`).

## A residual, disclosed rather than left to be found

`BindingAlias` (the FFI aliases) carries no site byte, so an `extern "C"` block inside `#if 0` still
contributes its aliases. Giving it one is a record-shape change — a `kCacheVersion` bump this defect does
not earn, since an alias whose target has no live definition mints no edge, and one whose target is live
names a symbol that really is indexed. Stated beside `bindSiteByte` and in the gate header. Likewise the L3
fn-pointer clobber sweep reads its own context state: a clobber inside a dead range can still suppress a
live edge, which *under*-reports — the floor the tool already publishes.

## Verification

617 gates: 615 pass, 2 skip, 0 fail (`tree_writes=0`). Warm == cold proved byte-identical across 6 corpora
with dead ranges through a private cache home, with a guard that a cache record was really written —
otherwise the warm side is a second cold parse and the arm degrades into the determinism gate. Determinism
×3 byte-identical, `xmllint` clean, ASan/UBSan clean on the tree and on the new path, `gatecount_build
--check` and `limits_build --check` green.

_Updated after merging main: #139 took parser version 93 and cache format 21, so this lands on 94 and keeps 21._


joyful-ii-V-I:
@coderabbitai review

Issue #63, opened 2026-09-08
C++: --impact/--callers return empty when qualified by declaring header instead of defining .cpp


mariadb-KyleHutchinson:
## Summary

`--impact`/`--callers` on a C++ method **qualified by the header that declares it** (rather than the `.cpp` that defines it) returns an empty result — `reaches="0"` / `count="0"` — even when the method has real, findable callers. Qualifying the identical symbol by its `.cpp` implementation file instead returns the correct non-empty result.

This looks like the exact failure class `docs/EVALS.md` benchmarks ripwire as having zero occurrences of:

> "Blast-radius calls returning an empty radius for a symbol with real callers: ripwire 0, GitNexus 1" — [`docs/EVALS.md:311`](https://github.com/redhat-et/ripwire/blob/main/docs/EVALS.md#L311)

Flagging in case this is a real regression against that benchmark claim, or the claim needs scoping (e.g. "for symbols queried by their defining file").

## Environment

- ripwire 0.5.0 (Release, AppleClang 15.0.0.15000309, built_from=bacfa3b7b)
- macOS, arm64
- Test corpus: [mariadb-columnstore-engine](https://github.com/mariadb-corporation/mariadb-columnstore-engine) at commit `db1594e12bb103d345e6359a9ad35624cab4b866`

## Repro

```
git clone https://github.com/mariadb-corporation/mariadb-columnstore-engine
cd mariadb-columnstore-engine
git checkout db1594e12bb103d345e6359a9ad35624cab4b866

# Header-qualified: returns empty
ripwire . --impact=storage-manager/src/CloudStorage.h:putObject
ripwire . --impact=storage-manager/src/LocalStorage.h:putObject
ripwire . --callers=storage-manager/src/S3Storage.h:putObject

# .cpp-qualified, same underlying symbol: returns real results
ripwire . --callers=storage-manager/src/S3Storage.cpp:putObject
ripwire . --uses=storage-manager/src/S3Storage.cpp:putObject
```

## Actual

```xml
<impact of="storage-manager/src/CloudStorage.h:putObject" defs="2" reaches="0" importers="5" .../>
<impact of="storage-manager/src/LocalStorage.h:putObject" defs="2" reaches="0" importers="3" .../>
<callers of="storage-manager/src/S3Storage.h:putObject" defs="2" count="0" .../>
```

vs. the same symbol qualified by its `.cpp`:
```xml
[...]

PR #173, opened 2026-09-12
fix(graph): a header selector answered with bodies it could not tie to that header

joyful-ii-V-I:
> **Do not merge before the disclosure lands.** This PR makes a wrong answer stop being wrong, but the
> residue it now drops reaches the reader as a plain, undisclosed zero. A silent zero is the exact defect
> #63 created this widening to kill, so shipping this alone trades one honesty defect for another. The three
> lines that close it are named at the bottom; they touch files another open PR holds.

## The defect

`#63` widened a `file:name` selector to the definitions a header's declarations stand for — right in
principle, wrong in predicate:

```cpp
const bool sameContract = s.name == name && s.scope == d.scope && hasBody( s.id );
```

`Symbol::scope` is the immediately enclosing class or namespace name **with namespaces dropped**
(`src/model.h:343`, already documented at `src/graph.h:1082`). So it equates `a::Store` with `b::Store`, and
for a free function `scope` is `""` and the test degenerates to name-only. Both reproduced on the shipped
v0.6.0 binary:

```
--callers=a/Store.h:putObject  →  count="1"  <s n="callB" p="b/Store.cpp:8"/>     truth: 0
--callers=api.h:helper         →  count="2"  two anon-namespace callers            truth: 0
```

The second is worse than the first: `other.cpp` and `third.cpp` define *different* `helper`s with internal
linkage and never include `api.h`. Neither can be what the header declares. Both counts carry
`counts_floor="1"`, so they exceed the truth in the direction the contract forbids — and the function's own
contract note says exactly this must never happen: *"Name alone would turn `Foo.h:size` into every free
`size` in the repository — an over-count inside an honesty fix, which is strictly worse than the silence it
replaces."*

## Positive proof, not a tighter guess

A candidate definition is kept only when its file is a declaration file, or **includes** one — resolved
path-precisely through the existing `resolvePreciseInclude`, against an index holding only the declaration
files. Never by basename. Unprovable ⇒ not widened, so the count stays a floor.

Cost is `O(includes)` byte lookups plus one resolve per include *of a candidate file* — no whole-corpus
pass. Single-root callers pass `ws == nullptr` and are byte-identical to `buildPreciseIncludeAdj`'s
single-root path; under multi-root the workspace context carries the same-root soundness gate with
`absIndex` deliberately empty, so the disk-shape escape probe misses rather than guesses.

## Blast radius — measured, and one-way

Every `file:name` selector whose selection is all declarations, base vs final:

| corpus | selectors | unchanged | shrank | **grew** |
| ripwire itself | 4322 | 4233 | 89 | **0** |
| a 13k-file C++ tree (3133 sampled) | 3133 | 2954 | 179 | **0** |
| synthetic, 20,202 files / ~100k includes | 201 | 200 | 1 | **0** |

Nothing grew, on any corpus. All 327 lost pairs on the large tree and 128 on ripwire were classified by
hand. The dominant cause is the bug: Bullet wraps classes in `ATTRIBUTE_ALIGNED16`, so members get
`scope=""` and `btBox2dShape.h:getAabb` was answering with `btCapsuleShape`'s and `btTriangleShape`'s
bodies — precisely the dynamic-dispatch claim the contract note says this resolver cannot make.

Warm `--callers` on the widening path is indistinguishable from the bare-name path (0.24 s user both on the
13k tree; 0.11 s both on the 20k synthetic).

## Two honest losses, disclosed rather than discovered later

- **CUDA is invisible to the proof.** `.cu`/`.cuh` are absent from `includeLangOf`'s extension table, so
  `reduceKernels.cu` including `"reduceShared.cuh"` never resolves and a `.cuh` selector now under-counts
  where it used to over-count. The fix is two rows in `src/resolve.h`.
- **The section-header split loses its widening.** `src/ingest.h` declares `astQuery`; the body lives in
  `src/ingest_astquery.h`, a section pasted into `ingest.cpp` that does not include `ingest.h`. **48 of the
  128 lost pairs in this repo are that shape**, so `--callers=src/ingest.h:astQuery` is now a zero. The
  evidence channel that recovers it is co-inclusion — "some file includes both" — which needs a
  whole-corpus resolve and is not in this PR.

A symmetric rule (also accept "the declaration's file includes the definition's file") was implemented and
measured, not shipped: **+5 selectors recovered on the large tree, 0 on ripwire, 0 over-counts
reintroduced** — 0.16%, which does not pay for the extra surface.

## Gate

`test/decltodefcheck.sh`, byte-identical in both runs. (A) and (B) are the defect and fail on main; (C/hdr),
(C/split), (D) are the regression guards — #63's own case, the same case split across `include/` and `src/`
with a relative `../include/api.h`, and decl-plus-def in one header — and must be green on **both** sides.
(E) counts the residue, (F) is a seven-shape mutation control. 23 PASS, ALL PASS.

617 gates: 615 pass, 2 skip, 0 fail. ASan/UBSan clean over the full map, every widening-path selector and the
gate itself. Determinism ×3 byte-identical, xmllint clean, `--quality-delta` exit 0.

## What still has to land

The residue count exists and is gated; the attribute does not, because `src/graph.h` has no emit sites:

- `src/callhierarchy.h` — `std::size_t unprovenDefs` on `CallHierarchyRows`, passed to
  `resolveAllByNameQualified`.
- `src/verbs_navigate.h` — mirror the three `bodyless_defs` sites as `unproven_defs`, without the
  `!wantCallers` gate: the residue is equally meaningful on `--callers`.
- `src/graphlegend.h` — the clause defining it on `kCallHierarchyLegend`, plus registration in
  `legendcoveragecheck` / `attrvocabcheck` and a `docs/COMMANDS.md` regeneration.


joyful-ii-V-I:
The disclosure this PR was blocked on is now folded in (`e8a1d743`), so the merge hold is lifted.

`unproven_defs=` is emitted on `--callers`/`--callees` in all three dialects — XML, JSON and columnar — absent at zero, with a legend clause that ships exactly when the attribute does. New arm (E2) in `decltodefcheck` asserts the rendered answer rather than the internal count: 6 red before, 7 green after.

Two constraints showed up and were obeyed rather than moved. The clause is conditional, not appended to the shared opener — unconditional would have taken `--callers`' legend to ~3822 B against a 3429 B pin, and past the 3179 B figure `graphlegendbudgetcheck`'s ratchet needs to stay red on the old binary. And naming the key in the MCP `tools/list` descriptions hit `mcpmanifestcheck`'s 42,200 B ceiling, whose standing rule is that the ceiling moves for a declared argument's obliged bytes and never for prose — so the prose went and the key stayed, matching the posture `bodyless_defs` already holds there.

Worth flagging for whoever comes next: **main sits at 42,177 B of that 42,200 B ceiling — 23 bytes of headroom.** The next lane that adds a declared MCP argument has to re-anchor it.

Also known and not fixed here: every other `resolveAllByNameQualified` call site still passes `nullptr` for the residue count, so `--uses`, `--impact`, `--expand`, `--edit-check`, `--owners`, `--safe-delete`, `--mentions`, `--whereis`, `--path`, `--connect` and several MCP verbs answer correctly but silently on a `file:name` selector. The wrong answer is fixed everywhere the resolver runs; the disclosure covers the two verbs #63 named. The rest is a follow-up lane.

PR #178, opened 2026-09-12
fix(sidecars): a symlink at a sidecar name turned our own write into someone else's file

joyful-ii-V-I:
Three sidecar writers opened a fixed-name destination with a truncating open that follows a symlink. If a repository carried a symlink at one of those names, ripwire's own write overwrote whatever the link pointed at. The link entry survived, so the target was destroyed rather than replaced.

Reproduced on the v0.6.0 release binary before the fix, on all three:

| sidecar | victim before | victim after | exit | refusal |
| `.ripwire_quality_baseline` | 20 B | 589 B of baseline content | 0 | none |
| `.ripwire_notes` | 20 B | 320 B of notes content | 0 | none |
| `.ripwire_arch_baseline` | 20 B | 98 B of arch content | 0 | none |

After: the target is byte-identical, exit 1, stderr names the refusal, and the user's symlink is left alone. The negative control (no symlink, sidecar written normally as a regular file) passes before and after the fix, which is what keeps "refuse always" from passing.

## The refusal is the open

`src/pathguard.h` (new) holds the one open all three writers share:

```cpp
::open( path, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, 0666 )
```

The kernel refuses a final path component that is a symlink at the moment it resolves the path, so nothing sits between deciding and creating. The first revision of this PR checked with `lstat` and then opened. Review pointed out the gap between the two, so the check moved into the open itself. The writers then write through that descriptor with `writeAllAndClose`, which fails on a short or failed write and on a failed close.

`isSymlink`, the `lstat` predicate, stays for two callers where it decides nothing about writing. One is `src/mcpedit.h`, which refuses before its tmp-then-rename publish and never opens the destination. The other is the ELOOP path, where it only picks the wording of the refusal.

The refusal **text** stays per-site on purpose. mcpedit's rename replaces the link and spares its target, while a sidecar's truncating open follows the link and destroys the target, so a single sentence would be wrong about one of them. `DEGRADED_PATH_ALERT` is per-site for the same kind of reason: the once-per-site dedup cannot then silence the second and third writer inside one long-lived MCP process.

It is not under `src/infra/`: that layer is vendored into a sibling repo, and `infraportcheck` rule (C) forbids naming the host project, which a user-facing refusal has to do.

## The arch baseline now reports a failed write

`archWriteBaseline` wrote through a `FILE*` adopted with `fdopen`. The emitters report no failed write, and `fclose` answers only for its own final flush. An earlier flush that failed, followed by one that succeeded, therefore returned true for a baseline that was not on disk. The writer now builds the sidecar in memory and returns `writeAllAndClose`'s verdict, the shape `notes::writeNotes` and `quality::writeBaseline` already had. The bytes are unchanged: old and new binaries write identical sidecars on four fixtures, the largest a 304-hash `--baseline-update`.

## Canonicalizing the path would make this worse

`realpath()` resolves the link and writes to the target on purpose, which is the outcome being prevented. The fix has to be refusal at the open. This is noted so nobody later "hardens" it that way.

## A fourth writer, deliberately not in this PR

`.ripwire_quality_acks` (`writeAckRecords`) is a fourth writer, and it works differently. It publishes through `atomicWriteFile`, tmp-then-`rename`, so the link entry is replaced and the target is untouched. That is silent data loss (a user's symlink quietly becomes a regular file), not an arbitrary write. A proper fix has to guard the temp file that `atomicWriteFile` writes inside the crawl root, and that function is shared with the cache-dir callers. So it gets its own lane and its own gate, and its fixture needs a real quality finding before the write happens at all.

Everything else was swept and cleared. `--html=`, `--export=` and `--pin-census=` write to destinations the user names. Six more writers target the 0700 cache dir with pid-suffixed names. `editpreview` writes into a private temp root.

## The CWD-relative arch baseline stays where it is

Root-qualifying the arch baseline would move the read half with it. A user whose committed sidecar stopped being found would then see their accepted debt reported again as new violations, with a CI exit 2. That is a breaking change inside a fix, and it needs a migration and a decision. The stale comment claiming the sidecar is "written next to the rules file" is corrected; the code never did that.

## Verification

- `test/sidecarsymlinkcheck.sh`, registered in `test/regression.sh` in the same commit: 69 rows, passing on the dev and ASan builds. It failed first on unmodified main, and on the pre-fix `archWriteBaseline` the structural write-completion arms (e5)/(e6) were the only failures. The byte arm computes the expected sidecar from the hash values itself rather than from a captured build.
- `test/fixedbufsweep.sh` gains a row for the new `char hex[ 17 ]`, because it tracks every fixed buffer handed to `formatTo`.
- ASan/UBSan clean on the normal writes, the symlink refusals, the directory refusal and the write-failure run.
- `gatecount_build --check` and `limits_build --check` rc 0. Full local suite: 617 pass, 2 skip (both need an older reference binary), and one `dispatchordercheck` failure. That gate passed in the previous full run and passes alone 3 of 3. CI on this head is the check.


joyful-ii-V-I:
@coderabbitai review

Issue #158, opened 2026-09-11
The callers answer's `next=` pointer lands on a `--uses` call that cannot list the site


joyful-ii-V-I:
## What is wrong

PR #136 made a declined call visible: `--callers`, `--callees` and `--impact` now carry
`declined_calls="K"`, and the callers answer hands the reader a `next=` pointer — a ready-made
`--uses` call — so the declined call site can actually be seen. On a bare name that pointer lands.
On a narrowed selector it does not: it points at a `--uses` call that cannot list the site, so an
agent following the tool's own pointer arrives at `count="0"`. A pointer that leads nowhere is an
honesty bug, not a cosmetic one.

## The evidence

On the `test/declinefix` fixture, the bare-name path works end to end:

```
--callers=jbody      →  next="--uses=jbody"
--uses=jbody         →  lists java/caller/JavaCaller.java:7
```

Take the same symbol by any narrowed spelling — `file:name`, `@FILE:LINE`, a canonical id, or
`Scope::name` — and `next=` still points at a `--uses` answer that cannot list the site. Five
narrowed spellings were reproduced, with three controls, against PR #136's head `c181d2be`, which is
the commit that merged; `src/` has changed since only in one comment.

The gate already exists: `test/declinecheck.sh`, section (E). The red-first step is one arm.

## Size

**Good first issue.** One pointer derivation shared by two emitters, one conditional legend clause,
and one gate arm. No new gate file, no MCP tool-description change. The fix is fenced by its own
condition: when `declined_calls > 0` and the selector is narrowed, point `next=` at the bare-name
`--uses`, in the CLI's XML and columnar roots, in MCP `find_referencing_symbols`, and in the callers
legend — byte-identical everywhere else.

## Where to start

`prompts/help-wanted/next-uses-bare-name.md` is a self-contained prompt for a coding agent. Every
pointer and every recorded output in it was verified by running the code that merged. It covers why
a pointer landing on `count="0"` is a honesty bug, the reproduction across five narrowed spellings
plus three controls, exact file:line pointers with their symbols for where `next=` is built and
where `declined_calls` is computed, the surfaces that deliberately stay unchanged and why, the
red-first change to `declinecheck` (E), the fix's constraints, the acceptance gates and the traps.
Like every prompt in `prompts/`, **it ends by writing a plan and stopping** — a maintainer agrees
the plan before any code is written.

A sibling kit, `prompts/help-wanted/uses-qualified-selector.md`, covers the separate, older gap
[...]

PR #182, opened 2026-09-12
Fix declined next-pointer callers

antoleod:
## Summary\n- make narrowed declined next-pointer results actionable at callers\n- preserve existing worktree and evidence\n\n## Verification\n- Existing issue #158 resume evidence in build/issue158-resume\n- Commit: 6570f446936b82f46a89a5f3c022f671312d32ea\n\nCloses #158


joyful-ii-V-I:
Thanks, @antoleod, and welcome! This is your first PR here, and you did it exactly the way the kit asks: you claimed #158, posted a plan, waited for the go-ahead, and then built it.

Three things we especially like:

- **One decision, two emitters.** `callHierarchyNextSelector` picks the pointer once. The CLI's XML and columnar roots and MCP `find_referencing_symbols` all read that one result, so they can't drift apart.
- **The legend is honest about the widening.** It switches sentences only when the emitter actually widened, and it says what the wider list includes: call sites bound to other definitions with the same name.
- **The tests can fail, and we checked.** One table drives seven narrowed selectors through the CLI, columnar output and MCP, and checks that `next=` really lists the declined site. The new (G) arm proves the predicate rejects the old narrowed pointer. We also ran your `declinecheck` against a binary built from main: 28 failures, which is your seven selectors times the three CLI checks, plus the seven MCP checks. Every control stayed green, and all 28 pass with your change.

main moved while you worked: #173 added a legend clause on the same `emitTo` line. You merged main back in and resolved the conflict yourself, and that saved a round trip.

To get it through CI I pushed one small commit on top. `src/graphlegend.h` is one of the files CI's formatting check covers, and clang-format would have joined your two legend strings into a single line of about 400 columns. So the strings now live in two named constants and the ternary picks between them. The output bytes are unchanged. It's merged, and thanks again!

PR #186, opened 2026-09-12
fix(dispatchordercheck): the whereis precedence pair compared two reads of a ref namespace every session writes

joyful-ii-V-I:
`dispatchordercheck`'s precedence arm runs each verb pair twice, the winner alone and then both flags, and requires byte-identical stdout. One of its pairs is `--whereis=d2 --field-affinity`.

`--whereis` does not read the fixture's files. It scans every branch of the repository enclosing the crawl root. When the gate runs on `test/queryfix` in place, that repository is the checkout's own `.git`, which every worktree and session in the clone writes to. A branch created between the two runs changed `refs_scanned=`, `hits=` and the rows. The gate then reported "--whereis=d2 did NOT answer" about a verb that answered correctly both times. The suite's tree tripwire cannot see a ref write.

## Ruled out first

- **The binary is not nondeterministic here.** With refs held still, 636 solo/paired pairs produced 0 differences. They covered cold, warm, racy-mtime and touched-between-runs caches, while querycheck and reachcheck looped on the same fixture.
- **No gate writes into `test/queryfix`.**

## Fix

The gate now copies the fixture into a private repository it creates:

- `main` has two commits and `side` points at the first.
- Commit dates are fixed.
- Inherited `GIT_DIR` and hooks are neutralised.
- `TMPDIR` points into the same private directory.

The warm-by-default cache stays on, so cold and warm runs are still compared, but nothing outside the gate can move the refs or the cache it reads. It deliberately does not use `--no-cache`, which would hide a cache-dependent difference instead of ruling one out.

Two new rows check the isolation:

- **Guard:** `--whereis=d2` must scan exactly one ref (`refs_scanned="1"`). If the gate is pointed back at a shared checkout, this row fails.
- **Control:** creating a branch in the private repository must change the bytes, and deleting it must restore them.

## Evidence

Run on a clone while a loop created a branch every 0.7 s:

- The stock gate exits 1, on exactly the `--whereis` pair.
- The fixed gate exits 0, with 79 PASS.
- With the fixture copy reverted, the guard fails.

## Verification

- 13 gates pass (`gates=13 pass=13 skip=0 fail=0`), including dispatchorder, query, reach, racymtime, gateexit, worktreeleak, manifest and binoverride.
- `limits_build --check` and `gatecount_build --check` both exit 0.
- Only the test file changed.

### DIFF
--- a/test/dispatchordercheck.sh
+++ b/test/dispatchordercheck.sh
@@ -30,16 +30,51 @@
 # Adjacent pairs pin every seam in the chain; the two cross-handler cases pin the chain's endpoints
 # against its neighbours in main()'s handler sequence.
 #
-# Fixture: test/queryfix (shared with querycheck.sh / reachcheck.sh; d1 -> d2 -> d3 -> d4).
+# Fixture: test/queryfix (shared with querycheck.sh / reachcheck.sh; d1 -> d2 -> d3 -> d4), run from a COPY
+# committed into a repository this gate owns — see "the fixture owns its refs" below.
 
 set -u
 
 BIN="${1:-${RIPWIRE_BIN:-./build/ripwire}}"   # CA4: this gate ignored $1, so a caller passing a binary
-FIX="${FIX:-test/queryfix}"                    # positionally silently measured build/ripwire instead (trap #20)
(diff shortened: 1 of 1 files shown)


joyful-ii-V-I:
@coderabbitai review

PR #189, opened 2026-09-12
fix(legend): compact --zoom, --tree, churn-decay and map rows carried twelve attributes with no definition

joyful-ii-V-I:
Follows #185. `--legend=compact` strips the legend prose and rebuilds definitions from `kCompactCompletenessTerms`. Twelve more attributes got nothing back, so compact answers printed them with no definition. Each was confirmed on main by running it:

| attribute | where it appears | how it is defined now |
| --- | --- | --- |
| `files=` | `--tree` root | in the root's purpose line (it appears on every answer), next to `files_unlisted=` |
| `symbols=`, `isolated=`, `top_modules=`, `levels_shown=` | `--zoom` root | in the root's purpose line, next to `levels=` |
| `children=` | `<module>` at the depth cut (`--zoom-levels`) | a term printed only when present, qualified to `<module>` |
| `<recent of=>`, `<rc age_d= w=>` | churn-decay map | a present-only term qualified to `<recent>`, because `of=` on the map root is `--around`'s seed |
| `lpin=`, `overloads=`, `prov=` | map rows | present-only row terms; each has one meaning across the tool |

## Gate

`test/compactlegendcheck.sh` gains three kinds of arm:

- **(D13)–(D16):** six rows covering the twelve attributes, all red on main. For example: `FAIL (D13) --zoom (its root's counts): symbols="14" is carried but the compact legend never defines it`.
- **(D17):** a present-only mirror that also checks the element qualification. When a reading was spliced into four real compact legends, it fired on all four.
- **(S), fourth population:** every row field that `serialize.h`'s always-on legend marks as absent by default. On main it is red on `lpin=`, `overloads=` and `prov=`. Under each of three mutations it goes red on exactly the mutated name.

Result: main has 87 PASS / 15 FAIL, this branch has 94 PASS / 0 FAIL, and all 86 existing rows stay green.

## Bytes

- **Pins:** none moved. The (L) loop stays at 4,089 B against its 4,100 B pin. `--zoom` goes from 324 to 394 B and `--tree` from 219 to 238 B, both under 400.
- **Everything else:** 93 of the 96 (U) answers are byte-identical, and the three that changed have identical payloads.
- **Over the nominal size:** four unpinned answers that carry these attributes now exceed 400 B (423–632 B).

## Still undefined under compact

These need a decision about the byte pins, so they are not in this PR:

- **`--communities`' `drill=`/`isolated=`:** its compact prose is already 385 B. The shortest honest pair makes it 434 B, and even a cryptic one makes it 422 B.
- **`--community=0`'s `modules=`/`partition=`:** 371 B would become 414–428 B.
- **`unresolved=` and `--impact`'s counts,** as noted on #185.

## Verification

- `gates=12 pass=12 skip=0 fail=0`.
- Output is deterministic and xmllint is clean.
- `--quality-delta` reports gating=0.

### DIFF
--- a/src/compactlegend.h
+++ b/src/compactlegend.h
@@ -109,8 +109,11 @@ inline constexpr CompactLegendSpec kCompactLegendSpecs[] =
     { "cochange",     "cochange",     "files that change together in git: <pair a= b= together= deg= conf_ab= conf_ba= surprising=>, or for of= <f p= together= conf_rev=>" },
     { "communities",  "communities",  "call-graph modules (Louvain): <community id= size= dir= label=> of <member t= n= p=>" },
     { "community",    "community",    "ONE module id=: <member t= n= p=> ranked members, its <bridge> edges; size= the TRUE count" },
-    { "zoom",         "zoom",         "nested module hierarchy: <module level= id= size= dir= shown= capped=> of <member t= n= p=>; levels= deep" },
-    { "tree",         "tree",         "each file with its top symbols by rank, files by best symbol: <file p= symbols=> of <s t= n=>; files_unlisted= have none" },
+    // zoom and tree (2026-09-12): symbols=/isolated=/top_modules=/levels_shown= and files= ride EVERY answer of their root and
+    // were defined only by the full legend, so they read here beside the levels=/files_unlisted= this line already named.
+    // zoom's line is the (U) --zoom probe's budget: 324 -> 394 B of its 400.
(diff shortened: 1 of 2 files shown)


joyful-ii-V-I:
@coderabbitai review

joyful-ii-V-I:
@coderabbitai review

joyful-ii-V-I:
@coderabbitai review

joyful-ii-V-I:
@coderabbitai review

PR #193, opened 2026-09-12
docs(captures): the naming-calibration demo reprinted the project's own rebrand; those rows are now withheld and counted

joyful-ii-V-I:
The showcase capture's `--naming-calibration` demo mines git history for identifier rename pairs. The largest family it finds is the project's own rebrand. Those rows name a private pre-release identifier, and every capture regenerated since the rename reprinted them. `docs/COMMANDS.md` copied them too. Nothing in the export scrub stopped them.

## The rule, keyed on the public side

`docs/docs_commands_build.py` now defines one predicate, and `test/showcase_capture.py` and the gate both import it. A rename row is part of the rebrand when:
- its old and new identifiers have the same number of `_`-separated parts,
- exactly one part differs,
- and the new part is `ripwire`.

The rule never spells the old name, and it matches the whole family exactly.

Each withheld block becomes one honest line where its first row stood:

`<!-- N rename rows withheld: the project's own rebrand, which names a private pre-release identifier -->`

If a matching row shares a line with other output, the write refuses rather than splitting it.

## What changed

| file | rows withheld |
| --- | --- |
| `COMMANDS_showcase_2026-09-05.md` | 5 |
| `COMMANDS_showcase_2026-09-07` … `-09-11.md` | 6 each |
| `COMMANDS_showcase_2026-09-12.md` (new) | 6 |
| `docs/COMMANDS.md` | 4 |

- **Older captures:** the dated captures are edited in place, only the rows replaced by their disclosure line. The gate checks every capture, and 0604eb42 handled an earlier leak the same way.
- **The new capture and `docs/COMMANDS.md`:** both were regenerated on this branch in a clone holding only `main`, the two `lane/*` branches main's capture already shows, and the public `v*` tags. That keeps local branch names from reaching the document. Its lane names are identical to main's copy.
- **Other occurrences:** the old identifier still appears elsewhere in the tree, in benchmark history, EVALS and code comments. This PR does not touch those.

## Gate

`test/docscommandscheck.sh` arms (E) and (F):
- **Red on main:** 39 rebrand rows found across the documents.
- **Green now:** all 14 documents are clean. Every naming-calibration block's counts add up: shown + withheld + past the display cut = `pairs=`, for example 14 + 6 + 48 = 68.
- **Controls:** a real rename (`copyRange→moveRange`) and four near misses are kept unchanged. A disclosure that is off by one, or a row dropped without a disclosure, turns the gate red.

## Verification

- Gates: `gates=6 pass=6 skip=0 fail=0` (showcasecapture, docscommands, ripwirepublic, namingcalibration, mdembed, readmedrift), plus `argvdiffcheck` on the new capture.


joyful-ii-V-I:
@coderabbitai review

PR #194, opened 2026-09-12
fix(gates): mcpremotecheck hung on a stalled listener and judged a dead one's silence as verdicts

joyful-ii-V-I:
`test/mcpremotecheck.sh` was the last `--listen` gate still on its own HTTP client. #188 moved the other three onto `test/lib/gatehttp.sh`. This gate had the same flaws: requests with no deadline, and an empty answer judged as if it were a payload. Main's gate was run unchanged against a scratch wrapper around the real listener that injected each fault:

| listener fault | main's gate | this branch |
| --- | --- | --- |
| pauses after its first answer | **hung**, killed at 150 s | `FAIL no HTTP answer to POST {"jsonrpc":"2.0","id":0,"method":"initialize",…}: HTTP request timed out after 30 s with 0 response bytes` |
| killed after its first answer | 39 FAILs that judged silence (`returned ''`, `got 000 (expected 403)`, `HTTP payload differs from stdio`) and **3 false PASSes**, among them "the target file is byte-identical after the refused remote edit" | `FAIL no HTTP answer to POST {…}: could not connect … Connection refused` |
| pauses once the port accepts | `FAIL loopback listener failed to start; stderr:` after 54 s | `… the HTTP listener never ANSWERED a request … within 30 s` |
| pauses on the malformed-HTTP probe | **hung** in an unbounded `head -1` | a named timeout FAIL on the next request |

With no fault injected, the gate passes 45/45 on main and on this branch.

## What moved
- **Readiness.** `start_server` now uses the shared `wait`, which returns only once the listener answers a request.
- **Payload checks.** Every request whose verdict is a JSON-RPC body goes through the shared `post` with a 30 s deadline. The slowest real request took 41 ms. The first request with no answer is a named FAIL and stops the gate, as in mcptoolprunecheck.

## What deliberately did not move
- **The 18 framing probes** (Origin, Accept, Content-Type, a 9 MiB body, the bearer token and similar) stay on curl, because each one varies a single header, method or size that `post` cannot express. Each now has a `curl -m` deadline, and a non-zero curl exit or a `000` status counts as a no-answer rather than a verdict.
- **The malformed-HTTP probe** keeps its raw socket, with its read now bounded by `read -t`.
- **#187's SIGPIPE drop arms** are byte-identical. They only inherit the shared readiness wait through `start_server`. That wait already required an answer, so its ceiling is the only change: about 55 s becomes 30 s.

## Verification
- `gates=7 pass=7 skip=0 fail=0` across mcpremote, mcpframehonesty, mcpcontract, mcptoolprune, mcprobust, mcpreadloop and mcpwatcher.
- The change is test-only.

### DIFF
--- a/test/mcpremotecheck.sh
+++ b/test/mcpremotecheck.sh
@@ -18,6 +18,9 @@
 #   hostile-input parity  — the mcpaudit4hardencheck corpus (25-digit start_line, XML-comment-breaking task)
 #                           degrades IDENTICALLY over HTTP — the hardened parser is transport-agnostic.
 #
+# A request nobody answered (a timeout, a refused connect, a close without a response, a short or garbled one) is a
+# named FAIL that stops the gate. It is never judged as a payload, a status or a PASS.
+#
 # Usage:  test/mcpremotecheck.sh   |   RIPWIRE_BIN=asan/ripwire test/mcpremotecheck.sh
 # Exits non-zero on any failure. Every mutation happens under a scratch mktemp dir; test/fixture is never
 # modified. Does NOT edit regression.sh or any other existing test file.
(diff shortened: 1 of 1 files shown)


joyful-ii-V-I:
@coderabbitai review

joyful-ii-V-I:
@coderabbitai review

PR #198, opened 2026-09-12
fix(scip): read typed_range, not just the deprecated range

dpunosevac:
`--scip` silently no-ops against any index `scip-java` produces.

Found this while testing `--scip` against a Java index. The patch is a byproduct of
confirming the diagnosis, so no attachment to it: if it is the wrong shape, or you would
rather fix it differently, just close it.

## What happens

ripwire 0.6.0, scip-java 0.13.1, spring-petclinic @ 818c413:

```
docker run --rm -v "$PWD":/sources --env JVM_VERSION=21 \
  ghcr.io/scip-code/scip-java:latest scip-java index --build-tool=maven
ripwire . --callers=getPet --scip=index.scip --legend=compact

ripwire --scip: index 'index.scip' matched no parsed (file,line) — proceeding name-based
```

stdout is byte-identical to the run without `--scip`.

## Why

`src/scip.h` reads `Occurrence.range` (field 1) only. SCIP deprecated that in favour of the
`typed_range` oneof, fields 8 and 9. Across all 6,677 occurrences in that index the only keys
present are `TypedRange`, `TypedEnclosingRange`, `symbol` and `symbol_roles`. Field 1 never
appears, so every `startLine` stays -1 and nothing joins the (file,line) map.

From `scip.proto`: consumers "SHOULD prefer the typed form when available and fall back to the
`repeated int32` form otherwise".

I isolated it to the encoding before touching any code: hand-built an index with the same
documents, symbols and line numbers using field 1, and that one attaches cleanly on stock
0.6.0, with `ambiguous` dropping 6 to 4.

`test/scipfix/make_index.py` writes `packed_int32(1, range_ints)`, the same field the reader
expects, so `scipcheck.sh` exercises the reader against its own generator and cannot catch
this. Happy to add a `typed_range` arm to the fixture if you want one.

## The change

One helper reading field 1 of either range submessage (`SingleLineRange.line`,
`MultiLineRange.start_line`), two branches for fields 8 and 9, and a flag so `typed_range`
outranks the deprecated field whatever order they arrive in. No new decode logic.

## Results

`test/scipcheck.sh`: 34/34 PASS, including the fuzz and degrade arms.

Same petclinic index, before and after:

```
                       stock +scip     patched +scip
overlay match          none            79% of occurrences (521/656)
--callers=getPet       7 callers       14 callers
  declined_calls       9               0
--callers=findAll      0 callers       7 callers
  declined_calls       5               0
graph_ambiguous        6               0
```

Scored against a scip-java oracle on the same repo, overload-resolution recall goes from 23%
to 100% with precision staying at 100%. Answers that were already correct do not change.

## Not checked

Whether `scip-clang` or `scip-python` still emit the deprecated form. If they do, this only
affects `scip-java` consumers today, and the patch leaves their path untouched.


joyful-ii-V-I:
Thank you, @dpunosevac. This is an excellent report and a clean fix.

You isolated the cause before touching any code: an index hand-built with the same documents and lines, using the deprecated field, attaches on stock 0.6.0. That pins the problem to the encoding, not the join. You're also right that `scipcheck.sh` can't see it, because `make_index.py` writes the same deprecated field the reader expects.

Two things before it lands:

1. **A `typed_range` arm, yes please.** This project writes the gate before the code it measures. The fixture should emit `single_line_range` and `multi_line_range` occurrences, plus one occurrence that carries both forms, to pin that the typed form wins whichever field arrives first. That arm should fail on stock 0.6.0 and pass with your change. If you'd rather not write it, say so and we'll add it as a maintainer commit on your branch. The fix stays yours either way.
2. **CI.** Workflow runs from a first-time contributor wait for a maintainer to approve them. Yours is approved now, so the checks will report here.

A small change to how `--scip` opens the index file is also in flight in `src/scip.h` (#197). It touches different lines, and your branch merges cleanly with it.

dpunosevac:
Arm added in 44d6e561.

`make_index.py --typed` re-encodes the fixture in the typed form, spread over four occurrences so each encoding is independently observable:

| occurrence | range form | wire order |
| --- | --- | --- |
| `handler` def, alpha.cpp:11 | `multi_line_range` | alone |
| `helperAlpha` def, alpha.cpp:6 | `single_line_range` | alone |
| `helperAlpha` ref, alpha.cpp:13 | both | deprecated first |
| `handler` ref, caller.cpp:9 | both | typed first |

Each both-forms occurrence puts line 99, past EOF in both fixture files, in its deprecated half. The S5 gate drops anything resolving there, so the edge survives only if the typed half won. One occurrence each way covers both wire orders rather than only the one my patch happened to hit first.

The `multi_line_range` spans 10 to 13 so `start_line` and `end_line` differ: a reader taking field 3 for the start binds that def to line 14, where no `handler` exists, and it surfaces as a def unmatched.

`helperAlpha` is new to the fixture. The third encoding needed a second def/ref pair, and alpha.cpp already had the call.

Arm 10 asserts 100% (2/2), 0 defs unmatched, `precise=2`, `prov="scip"` on both edges, no `amb=` on `run`, no staleness hint, determinism on stdout and stderr, well-formed XML.

Arm 10b is the vacuity check. `--typed --blind` re-emits the same index with fields 8 and 9 stripped, exactly the view a typed_range-blind reader has of it, and asserts it pins nothing.

Results:

* patched: `scipcheck: ALL PASS`, 50 PASS lines, up from 34.
* stock: 5 FAIL, all in arm 10. Refs unmatched, `precise=` absent from the header, `run` back to a two-way split with `amb="1"`, `handler` to `helperAlpha` unpinned. stderr reads `SCIP matched 0% of occurrences (0/0)` against `100% (2/2)` patched. Arm 10b passes on both, as it must.

One qualification on "stock": I built `f82177af~1`, the PR base, not the `v0.6.0` tag. `git diff v0.6.0 f82177af~1 -- src/scip.h` is comment-only, three lines in the `loadScipOverlay` doc block about the missing-file refusal, so the decode path is identical and the base stands in for the release.

Also green: `scipjoincheck`, `namedfileinputcheck`, `commentcoherencecheck`, `bashsourcecheck`, `gatecountcheck`, `shellgateindexcheck`, `pargatescheck`, `formatgatecheck`. The regenerated `index.scip` is byte-identical to the checked-in one, so the `occurrence()` refactor is wire-neutral.

`.github/pargates-shard-weights.json` lists `scipcheck.sh` at 11.6 and the arm adds six binary runs. `pargatescheck` passes so nothing pins it, but say the word if you want it renumbered.

joyful-ii-V-I:
Thank you, @dpunosevac. The arm is exactly what we hoped for:
- four occurrences, so each encoding can be observed on its own;
- the line-99 decoy in the deprecated half, so only a typed-range win survives the S5 gate;
- a `multi_line_range` whose start and end differ;
- the `--blind` vacuity arm.

Five FAILs on the old reader and 50 PASS with the fix is the evidence this project asks for.

One thing had moved under the branch. #197 merged a few hours ago, and it also edited the header comment of `test/scipcheck.sh`, so the PR showed a conflict and CI could not start. I merged main into your branch as a maintainer commit and kept both comment blocks: #197's lines about the refusals, then your TYPED RANGES bullet. No code changed. `scipcheck`, `scipjoincheck` and `namedfileinputcheck` pass on the merged tree, and CI is running now.

No need to renumber the shard weights.

joyful-ii-V-I:
Merged. Thank you, @dpunosevac.

`--scip` had silently ignored every index scip-java writes, because ripwire read only SCIP's deprecated `range` field. You tracked the problem to the encoding before touching code, and fixed the reader so the typed form takes precedence, as `scip.proto` asks. Then you gated it: arm 10 goes red on the old reader, and arm 10b proves the fixture cannot pass without fields 8 and 9. On spring-petclinic that took the overlay from no matches to 79% of occurrences, and `graph_ambiguous` from 6 to 0.

We merged main into your branch once, to resolve a comment conflict with #197 in `test/scipcheck.sh`. Your code is unchanged.

It is merged, not squashed, so your commits keep your name in the history, and you are credited in the 0.6.1 release notes.

dpunosevac:
@joyful-ii-V-I thank you!
Glad I could help.

PR #224, opened 2026-09-14
feat(doctor): detect cross-translation-unit layout drift

### LINKED ISSUE #162: `ripwire --doctor` should detect a binary whose translation units disagree on struct layout
## What is missing

If `src/model.h` changes while a build is running, some object files end up compiled against the old
`Symbol` layout and some against the new one. Their mtimes are newer than the header, so `make`
concludes everything is up to date and never recompiles them — the build reports success, exit 0, no
warnings, forever. Nothing in CMake can repair this, and `test/g1freshcheck.sh` cannot see it,
because the binary is *newer* than its sources and only its contents are stale. The ask is for
`ripwire --doctor` to say it in one line: each translation unit records the struct layouts it was
compiled with, and the doctor reports when they disagree.

## The evidence

`CLAUDE.md` records three incidents from this failure mode, all from building across or editing
under a branch switch:

- A sanitizer build finishing after a checkout that changed `Symbol` produced a genuine
  AddressSanitizer heap-buffer-overflow report for a bug that did not exist. The overflowing region
  was 1344 bytes — exactly 14 × 96, a multiple of the *previous* struct size.
- A header's mirrored constant kept emitting its old value through repeated successful rebuilds; a
  gate failed with `parserVer=41, expected 12/42` and looked like a missed mirror update.
- Ten identical uncaught `std::length_error` aborts in one morning of worktree churn, thrown from a
  `resize( symbols.size() )`, with zero repro in 38 runs on clean rebuilds of the same commit.

Facts read from the tree that the kit rests on:

- The `ripwire` target is `src/main.cpp` plus `src/ingest.cpp`, `src/pagerank.cpp` and
  `src/infra/diagnostics.cpp`; `IngestResult` carries vectors of `Symbol`, `Reference` and eight
  other structs across that boundary.
- `Symbol` is pinned by `static_assert( sizeof( Symbol ) == 64 + 2 * sizeof( std::string ) )` — which
  every translation unit checks against *its own* view of the header, so it passes in both halves of
  a mixed binary. `Reference` has no size pin at all.
- `--doctor` (`src/verbs_doctor.h`) emits one `<c n= ok=>` row per check and exits 1 when any fails;
  `test/doctorcheck.sh` asserts a named row set.
- `--layout` and `--stray-content --abi` model struct layout from source text ("a MODEL, NOT THE
  ABI") and cannot see what a compiler actually produced.

## Size

**Good first issue.** One recording header, one registry, one doctor row, one gate, and a short
audit of the cache-format size pins. No external corpus, no network, no API keys — a C++23
toolchain and CMake able to produce both the plain dev build and a Release build, because the
records must survive `NDEBUG` and LTO.

## Where to start

`prompts/help-wanted/struct-layout-doctor.md` is a self-contained prompt for a coding agent. It
carries the hazard and its three recorded incidents, the translation units involved, why a header
`static_assert` cannot catch it, how to reproduce it safely in a scratch tree, the recording and
comparison design space (internal linkage, a per-unit identity, surviving Release and LTO, comparing
for agreement rather than pinning numbers), the doctor row's three states, the gate arms, the
acceptance criteria and the traps. Like every prompt in `prompts/`, **it ends by writing a plan and
stopping** — a maintainer agrees the plan before any code is written.

Build the gate first: a two-translation-unit fixture whose struct definitions differ must make the
comparator report a disagreement before the real row exists.

Comment here to claim it.

lennix1337:
Closes #162

## Summary

- add a release/LTO-safe cross-translation-unit layout registry for the shared model types;
- make `--doctor` report `layout` agreement, disagreement details, or the explicit `not-checked` / `no-records` state;
- add a deterministic two-translation-unit fixture that proves incompatible layouts are detected without a build race;
- cover the plain, Release, and ASan builds and register the new gate in the full suite;
- refresh the measured showcase reduction caption so its published numbers match the live recount.

## What is recorded

Each translation unit that includes `src/model.h` records `sizeof` and `alignof` for `std::string`, `Symbol`, `Reference`, `Include`, `ConstOpen`, `Binding`, `BindingAlias`, `RouteDef`, `RouteUse`, `FileHealth`, `SkippedOversize`, and `IngestResult`. Records use internal linkage, `__BASE_FILE__` for TU identity, function-local registry construction, and a `used` retention attribute so Release/LTO cannot discard the evidence.

The existing cache-format audit found no new pin to add: `CacheEntry` is already a raw 32-byte on-disk row with `static_assert( sizeof( CacheEntry ) == 32 )` and `alignof( CacheEntry ) == 8`; the remaining cache data is encoded field-by-field and guarded by the existing version/header checks.

## Validation

- plain, Release, and ASan/UBSan CMake builds;
- mixed fixture: `state=disagree`, `FixtureLayout`, sizes `4` and `8`, alignment `4` and `4`;
- single-TU fixture: `state=not-checked`, `units=1`;
- final plain/Release/ASan binaries report `built_from=4d3e1bf62`;
- `doctorcheck`, `structlayoutcheck`, `legendcoveragecheck`, `manifestcheck`, `gatecountcheck`, `g1freshcheck`, `formatgatecheck`, `ripwirepubliccheck`, `showcasecapturecheck`, and LSan smoke;
- full `pargates.py` battery in four shards with the origin/main reference binary and pinned clang-format 22: 628 passes, 0 failures, and the single expected Apple Silicon portability skip on WSL; the base-sensitive `editchecknote` arm also passed when supplied `RIPWIRE_BASE_BIN`;
- `argvdiffcheck`: all 642 vectors byte-identical after excluding only revision metadata from the two `--version` vectors.

The detector intentionally cannot identify a field reorder that preserves both size and alignment, or a TU that never includes/registers the shared model header; the doctor row discloses only evidence actually present in the binary.


joyful-ii-V-I:
This is a good contribution, and it targets the single worst diagnosis trap in this repo — the one where objects compiled across a branch switch disagree on `sizeof( Symbol )` and a sanitizer reports a real heap-buffer-overflow of a fake bug. Turning that into a `--doctor` row is exactly the right move. I've read the whole diff and approved the workflow run, so CI is going now.

**What I checked and liked.** The vacuous-pass question is genuinely closed, in both directions: `compare()` returns `NotChecked` for a single record and again when no record carries a type, `doctorLayoutCheck` only sets `ok` in the `Agree` arm, and the gate *asserts* both — arm B links a single-record binary and requires `state=not-checked units=1`, arm A links a deliberately mismatched pair and requires `state=disagree` with both unit names and differing sizes. The internal linkage, the `used` attribute against LTO, the function-local registry for static-init order, and the decision to register `std::string` as a type rather than pin a number are all the right calls, and `sortedRecords()` sorting by unit name before comparing is what keeps the determinism contract intact. `__BASE_FILE__` for TU identity is a better answer than the per-source CMake define I'd have suggested, because it touches no build system.

**Five things to fix or decide, none of them fatal:**

1. **`docs/COMMANDS.md` has a recorded `--doctor` capture whose row list stops at `git-config-trust`.** The new `layout` row is missing from it, and `docscommandscheck` is in the suite. Expect a red there; regenerate rather than hand-editing.
2. **The showcase caption edit owes its audit-trail clause.** The top-50 figure moves 81.8 to 83.7, which exceeds the recount gate's own 1.5-point tolerance — so the edit was *necessary*, and that is precisely why it needs a new dated "re-derived … at the struct-layout doctor" entry in the caption's parenthetical history. As it stands the trailing narrative reads as though the 2026-09-10 change produced 83.7, which it did not. For what it's worth I don't think this means main's number was wrong: your diff adds ~380 lines of exactly the shape that enters a ranked top-50, so the denominator legitimately moved. 81.8 is right for main's tree and 83.7 for yours. Worth one independent recount anyway, because top-50 rising while top-100 falls to the same 83.7 is a coincidence I'd rather see confirmed than assumed.
3. **The `test/argvdiffcheck.sh` `--version` relaxation is out of scope.** Normalising `built_from=` is well motivated and it's good evidence you actually ran the gate with a base binary — but that gate SKIPs in normal CI because nothing sets `RIPWIRE_BASE`, so the change is never exercised by any run here, and loosening a differential gate's strictness is a call I'd rather make separately from this feature. Please split it out.
4. **The Release/LTO arm will never actually gate.** Arm D runs only when `RIPWIRE_RELEASE_BIN` is set, and nothing in the CI workflow sets it. It prints SKIP rather than passing silently, which is honest — but "survives Release and LTO" is then asserted in the PR body rather than proven by a run. I'd like to fix that by having CI stage a Release binary for it; that's on me, not you, unless you want to take it.
5. One cosmetic consequence worth knowing: under CMake, `__BASE_FILE__` is typically an absolute path, so a `disagree` row would emit build-machine absolute paths in `unit0=`/`unit1=`. It's a failure-path diagnostic only, but it is a path leak into XML that the redaction gates don't currently look at.

**On timing, so you're not left guessing.** This will land *after* the 0.6.1 tag rather than in it. Not a reflection on the change: 0.6.1's release PR is built on a regenerated showcase capture, a regenerated `docs/COMMANDS.md`, and a gate count that four in-flight branches are already moving, and this PR touches all three. Merging it into that window would mean redoing work the release is standing on, for no benefit to you — the change lands either way, and landing first in 0.6.2 gets it a clean CI run and a proper entry instead of a scramble. You'll be credited in the changelog when it lands.

Also: the acceptance criterion asking for a sentence in `CLAUDE.md`'s hazard section next to the `g1freshcheck` note is still open — the sentence went into `docs/ARCHITECTURE.md` instead. Both are worth having, but `CLAUDE.md` is where an agent hits the trap.

joyful-ii-V-I:
Correcting myself: **CI is fully green, 31 of 31, and the `docscommandscheck` red I told you to expect did not happen.** I was wrong, and the reason is worth knowing because it changes what you owe on that point.

I assumed a gate called "docs commands check" verifies the recorded *output* in `docs/COMMANDS.md`. It does not. Arm (B) compares the documented flag *name set* against the binary's, in both directions, and arm (G) requires the document to be byte-reproducible from the committed capture plus the binary's own `--help`. Your change adds a **row** to an existing verb's output rather than a new flag, so the flag set is unchanged and the document still regenerates byte for byte. Nothing in the suite looks at whether the recorded `--doctor` sample still shows what `--doctor` actually shows.

So the practical position is the opposite of what I said: the `--doctor` capture in `docs/COMMANDS.md` is stale with respect to your new `layout` row, and **no gate will ever tell you that**. It needs doing deliberately rather than waiting for a red. That is not a defect in your PR and it is not something I want you to fix in it, since the release PR regenerates that document anyway over the final tree. I am noting it so the stale sample is a decision rather than an oversight.

The rest of my review stands as written, in particular the showcase caption owing its dated "re-derived" clause, and the `argvdiffcheck` `--version` relaxation being worth splitting out. And the Release/LTO arm point remains mine to fix: your arm D is honest in printing SKIP, but nothing in CI sets `RIPWIRE_RELEASE_BIN`, so "survives Release and LTO" is asserted rather than gated. I have since found the same shape elsewhere in the suite, so I am treating "an arm that can only ever SKIP in CI" as a category to sweep rather than a one-off in your PR.

Thanks for a genuinely good contribution. Turning the `sizeof( Symbol )` mid-build trap into a `--doctor` row is the right answer to it, and the fact that your gate asserts the not-checked and disagree states in both directions is what made it quick to review.

lennix1337:
Follow-up to the review:

- Added the requested hazard sentence to `CLAUDE.md`, pointing agents at the `--doctor` `layout` row and the clean-rebuild action.
- Added a dated showcase audit entry for the struct-layout re-derivation (`top-50: 81.8 -> 83.7`) and reran the independent recount (`89.6 / 83.7 / 83.7`).
- Removed the `argvdiffcheck.sh` `built_from=` normalization from this PR so that harness change stays out of the layout feature scope.

I left `docs/COMMANDS.md` regeneration and the Release/LTO CI-arm sweep for the release/sweep work you mentioned. The absolute-TU-path observation also remains unchanged for now; it is a failure-path diagnostic format decision rather than a blocker for this PR.

joyful-ii-V-I:
Thanks @lennix1337. This PR turns the worst diagnosis trap in CLAUDE.md, a sanitizer report of a bug that exists only because half the objects saw an old `Symbol`, into one `--doctor` row, and its fixture proves `disagree` and `not-checked` without a build race. The follow-ups from the last round (the CLAUDE.md sentence, the dated caption clause, the argvdiffcheck split) all landed.

**The conflict is ours to fix.** `main` moved to 0.6.1 underneath you (gate count, the `regression.sh` loop, the showcase caption). We'll merge `main` into this branch ourselves after #235/#233 land, and re-derive those on the merged tree, so please don't rebase or resolve it.

**Must fix before merge**
1. **Move compiler and Release discovery into the gate, and drop the three `ci.yml` additions.**
   - Job-scope `CXX` reaches every gate that reads `${CXX:-…}`, 19 besides yours (`clonebandcheck`, `strkerncheck`, `adaptivecutshapecheck`, `noaliascheck`, …), and silently changes their compiler on both Linux legs. None of that has run in CI yet.
   - Suggested shape: default `CXX` to `CMAKE_CXX_COMPILER` from the `CMakeCache.txt` next to `$BIN`, as `test/noaliascheck.sh:86-87` does.
   - Choose arm D from `"$BIN" --version`'s build type: Release runs it on `$BIN`, `dev` keeps your SKIP banner. The Release legs then prove the Release arm with no workflow change, and a plain binary can never be labelled "release".
2. **Say in the legend what `agree` cannot see.**
   - `ok="1"` covers only `sizeof`/`alignof` of the 12 types registered in `src/model.h`. A same-size change reads `agree`, and 0.6.1 just made one: `Symbol::testScope` became a 1-bit field beside `internalLinkage` with `sizeof` unchanged.
   - Something like "agree compares only the types= registered in src/model.h; a same-size layout change or a stale constant is invisible, so agree does not rule out a mixed binary" would do. Please also define `state="no-records"`.
   - Carry the same caveat into the CLAUDE.md sentence, and soften "fails a mixed binary" in `docs/ARCHITECTURE.md`.

**Smaller, same push if convenient:**
- Use `ripwire_cxx_std_flag "$CXX"` from `scripts/cxxstd.sh` instead of a hardcoded `-std=c++23`; AppleClang 15 rejects `c++23`.
- Have arm C expect `types=` equal to the `RIPWIRE_LAYOUT_TYPE_ENTRY` count in `src/model.h`, not just ≥1.

- Release arm only SKIPs in CI: already addressed, and item 1 keeps that coverage without the env plumbing.
- Pass the matrix compiler to the gate: addressed; item 1 only narrows how.
- `permissions: contents: read` / `persist-credentials: false` on `release`: pre-existing job settings your diff doesn't touch; that hardening is ours.
- Docstring coverage 10%: not a convention in this repo.

Queue: this PR is 5th in landing order — #231 → #237 → #236 → #235 + #233 (reconciled together) → #224 → #44 (Windows, being refactored).

joyful-ii-V-I:
@coderabbitai pause

joyful-ii-V-I:
Hi @lennix1337. We'd like `--doctor`'s layout-drift row in the next release, so we'll finish the review items above ourselves:
1. Compiler and Release discovery move into the gate (`CXX` from the `CMakeCache.txt` next to `$BIN`; arm D chosen from `--version`'s build type), and the three `ci.yml` additions go.
2. The legend states what `agree` cannot see (same-size layout changes, stale constants) and defines `state="no-records"`. CLAUDE.md and `docs/ARCHITECTURE.md` get the same caveat.
3. `ripwire_cxx_std_flag` replaces the hardcoded `-std`, and arm C checks the registered type count.

They'll be maintainer commits on top of your branch head, landed through an integration train. Your commits stay as they are, and you're credited in the CHANGELOG. We won't push to your branch, so #44's refactor isn't affected. Thanks again for turning the worst diagnosis trap in CLAUDE.md into one `--doctor` row.

joyful-ii-V-I:
A correction to my note above, @lennix1337: you had already pushed all three items in `e9c5f283`, before we wrote it. Sorry we missed that, and thank you.

We checked each one on a build:
- **Discovery:** compiler and Release discovery now happen in the gate, the `ci.yml` additions are gone, and arm D SKIPs on a `dev` binary but really runs on a Release tree.
- **Caveats:** the `agree` caveat and `state="no-records"` are in the legend, CLAUDE.md and ARCHITECTURE.md.
- **Arm C:** it uses `ripwire_cxx_std_flag` and checks the exact type count.

No maintainer commits are needed for those. This lands as-is in our next integration train, where we reconcile the shared pins. The remaining quality rows on `structlayout_probe.cpp` are ours to handle at merge, as the review said.

joyful-ii-V-I:

PR #231, opened 2026-09-14
fix(uses): a :: selector resolved defs= and then answered a silent count=0

### LINKED ISSUE #164: `--uses` answers a silent `count="0"` for a `::` selector that `--callers` resolves
## What is wrong

Every symbol-taking verb accepts the two `::` spellings ripwire prints about itself: the canonical id
`path::scope::name` that every map row carries in `id=`, and the `Scope::name` that `--edit-check`
prints as `sym=`. `--callers`, `--callees`, `--impact` and `--expand` all resolve them. `--uses`
resolves them too — and then answers `count="0"`, with no rows and no disclosure, for symbols that
plainly have call sites. A zero that reads as "defined, and nothing uses it" is exactly the
misreading `CLAUDE.md` non-negotiable 3 exists to prevent, and the callers answer's own `next=`
pointer sends the reader straight into it.

## The evidence

On the `test/declinefix` fixture, three spellings of one symbol:

```
--callers=cpp/pair/one.cpp::One::ctwin   count="1"  next="--uses=cpp/pair/one.cpp::One::ctwin"
--uses=cpp/pair/one.cpp::One::ctwin      defs="1" external="0" count="0"      (no rows)
--uses=cpp/pair/one.cpp:ctwin            count="1"  <u role="call" p="cpp/pair/user.cpp:3" …/>
```

On this repository itself, `--callers=NoteIndex::empty` counts 816 callers and
`--uses=NoteIndex::empty` counts 0.

The cause is one comparison: `resolveUsesSelector` in `src/verbs_navigate.h` keeps the whole `::`
spelling as the site match key, and `collectUseSites` compares that key against reference names,
which are always bare. The same zero reaches `--safe-delete`'s `uses=`, `--verify="uses(…)"` and
`unused(…)`, which share the scan, and the MCP `uses` verb, which carries its own copy of the
comparison. It was characterized across 16 selectors in C++, Python and Rust.

`test/usesselectorcheck.sh` already pins the gap. On a build of `main` it is ALL PASS, 57 checks
with 12 of them labelled KNOWN GAP; each asserts today's wrong answer and names its FIXED
expectation in a comment beside it. Against a wrapper faking the naive "strip the scope" fix, 12
arms go red — nine read MOVED, and the two precision controls and the wrong-scope control fail.

## Size

**Good first issue**, though not a trivial one. The root cause is a single comparison in a single
function, and fixing it there reaches all three CLI verbs at once; there is no new gate file to
write, because the arms to flip already exist and every code pointer is verified. What keeps it from
being tiny: the MCP twin holds its own copy of the comparison and needs a decision, the `--uses`
legend has 25 bytes of budget left, and `--help` plus the generated `docs/COMMANDS.md` move with the
change.

## Where to start

`prompts/help-wanted/uses-qualified-selector.md` is a self-contained prompt for a coding agent. It
carries the reproduction, the characterization table, file:line pointers with their symbols, the fix
shape and its edges, the honesty rules that apply, the acceptance criteria and the traps. Like every
prompt in `prompts/`, **it ends by writing a plan and stopping** — a maintainer agrees the plan
before any code is written.

Acceptance is flipping every KNOWN GAP arm in `test/usesselectorcheck.sh` to its FIXED line, red
against a build of `main` and green on the fix, with every control unchanged.

A sibling kit, `prompts/help-wanted/next-uses-bare-name.md`, changes the callers answer's `next=` on
a *declined* call and names this *bound* `::` case as a separate, older gap. The two touch different
code, read the same fixture, and can land in either order; neither should absorb the other.

Comment here to claim it.

aniruddhaadak80:
Fixes #164. `--uses` on a `::` selector resolved `defs=` and then answered a silent `count="0"`.

## Defect

Every symbol-taking verb resolves the two `::` spellings the tool prints about itself — the canonical id `path::scope::name` and `Scope::name` — through `resolveAllByNameQualified`. `--callers`, `--callees`, `--impact` and `--expand` then read the graph by NodeId, so the spelling stops mattering. `--uses` did not: `resolveUsesSelector` (`src/verbs_navigate.h`) kept the whole spelling as the site-match key, and `collectUseSites` compared that key against reference names, which are always bare. `defs="1"` beside `count="0"` reads as "defined, nothing uses it" — the misreading non-negotiable 3 exists to prevent — and `--callers`' own `next=` pointer sends the reader straight into it. `--safe-delete`'s `uses=`, `--verify="uses(...)"/"unused(...)"` and the MCP `uses` twin ride the same scan/comparison.

Reproduction (code path, no binary on hand — see the toolchain note): `sym.find("::")` at `resolveUsesSelector` forced `fileQualified=false` and `siteMatchName=sym`; `r.calleeName != sel.siteMatchName` then skipped every reference. The pre-fix behavior is pinned on `main` today by the 12 KNOWN GAP arms in `test/usesselectorcheck.sh` (each PASSes asserting the silent zero).

## Change

- `resolveUsesSelector` takes the resolved `defs` and, for a `::` spelling whose defs are non-empty and share one name (`commonDefsName`, read off the defs — never a strip of the spelling), binds `siteMatchName` to that name and sets `scopeNarrowed`. The call role narrows through `usesChosenCallers` exactly as a `file:name` selector does; other roles stay name-matched. Mixed names (Elixir arity suffixes) keep today's whole-spelling key rather than a strip-and-match the precision controls forbid.
- The flag is set only on a resolving spelling, so `Nope::ctwin` keeps the generic refusal's bytes, and member spellings with no symbol behind them still reach `memberUsesArm` untouched (it only fires on empty `defs`).
- `--safe-delete` and `--verify` get the narrowed scan with no grammar change; `--callers`' `next=` is untouched (the sibling kit owns it).
- MCP twin, the plan's option (b): a resolving `::` spelling refuses as CLI-only with the bare-name retry and the CLI form — the way `file:name` already refuses — instead of the silent zero. Member answers are exempt (consulted first, mirroring the CLI's member precedence). Tool descriptions untouched (the manifest ceiling has 23 B of headroom and this needs none).
- Legend and `--help` say what narrows now: `A "file:name" or "::" SYM narrows ... (qualifier only)`. Legend +2 B (3956 vs the 3979 budget), help +15 B. `docs/COMMANDS.md` prose and the newest capture re-derived by the same substitution the binary emits.

What deliberately did not change: the `next=` pointer, every "Not this issue" row in the kit, the per-enclosing-symbol narrowing granularity, and the `Owner.field` member path (`test/fieldusescheck.sh` arm F pins `Tally::limit` as the symbol spelling and only asserts rc/defs/no-member, all preserved).

## Tests

- `test/usesselectorcheck.sh`: every KNOWN GAP arm (section f plus arm d) flipped to its FIXED line, plus the kit's two extra assertions — same-definition-same-rows (`::` vs the proven `file:name` twin, on the fixture and on this repo) and verbs-agree (every `role="call"` row sits inside a `--callers`-listed caller, via `declinecheck.sh`'s `call_sites` relation). Premises, controls, precision, negative and wrong-scope arms are byte-identical.
- Toolchain note: this box (Windows) has no C++ compiler, CMake, or bash — only Node, `gh`, and Python via the `py` launcher — so no local build or gate run was possible. Proof is CI on this PR: the flipped gate green, plus the full suite. I verified statically what can be verified statically: the COMMANDS.md prose against `docs_commands_build.py`'s own parse/join/split/caveats (Answers, rest-prose, all 3 caveats match), the legend budget arithmetic, Allman/brace/span/container rules by review, and that no other caller of the old `resolveUsesSelector` signature remains.
- Known upcoming red, owned: `test/printffmtparitycheck.sh` pins `--help`/`--help=all`/`--help=--uses` and the `--uses=distance` answer by SHA-256, and both changed bytes by design (help prose, legend sentence). No other label should move — every pinned selector is a bare name and the fix is confined to `::` spellings plus unconditional legend prose. I will re-pin exactly the red labels from CI-reported actuals in a follow-up commit on this branch, after reading which labels moved and why. If anything else moves, that is a defect in this change, not a pin update.


aniruddhaadak80:
Note for the maintainer: the CI run on this fork PR is sitting at action_required (first-time-contributor approval gate) — no jobs have started, so there is nothing to diagnose yet on my side. Once the run is approved and going, I expect exactly four printffmtparitycheck labels red by design (help, help_all, help_one, uses — help prose + legend bytes), and I will re-pin those hashes from the CI-reported actuals in a follow-up commit on this branch. Anything else red is a defect in the change and I will fix it here.

aniruddhaadak80:
Diagnosis of run 34864502790 (all three failure groups), plus a fix just pushed in c80b3be:

**1. \printffmtparitycheck\ FAIL (\uses\ stdout e8eac2b7 -> 3ed4a98e) — STALE, no action needed.** That run tested headSha \40e69fb\; the re-pin commit \4efa187\ (manifest now carries \3ed4a98e\, the fixed output) landed after. A re-run on current HEAD should clear this gate.

**2. \usesselectorcheck\ order-gate (\call_sites\ used at line 170, defined at line 220) — REAL, fixed in \c80b3be\.** My rewrite left the helper defined with the (f)-arm helpers while arm (d) calls it earlier. Pure move of the byte-identical function above first use; no behavior change.

**3. \elixirsemanticcheck\ \KeyError: 'result'\ — REAL, needs a maintainer call.** The gate's MCP arm (\elixirsemanticcheck.sh:256-262\) sends \	ools/call\ \uses\ with a \::\ selector (\Contracts::@limit\) and asserts the MCP answer equals the CLI answer. My \mcpverbs.h\ change deliberately refuses \::\ spellings on that verb (JSON-RPC error reply, hence no \
esult\ key) on the grounds the verb has no narrowing machinery. So the gate's premise (MCP answers \::\) and my change (MCP refuses \::\) directly conflict. Options: (a) I extend the MCP verb to serve the narrowed answer like the CLI — real work, say so and I will do it; (b) the gate's MCP arm switches to the bare name for \::\ cases. Which do you prefer?

Note: no C++ toolchain on this box, so CI is the proof for all of the above — the push will re-run the full matrix.

aniruddhaadak80:
Rebase hygiene: rebased \ix/uses-qualified-selector\ onto \upstream/main@30f14a2\ (was based on \2f90ba\). Before \c80b3bea61b62195227183b8d5dfaada9767f526\ -> after \ef0f9b81723af872e39caa2757c5ad2efc060bba\, 4 commits carried, no logic changes. Conflict in \	est/printf_parity.manifest\ (re-pin commit): kept upstream \help\/\pack_task\ pins, carried this PR's \uses\/\help_all\/\help_one\ pins. Caveat: merged help text is upstream changes + this PR's \cli.h\ delta, so \help_all\/\help_one\ may need a fresh re-pin from CI actuals on this head — will follow up from CI output if \printffmtparitycheck\ reports new values.

joyful-ii-V-I:
Thank you @aniruddhaadak80. You read the site-match name off the resolved defs instead of stripping the spelling, and that is the right shape: `--callers`' own `next=`, `--safe-delete` and `--verify` on a `Scope::name` now reach the real call site, and the precision controls stay green. You also did it with CI as your only compiler.

On your Elixir question: neither (a) nor (b). A `::` spelling whose definitions are all Elixir never hit the silent zero. Both `usesText` and `collectUseSites` already match those references through the Elixir resolver, which is why `elixirsemanticcheck` can compare the two surfaces.

**Must fix before merge**
1. Leave that all-Elixir population as it is on main, on both surfaces. When every resolved def is `Lang::Elixir`, skip the new refusal in `usesSelectorRefusal` and skip `scopeNarrowed` in `resolveUsesSelector`. `elixirsemanticcheck` (`KeyError: 'result'`) then goes green with no gate edit.
2. `mcpverbs.h:2498`: drop `&& resolveFieldSelector( ing, symbol ).empty()`. `usesText` serves a field only when `defs` is empty (line 2569), so a `::` spelling that resolves to both a symbol and a field skips the refusal and gets the old silent `count="0"`. The CLI checks the symbol first too: `memberUsesArm` returns early when defs are non-empty.
3. `usesselectorcheck.sh` arm (d)'s "verbs agree" check has never run. At 40e69fb `call_sites` was called before it was defined, so the arm passed on empty output. At this head it compares the default 100-row `--uses` window with the 40-row `--callers` window of an 858-caller symbol. Pass a `--limit` that covers both runs. In (d) and (f), fail when `call_sites` exits non-zero, and require at least one call row. This is CodeRabbit's line-170 comment; c80b3be moved the function but didn't fix that.
4. Revert your edit to `docs/captures/COMMANDS_showcase_2026-09-13.md` and the sample-legend hunk in `docs/COMMANDS.md`, but keep the prose line that comes from `--help`. Since the rebase the generator reads the newest capture, `2026-09-14`, so the doc no longer regenerates byte for byte (`docscommandscheck` arm G). Captures are re-recorded by the release PR.

Nice to have in the same push: moving the `::` branch into a small helper keeps `usesSelectorRefusal` under the quality-delta bars (complexity 17→20, 59→73 lines).

**We handle at merge:** the `printf_parity` pins. `help_all` changed on main between your old base and 30f14a27, so the pin carried through the rebase matches neither tree. We re-record `uses`/`help_all`/`help_one` from a build of your final head, so please don't re-pin from CI output.

- "Align the Elixir MCP uses gate…": item 1 fixes this in the code, so the gate stays as it is.
- "Preserve resolved member selectors": the behaviour is real, but it is the same on main (your new branch never sees those spellings), so it is out of scope here.
- Pre-merge "Out of Scope" (churn-decay) and "Docstring coverage": the first comes from reviewing across your rebase (this diff has no such change), and docstring coverage is not a project rule.

Queue: this PR is 1st in landing order — #231 → #237 → #236 → #235 + #233 (reconciled together) → #224 → #44 (Windows, being refactored).

joyful-ii-V-I:
CI note for @aniruddhaadak80: the finished run on `ef0f9b81` matches the review above exactly, with nothing new.
- **`elixirsemanticcheck`** shows the `KeyError` (item 1).
- **`usesselectorcheck`** arm (d) (item 3).
- **`docscommandscheck`** (G) (item 4).
- **The `printf_parity` byte-parity step:** the `help_all` pin we re-record at merge, so no action needed from you.

Once items 1–4 are pushed, this is first in line.

joyful-ii-V-I:
@coderabbitai pause

aniruddhaadak80:
CI diagnosis (run 34968692448, all 14 failures share one root cause — no fix pushed yet, holding for the rebase below): the printffmtparitycheck.sh gate fails only on help_all — STDOUT bytes moved (23a83bad -> bd532391), i.e. a printf-family conversion altered the help output. That is this PR working as intended: the uses-legend sentence now documents the :: selector narrowing, and docs/COMMANDS.md plus the showcase capture mirror it. The manifest I blessed (23a83bad) went stale because the branch is also CONFLICTING with main — main moved the same legend region after I hashed. Plan: rebase onto current upstream/main, then re-bless the manifest to whatever hash the rebased tree produces (verifying the delta is only the intended legend sentence), and push. If the rebased actual hash surprises, I will report back instead of blessing blindly.

joyful-ii-V-I:
Hi @aniruddhaadak80. We'd like this fix in the next release, so we'll finish the four review items above ourselves instead of asking you for another round:
1. the all-Elixir population keeps main's behaviour;
2. the `mcpverbs.h` field condition;
3. `usesselectorcheck` arm (d) actually runs;
4. the capture and doc reverts.

They'll be maintainer commits on top of your branch head, landed through our next integration train with the shared pins. Your commits stay as they are, never squashed, and the CHANGELOG entry credits you. If you'd rather push them yourself in the next few hours, just say so here and we'll take your push instead. Thanks for reading the site-match name off the resolved defs; that was the right design.

joyful-ii-V-I:

PR #226, opened 2026-09-14
docs(readme): the badge advertised a daily ranking the repo has since beaten

joyful-ii-V-I:
## What

The Trendshift badge in the README pointed at the **daily** C++ feed. The repository took **#1 C++ for the week**, so the badge was understating its own result.

Two substitutions on one line:

- `repositories/217924/daily?language=C%2B%2B` → `repositories/217924/weekly?language=C%2B%2B`
- alt text `Trendshift: C++ Repository of the Day badge for redhat-et/ripwire` → `… Repository of the Week badge …`

The alt text moves with the image so the accessible name states the same fact the image does.

## What is deliberately unchanged

The repository id `217924`, all three `utm_` parameters, and the `width`/`height`. House markup style is kept as-is: descriptive alt text, no `target=` or `rel=`, and `>` rather than `/>`.

## Gates

None apply. I checked before writing the commit: nothing under `test/` matches the badge image URL or the centred-paragraph wrapper, so this moves no pin and changes no gate. It is one line of README markup.

Split out of the 0.6.1 release PR on purpose — a weekly ranking is time-sensitive and the release has several lanes still landing, so this should not wait for them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #234, opened 2026-09-15
release: 0.6.1 — the release PR (version, CHANGELOG, README, lineage, deck, regenerated artifacts)

joyful-ii-V-I:
The last PR before the 0.6.1 tag. All five feature lanes (#212, #214, #215, #219, #223) plus #216 and
#226 are already on `main`; this one writes the release down, moves every number that moved with it, and
fixes the handful of things the lanes deliberately left to the release.

**The owner tags. Nothing here tags, and nothing dispatches `release.yml`.**

## What is in it

**Version.** `project( ripwire VERSION 0.6.1 … )` in `CMakeLists.txt` — the one place that holds it.
`src/version.h` is generated from it and `test/versioncheck.sh` pins it against the binary's own
`--version`. The ~20 other files that mention `0.6.0` are historical claims about when a requirement
started, gate fixture data, regenerated captures or source comments, and are deliberately untouched.

**CHANGELOG.** A curated `## [0.6.1]` section covering everything since the v0.6.0 tag, with the per-lane
entries that had accumulated under `[Unreleased]` folded in: the ones the narrative already covers are
superseded by it, and the ten it did not cover are carried, condensed to a paragraph each. `[Unreleased]`
is left as an empty stub. Credits to **@henry-hz** (#207), **@dpunosevac** (#198) and **@andriytyurnikov**
(#216), each beside the entry their work produced, plus @antoleod and @heliocipher for the work of theirs
that landed in this window.

**README.** The 0.6.1 paragraph goes above the 0.6.0 one, newest first, leading with the size reductions —
every figure in it carries the corpus and method its CHANGELOG entry states, and the memory figure gets its
own sentence so it cannot be read as an output-size claim. "the cold parse drops from 194 s to 156 s" gains
`of CPU` at both sites that lacked it (the three that already carry the unit are structurally out of reach
of that substitution). The reference-guide preface is replaced: it said the tool "uses a network only when
you give it a git URL", which omits `--mcp --listen=HOST:PORT` — the path with security consequences — and
"no daemon", which the shipped MCP server mode contradicts. The opening blockquote is kept verbatim.

**Lineage.** `docs/LINEAGE.md` gains the Algorithmica S-tree chapter that `src/infra/dynamic_map.hpp`
already credits in its own header comment. The row says what transferred — the branchless key-scan kernel,
not the structure — and the paper count moves 70 → 71 at every site that publishes it, with §2's prose
count of not-peer-reviewed sources moving with it.

**The gate for that count.** `counts_from()` in `test/readmedriftcheck.sh` reads the first match of each
claim and stops. Arm (E2b) already covered that for the repository/paper pair; nothing covered it for the
survey figure. New arms: **(E2d)** every copy of the survey sentence must agree on one value, **(E2e)** its
mutation control on the SECOND copy — the one a `head -1` extraction can never see — and **(E2f)** the
population these matchers do *not* reach, because README spells that figure a third way no arm examines.
The `head -1` calls are now documented as deliberate and covered rather than left looking like an oversight.

**CONTRIBUTING.** The exception rule this codebase follows and had never written down: avoid exception
handling, and where a throw is unavoidable, RAII is what makes it safe — cleanup belongs in a destructor,
never in a `catch`. Both numbers in it were re-derived on this branch's base (27 real handlers under `src/`,
exactly one of which releases a resource by hand; a bare grep reports 35, eight of them the word in a
comment or a query string), and the bullet says how to re-derive rather than asking to be trusted.

**The six `VERIFY( call )` sites**, read one at a time rather than counted, are three risk classes. Under
`NDEBUG` `VERIFY` lowers to `__builtin_assume`: the argument is never evaluated and the optimizer is
*promised* the predicate holds.

- `gitmine.h`'s `sinceLogArgs` promised that a git object name arriving from **outside the process** is well
  formed. That one could be false at runtime, so it is now a real check that degrades to the caller's own
  fallback window — the window an inactive scope already yields — with `DEGRADED_PATH_ALERT`, instead of
  reaching `git log` as a positional argument. `CONTRIBUTING.md` already scopes `VERIFY` to what the code
  *guarantees*, so this site was a misuse of the macro, not evidence against it — and the doc is untouched.
- `pagerank.cpp`'s teleport-mass identity was a floating-point promise in the one translation unit compiled
  without FP reassociation, which works against the determinism contract.
- The four internal-invariant sites keep the check and drop the release promise: a corrupt CSR is not
  impossible, and in release the bounds reasoning has already been optimized away on the promise's strength.

`VERIFY_DEBUG_ONLY` is the form for the last two classes — the same check in debug, and nothing at all in
release. `VERIFY` stays the default for a precondition the code guarantees.

**Deck.** One new slide, "0.6.1 at a glance", in the existing four-quadrant shape, every figure naming a
merged PR and quoted back to its source in the notes block. The slide count moves 33 → 34 at all four sites
that state it and the flag count comes from `deckclaimcheck` (180), never typed. The `.pptx` and its PDF
render are rebuilt from the generator.

**Regenerated, in the order that keeps each product valid.** The showcase capture is recorded **ref-clean and
detached at the release commit**, from a fresh clone whose only local ref is `main` and whose binary was built
in that clone (`built_from=` is the release commit, no dirty marker): the recording checkout's local branch
names reach the published document through `--stray-content` and `--whereis`, and the machine this was built
on holds 256 branches that are not on origin. 254 commands recorded, 162 lifted into `docs/COMMANDS.md` as
samples. README's claim about that document is re-derived from the generator that writes it — **176 documented
flags, 162 with a recorded invocation**, where it had said 145 and 94 with a three-week-old "re-derived" date.
`docs/LIMITS.md` (212 caps), `docs/TUNING.md`, the gate count (618 at 8 marked sites) and
`test/printf_parity.manifest` all regenerate byte-identical on this tree — run, not skipped.

## Verification

- Full suite in one run: **gates=632 pass=630 skip=2 fail=0**, `wall=723.4s`, `tree_writes=0`. The two skips are the
  expected environmental pair: `argvdiffcheck` skips entirely (no `RIPWIRE_BASE` reference binary) and
  `editchecknotecheck` runs with only its arm (c) skipping (no `RIPWIRE_BASE_BIN`).
- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD` — **`gating="0"`, `regressions="0"`**, exit 0.
  The 83 `stale=` rows are ack-ledger hygiene, never gating and never counted in `regressions=`; the ledger
  is not hand-edited.
- ASan/UBSan run over the repository clean under the committed LSan suppressions (exit 0, empty stderr).
- `g1freshcheck` run explicitly (it is one of the individually-invoked gates, not part of the parallel suite): the asan binary is fresh.
- `xmllint --noout` clean, and two runs of the default map byte-identical.

Three gates went red on the first suite run and are fixed in their own commit, each a rule the suite already
states: a three-space trailing comment where `.clang-format` asks for one; the new macro's comment naming a
host domain type under `src/infra/`, which `infraportcheck` (B) forbids; and the new slide saying "9 verbs"
of CLI verbs, which `mcpverbscheck` (8) reads as the MCP `tools/list` count.

## Two things the plans did not anticipate, recorded rather than smoothed over

1. **The release-notes body limit is the binding constraint on the CHANGELOG section.** Carried verbatim, the
   `[Unreleased]` entries the curated narrative does not cover took the extracted section to **152,365
   characters**, past GitHub's 125,000-character release-body limit, which `release.yml` does not check and
   which would have failed the job *after* the tag. Condensed to a paragraph each, the section is
   **117,429 characters** — and `generate_release_notes: true` appends GitHub's own notes to that body, so
   the real margin is smaller than it looks. Worth a gate in 0.6.2.
2. **Two shapes the draft stated from an earlier tree are corrected against the binary**: the `--in=` map stub
   is `<symbols stubbed="1" would_show=N next=/>`, not `total=`/`shown="0"`, and it measures 68 B, not 61.
   `src/compactlegend.h` carried the same stale spelling in a comment, and `recentscopecheck` arm 5a2 fails if
   `total=` ever appears there — so the comment documented behaviour the gate forbids.

Not here, each for a stated reason: the round-2 loop figures (the re-measure is not part of this release, so
the provisional numbers are cut rather than published, and the sentence that promised them no longer does);
the companion shape-and-routing slide, for the same reason; and no comparison with any other tool, which is
held for 0.6.2 by owner decision.


joyful-ii-V-I:
@coderabbitai full review

Issue #228, opened 2026-09-14
--quality-delta gates a clean tree on Python: 58 dead-code false positives for Django test hooks dispatched via cls./self.


hnipps:
## What is wrong

On a Django monolith (~7,100 tracked files, Python), `ripwire . --quality-delta` on a **clean working tree** (identical to `git HEAD`, only untracked directories present) reports 58 gating regressions, all `kind="dead-code"`, all preexisting-worse, exit 2:

```
<quality-delta baseline="git-HEAD" regressions="58" minor="0" acked="0" stale="0"
  preexisting-worse="58" new-symbol="0" gating="58" register-macro-excluded="0"
  api-new-surface="0" at="a2c192068+dirty+shallow">
<r kind="dead-code" sym="api_v2/tests/test_boundary_meter_status_logs.py::BoundaryMeterStatusLogTests::create_test_objects" p="api_v2/tests/test_boundary_meter_status_logs.py:82" gating="1" .../>
<r kind="dead-code" sym="api_v2/tests/test_boundary_meter_status_logs.py::BoundaryMeterStatusLogTests::detail_url_kwargs" p="api_v2/tests/test_boundary_meter_status_logs.py:99" gating="1" .../>
...
```

Two problems stacked:

1. **A no-op diff should never gate.** Working tree == HEAD, yet the baseline snapshot and the working-tree snapshot disagree on whether these symbols are dead. Whatever differs between the two parses (ordering, resolver tie-breaks, shallow-clone history) makes the exit code unreliable as a pre-commit hook on this repo. `--quality-delta=HEAD~1..HEAD` on a one-file test commit also reports 28 preexisting-worse dead-code rows in unrelated files.

2. **The symbols are not dead.** They are Django test-class hooks dispatched dynamically. A base class in `api_v2/tests/utils.py` does, inside `setUpTestData`:

   ```python
   cls.create_test_objects()
   ```

   and ~30 test classes each define their own `create_test_objects`. Likewise `detail_url_kwargs` is called as `self.detail_url_kwargs(...)` from base helpers and overridden per subclass. The name-based resolver sees one call site with ~30 same-language definitions in other files, declines the bind, and every override reads as unreachable. This is the Python equivalent of the C++ self-registering test macro case that `register_macros` already exempts, but there is no Python-side hook for it.

The `--callers=create_test_objects` verb reflects the same blind spot (`counts_floor="1"`, callers found only via `super().create_test_objects()` chains).

## Environment

[...]

PR #237, opened 2026-09-16
fix(quality): inherited Python hooks were counted as dead overrides

qinghuanandejiangshi:
Removing redundant direct callers from Python subclasses reported inherited hooks as new dead code, even though the parent still called them through `self` or `cls`. The regression exercises 30 subclasses at runtime: all 60 hooks remain callable, but the previous quality delta reported them as regressions.

Use direct class ownership, the parser's first-parameter bindings, and the existing inheritance graph as conservative liveness evidence. Snapshot and delta share the same eligibility rule; the quality snapshot cache scheme advances to retire the old dead sets. This does not add precise call-graph edges or exempt every test method.

The gate keeps unrelated and sibling orphans visible, rejects nested-function and second-parameter receiver lookalikes, and reports the hooks again when the inherited calls are removed. It also covers sidecar, HEAD, cached HEAD, and committed-range baselines.

Refs #228. This addresses the inherited-hook false positives only. I have not reproduced the separate unchanged-working-tree mismatch on the current baseline, so that part remains open.

Validation on macOS:

- The new regression fails against an independently built, verified `30f14a274` binary and passes with this change.
- Development and G1 sanitizer builds completed; sanitizer full-repository scan has no diagnostic output, emits valid XML, and passes `qualitycheck`.
- Determinism gate: baseline plus three comparisons match; nested filtering and long-path XML checks pass. The Python fixture also produces identical quality deltas across cold/warm/warm/cold caches.
- `portablebuildcheck` and the quality semantic fingerprint gate pass; snapshot scheme is 12.
- Self quality delta has `gating=0` without added acknowledgements. It still reports two new-function complexity findings (18 and 22), one minor parameter-count finding, and four minor churn findings.
- Full suite: `632` gates, `629` passed, `3` skipped, `0` failed; `tree_writes=0` (38m 40s, six workers). The skipped gates were `argvdiffcheck`, the baseline arm of `editchecknotecheck`, and `g1freshcheck` (the sanitizer build lives outside the checkout). Follow-up runs passed `editchecknotecheck` against the verified HEAD binary and `g1freshcheck` using a temporary link to the external sanitizer build (removed afterward).

The optional `argvdiffcheck` follow-up exits 1: 643/645 vectors match stdout, stderr and exit status; the only two differences are `--version` (with/without a corpus), whose `built_from` value is `30f14a274` versus `30f14a274+dirty`. Both its differential and tree-mutation controls pass. This is recorded as a failed comparison, not folded into the full-suite pass count.


joyful-ii-V-I:
Thank you, @qinghuanandejiangshi. This is the fix #228 needed. A Django-style project can now run `--quality-delta` as a
pre-commit gate without acking a ratchet of overrides that are demonstrably live. We especially appreciate:
- the fixture executing the inherited `setup()`/`exercise()` dispatch in Python itself, so the gate proves reachability
  rather than name matching;
- the sibling, unrelated, nested-function and second-parameter controls, which keep the dead-code kind firing where it
  should;
- the dispatch-removal arm that brings all 60 rows back;
- doing this as an arm in `qualitycheck.sh` rather than a new gate file;
- recording the scheme bump in the re-pin log.

We re-derived `test/qschemetrip.hash` from source at the base and at your head and both match, and the rule can only
ever exempt Python methods.

**Must fix before merge:** nothing blocking.

**CodeRabbit comments you can safely ignore:**
- *Docstring Coverage (50% < 80%)*: house style is a `//` rationale comment above each function, which every new
  function here already has.

**Notes we'll handle on our side (no action needed):**
- A comment sentence noting that inheritance edges are resolved by base-class name, so same-named bases in different
  packages share a closure (errs toward "live", the intended direction).
- The multiple-inheritance wording in `pythonDispatchClasses`' comment.
- Adding the new exemption to `isDeadCandidate`'s doc list.
- Expanding the re-pin log entry when we re-derive the hash alongside #235/#233.

`Refs #228` is right. We'll keep #228 open for the unchanged-working-tree mismatch you noted.

Queue: this PR is 2nd in landing order — #231 → #237 → #236 → #235 + #233 (reconciled together) → #224 → #44 (Windows, being refactored).

joyful-ii-V-I:
Merged, thank you @qinghuanandejiangshi!

**What this changes:** Django- and unittest-style projects can now run `--quality-delta` as a pre-commit check without acking a growing pile of overrides that are demonstrably live. Hooks a base class calls through `self.` dispatch no longer read as dead code.

**What made it easy to trust:**
- The fixture actually executes the inherited `setup()`/`exercise()` dispatch in Python.
- The sibling, unrelated-class, nested-function and second-parameter controls each keep the dead-code kind firing where it should.
- The dispatch-removal arm brings all 60 rows back.

The small comment notes from the review, and the fuller re-pin log entry, will land with our next reconciliation. You'll be credited in the 0.6.2 changelog.

qinghuanandejiangshi:
Thank you, sir. I am delighted to contribute to the open-source community.

PR #241, opened 2026-09-16
fix(cache): a cached enum byte past its enum's last value was believed, and a span-tier memo byte wrote past a stack array

joyful-ii-V-I:
## The defect

Two on-disk readers built enums straight from bytes with no range check.

- **The ingest cache** read ten enum bytes this way: `SymKind` and `Lang` on a definition (`readDef`), `Lang`, `RecvKind` and `RefRole` on a reference (`readRef`), `Lang` and `LocalBindKind` on a binding (`readBind`), `BindKind` on an FFI alias (`readFfi`), and `HttpMethod` on both route records (`readRouteDef` / `readRouteUse`).
- **The span-tier memo** (`ripwire-stier-*`, `spanTierMemoLoad`) read one `SpanTier` byte per span. It is the `--grep` classifier's per-file blob.

An enum with a fixed underlying type holds any u8, so the cast itself is fine. The damage is downstream.

### Consequence, measured on the unfixed binary at `3bf884e2`

**Method.** The fixture has 15 files: `test/fixture`, `test/ffifix` and `test/routeedgefix`. For each field class I set the byte to 255 at every site and rebuilt every record digest and the table digest. Then I ran 24 verbs and diffed each against `--no-cache`. The control was the same battery on the unmodified caches, and all 24 verbs matched.

| field class | record accepted | verbs whose answer changed |
|---|---|---|
| `SymKind` (def) | 15 of 15 | 18 of 24 (served `t="other"`; a Field became a map symbol) |
| `Lang` (def) | 15 of 15 | 18 of 24 |
| `Lang` (ref) | 15 of 15 | 17 of 24 |
| `RefRole` | 15 of 15 | 17 of 24 (served `role="read"`; the call left the call graph) |
| `RecvKind` | 15 of 15 | 13 of 24 |
| `BindKind` | 15 of 15 | 12 of 24 |
| `LocalBindKind` | 15 of 15 | 11 of 24 |
| `Lang` (bind), `HttpMethod` | 15 of 15 | 0 of 24 (no visible change on these verbs) |

**UB.** A `Lang` byte of 32 or more reaches `src/clones.h:135`, which shifts `kHashLineCommentLangMask >> std::uint32_t( lang )`. UBSan reports `shift exponent 255 is too large` on `--for`, `--clones`, `--readability` and `--pack-task`.

**Memory safety (the memo).** A memo tier byte of 3 or more indexes `std::uint32_t tierHitCount[3]` in `grepApplySpanTiers` at `src/search.h:2174`. That is an out-of-bounds **write** on the stack:
- ASan reports `stack-buffer-overflow`, and UBSan reports `index 255 out of bounds for type 'std::uint32_t[3]'`.
- The plain binary served a different `--grep` answer.

**How reachable.** Each ingest-cache record carries its own 32-bit digest, and the offset table carries another. A random bit flip is refused before any enum is read. An out-of-range byte gets that far only from a blob written wrong, or edited with its digests rebuilt, such as a committed team artifact or a copied cache directory. For the ingest cache this fix is **defence in depth**. The memo has **no checksum**, so any flipped or hand-written byte used to reach the write.

## The fix

**Validate at the read.**
- **Ingest readers.** All ten sites go through `ByteR::enumU8<E>( count )`. A failure folds into the reader's existing `ok` flag, so the record takes the refusal path a short read already takes: that file reparses, the rest of the blob stands, and the same debug-only alert fires as its sibling guards. The helper never uses `VERIFY`, because these bytes are external input.
- **Span-tier memo.** `spanTierMemoLoad` checks every tier byte with one `std::all_of` and refuses the whole memo on a bad byte.
- Both sites are marked `VALIDATE-SITE:` with the exact `VALIDATE( … )` spelling they become when the macro vocabulary lands.

**The bound sits beside each enum, proven at compile time.**
- New count constants: `kSymKindCount`, `kRecvKindCount`, `kRefRoleCount`, `kLocalBindKindCount`, `kBindKindCount`, `kHttpMethodCount` and `kSpanTierCount`. `kLangCount` already existed.
- The new header `src/infra/enumcount.h` provides `enumCountIsExact<E, Count>()`. It asks the compiler whether `Count-1` names an enumerator and `Count` does not, using the `__PRETTY_FUNCTION__` spelling of an enum value. A static_assert on `count == Last + 1` cannot see an append; this check can.
- The header tests itself on both polarities, so a compiler that spells enum values differently gets a compile error, not a silent pass.
- **Negative control**, compiled on a scratch copy of `model.h`: appending an enumerator to `SymKind`, `RefRole` or `LocalBindKind` without moving its count fails the build with `static assertion failed due to requirement 'enumCountIsExact<rw::SymKind, kSymKindCount>()'`. The unmodified copy compiles.
- **Clang only.** GCC's spelling was not verified, so under GCC the check is not evaluated. The macOS and Linux clang CI legs evaluate it.
- `gitoracle.h`'s `Fate` was already range-checked at the read. It gains the same compile-time proof.
- `tsprobe.cpp`'s private `kSymKindCount` now uses `rw::kSymKindCount`, and `htmlexport.h`'s shape-table pin checks against it.
- The stale `ingest.h` note that tier spans are "never cached or serialized" is corrected.

**No change to formats or versions.** The cache format, `kCacheVersion`, `kParserVer` and the qsnap scheme are all unchanged.

**Codegen, `-DNDEBUG`, Apple clang 21, arm64.**
- `loadCache`, where all the readers inline, goes from 3,936 to 3,962 instructions (+0.66%). It gains 15 compares, 11 `csel` and 4 conditional branches. Most checks fold into the existing `ok` flag with no new branch.
- `spanTiersOfFiles`, which inlines the memo load, gains 37 instructions on its per-file load path, not on the per-hit path.
- No other function in the ingest translation unit changed.

## Gates

`test/cachefuzzcheck.sh` gains two parts.

**Part 3.**
- The input is a blob where exactly ONE enum byte per field class has changed and every digest has been rebuilt.
- The walker mirrors `readFileRecord` field by field and fails if it does not land on each record's exact end.
- The enumerator counts are parsed from `src/model.h`, not written into the gate.
- **(a)** Values `count` and 255: that one record is refused (`cached_records=14` of 15), and the output is byte-identical to `--no-cache`.
- **(b)** The same mutants under the ASan binary with `--clones` raise no sanitizer report.
- **(c)** An in-range edit of the same byte is accepted (15 of 15). This proves the refusal comes from the range check, not from a digest.

**Part 4.**
- **(a)** A memo tier byte set to `count` or 255 is refused, and the output is byte-identical to `--no-cache`.
- **(b)** The same mutants raise no sanitizer report under the ASan binary.
- **(c)** The control re-labels a comment span as code. It must change the answer, which proves the memo was read and believed.

**Red, on the unfixed dev and ASan binaries at `3bf884e2`: 27 FAIL rows.**
- 20 rows: an out-of-range mutant accepted, e.g. `[enum:def.kind=10] SymKind byte past the last enumerator (count 10) was ACCEPTED: cached_records=15, expected 14 of 15`.
- `[asan:enum:def.lang.255] … src/clones.h:135:41: runtime error: shift exponent 255 is too large for 32-bit type 'std::uint32_t'`, plus that part's sweep row.
- `[stier:count]` and `[stier:255]`: the memo served a different answer.
- `[asan:stier:count] ERROR: AddressSanitizer: stack-buffer-overflow`, `[asan:stier:255] src/search.h:2174:11: runtime error: index 255 out of bounds for type 'std::uint32_t[3]'`, plus that part's sweep row.

**Green:** `cachefuzzcheck: ALL PASS`, 161 PASS, 0 FAIL, with dev and ASan binaries.

**Targeted gates, each run standalone, all rc=0:** qextractionkeycheck, qschemetripcheck (no hashed body moved), cacheisolationcheck, evictioncheck, cacheidentitycheck, cacheoffsetcheck, portablecachecheck, savecachecheck, grepfastcheck (incl. arm 6 against the unfixed binary: byte-identical on all 13 vectors), historyoraclecheck, fficheck, routeedgecheck, manifestcheck, gateexitcheck. `docs/limits_build.py --check` and `docs/gatecount_build.py --check` both pass; no gate was added.

**`--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: `gating="0"`, 3 rows.**
- `spanTierMemoLoad` complexity 17 → 18 and length 74 → 80 lines, both `sev="minor"` and both already over their bars before this change.
- `ByteR::enumU8` flagged `dead-code`, `origin="new-symbol"`. This is a false positive: ripwire records no call edge for a member call with explicit template arguments (`r.get<K>( 1 )` gives `--callers` count 0, `r.plain( 1 )` gives 1). That is a separate extraction defect needing a parser-version bump, so it is not fixed here.

**Full suite (`python3 test/pargates.py . ./build/ripwire -j 6`, local, macOS arm64, `asan/` configured).** No single run was all-green, so here is what each run found.
- **Run 1 at `b6ccba11`** (after merging `origin/main` #237): `gates=632 pass=629 skip=2 fail=1 wall=751.0s`.
  - The one failure was a true positive: `infraportcheck (B)` flagged three comment lines in the new `src/infra/enumcount.h` that named a host domain type.
  - Fixed in `b0a54fdc` (comments only). `infraportcheck` then passed standalone: `ALL PASS`.
- **Run 2 at `b0a54fdc`** (this head): `gates=632 pass=629 skip=2 fail=1 wall=797.6s`.
  - The one failure was `kotlincheck`'s `hostile nesting: 'dev' is a non-NDEBUG build, yet the Kotlin refusal raised no DEGRADED_PATH_ALERT`. That is the known flaky arm, and it passed in run 1 on code that differs only in comments.
  - This change touches no Kotlin path and not that alert site.
  - Run standalone right after the suite, at load average about 19: **5 of 5 runs rc=0, 103 PASS, 0 FAIL**, 2–4 s each.
- **Together:** every gate passed in at least one of the two runs. Run 1's only failure was a true positive that is now fixed. Run 2's only failure is the known kotlincheck flake, and it passes 5 of 5 standalone.
- **Skips:** only the two environmental ones in both runs, `argvdiffcheck` and `editchecknotecheck`. `g1freshcheck` ran and passed against a freshly rebuilt `asan/`.

## Pins moved

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache safety by rejecting corrupted records containing invalid enum values while keeping unaffected cached data usable.
  * Invalid span-tier memo data is now rejected and automatically reparsed.
  * Prevented malformed cached data from causing unsafe lookups or sanitizer errors.

* **Tests**
  * Added fuzzing coverage for corrupted and valid cache values, memo data, reparsing behavior, and sanitizer safety.


joyful-ii-V-I:
CR: no actionable comments; docstring-coverage warning declined (house style documents with `//` blocks, not docstrings).
Independent review of #241:
- F1 (CHANGELOG reachability wording) → fixed in ab615f0e: the memo is read only from the per-user 0700 owner-checked cache dir (no repository or --cache= path reaches it), and it still has no checksum, so an in-range flip is still believed.
- F2 (ASan arms lacked a liveness control) → fixed in fed5a6c6: the in-range mutants also run under ASan (Part 3 cached_records=N; Part 4 answer changes).
- F3 (search.h tierHitCount[3] / kUnclassifiedTier = 3) → fixed in e6b33f8d: both spelled with kSpanTierCount.
- F4 (enumcount.h clang-only) → declined: the 16 clang CI legs evaluate it on every PR and Part 3's controls go red on gcc; a GCC path is optional follow-up.
- F5 (memo has no integrity check) → out of scope; disclosed via F1.
- F6/F7/F8 → no change (nits, gate-covered, or owner-sanctioned markers).
Sweep: every enum built from disk bytes is covered; nothing missed.

joyful-ii-V-I:
Merged `main` in as `2aa78390` so this merges clean after #237/#239/#240. The conflict was only in CHANGELOG.md. Both [Unreleased] entries are kept, main's first. Nothing else in the lane changed, and CI runs again on the merge.

Issue #71, opened 2026-09-08
Delivery coordination: five lanes queued for main (2026-09-08)


joyful-ii-V-I:
Several rounds are running in parallel and all of them want `main`. This is the running order, what each
lane is blocked on, and the two traps that have already cost a red `main` once today. **Edit this issue as
lanes land** rather than starting a new thread.

`origin/main` = **4c10be9d** (lane 5 landed 2026-09-09, on top of lane 3's f2589419), gate count **563**.

## Order, and why

Not "who finished first". Two things serialise us; everything else is free:

1. **Generated files.** `docs/COMMANDS.md` is built by `docs/docs_commands_build.py`. Three lanes touch
   it. **Regenerate after rebasing — never hand-merge.** `docscommandscheck` arm (G) checks which capture
   produced it.
2. **Counts a gate re-derives.** The gate count (8 sites, from `regression.sh`'s loop) and the paper/repo
   counts (6 sites, from `docs/LINEAGE.md` rows).

| # | lane | state | blocked on |
| --- | --- | --- | --- |
| 1 | `lane/pr55-installer-isolation` (PR #55) | ahead 3, **behind 4** | rebase onto 26abcc79 + re-dispatch. Mine |
| 2 | `ci/std-print-floor-2026-09-08` | **LANDED** — main `88a5503b` (a `--no-ff` merge of `1095ed70`; tree identical; `git revert -m 1 88a5503b` undoes it). Proof: `48c42046` green on every job in [34293608628](https://github.com/redhat-et/ripwire/actions/runs/34293608628); `1095ed70` re-proving in [34298150602](https://github.com/redhat-et/ripwire/actions/runs/34298150602); main's own run [34298547133](https://github.com/redhat-et/ripwire/actions/runs/34298547133) **green on every job** (the branch run's one macOS red was the crawl-root-mutation flake in the standing items — same tree passed on main) | What is now on main for everyone: `--version` carries `emit=std::print` (versioncheck #6); CMakeLists' libstdc++ ignorelist exempts `<format>`/`<print>` (spelled `c\+\+`) and g1configcheck's audited list enumerates it — any new `src:` rule must move those counts; ubuntu gcc legs pin gcc-14, RHEL probe floor 14, macOS legs Xcode 16.2, every job asserts the emitter after Build; `fallback-emitter` job; the shotgun round's `--cochange`/`--situ` naming and EVALS section |
[...]

PR #244, opened 2026-09-16
fix(ts): literal receivers no longer pin unrelated same-named functions

### LINKED ISSUE #59: TypeScript: built-in member calls resolve to unrelated functions with the same name
Ripwire links `String.replace` and `Array.map` calls to unrelated functions that the calling file does not import.
The result reports zero ambiguous calls and lists both functions as callees.

Environment:

- Ripwire 0.5.0, release binary, `built_from=bacfa3b7b`.
- macOS ARM64, AppleClang 15.0.0.15000309 release build.
- Node.js v26.8.1 for runtime checks.
- Each example uses a new temporary directory and `--no-cache`.

Reproduce:

```sh
mkdir -p src
cat > src/bounded.ts <<'EOF'
export function bounded(text: string): string {
  return text.replace(/x/g, "").split(" ").map(part => part.trim()).join(" ");
}
EOF
cat > src/unrelated.ts <<'EOF'
export function map(value: number): number { throw new Error("unrelated map was called"); }
export function replace(value: number): number { throw new Error("unrelated replace was called"); }
EOF
ripwire . --no-cache --callees=src/bounded.ts:bounded
```

Actual output, with the explanatory XML comments removed:

```xml
<callees of="src/bounded.ts:bounded" defs="1" count="2" root="."
 hop_tested="0" hop_untested="2" graph_ambiguous="0" graph_unresolved="0"
 counts_floor="1" next="--expand=src/bounded.ts:bounded">
  <s t="fn" n="map" p="src/unrelated.ts:1"/>
  <s t="fn" n="replace" p="src/unrelated.ts:2"/>
</callees>
```

The command exits zero. `--expand=src/bounded.ts:bounded --top-k=0` also includes these functions in its callee signatures.

Runtime control:

```sh
node --experimental-strip-types --input-type=module -e 'import "./src/unrelated.ts"; import { bounded } from "./src/bounded.ts"; console.log(JSON.stringify(bounded("x value")));'
```

This prints `" value"` and exits zero. Neither throwing function runs.

Expected behavior:

Do not resolve these built-in member calls to unrelated top-level exports.
Leave the targets external or unresolved when the receiver cannot be proved.

The legend discloses name-based extraction and missing edges. This example adds false edges rather than omitting real edges.
The false links affect dependency inspection and can direct an agent to unrelated implementation code.

### LINKED ISSUE #163: TypeScript/JavaScript literal receivers bind a confident false edge — `"x".replace()` hits an unrelated user function (#59)
## What is wrong

In every TypeScript and JavaScript codebase ripwire indexes, a built-in method call can link to an
unrelated user function that happens to share its name. `"a-b".replace(…)`, `[3, 1, 2].map(…)` and
`/x/.test(s)` all do it. The graph's ambiguity gauge stays at zero, so the wrong edge reads as
confident, and an agent is sent to code that never runs. This is issue #59.

An earlier fix vetoed built-in method names wholesale and was backed out, because it deleted true
edges: on webpack, `--callers=stringify` went from 361 to 0. The safe first step is to decide by the
receiver's **type**, not its name — starting with receivers whose type the syntax already proves,
which is to say literals. Everything else keeps today's behaviour.

## The evidence

`test/fieldnarrowcheck.sh` carries a KNOWN GAP block on a freshly generated fixture, rewritten
minimally from #59's own reproduction:

- Five arms pin today's wrong behaviour — a string, a template, an array, a regex, and a chain
  starting from a string literal. Each binds an unrelated function in `src/unrelated.ts`, with
  `graph_ambiguous="0"`. They pass today. **Flipping them is the finish line.**
- Two controls pin true edges any fix must keep: a typed user-object receiver `r.replace()` binding
  `Rewriter.replace`, and `"x".shout()` binding the repo's own `String.prototype.shout`.
- A presence guard keeps the block from passing vacuously.

Confirm the arms pass today with `bash test/fieldnarrowcheck.sh` on a plain dev build.

## Size

**Medium.** The code is small: a receiver classification at extraction (an append-only enum, with a
version bump and its mirror) and one resolver arm. Most of the work is measurement — establishing
the *full* population of edges the change removes on a real corpus, not a sample. webpack is the
historical corpus for this one.

## Where to start

`prompts/help-wanted/ts-literal-receivers.md` is a self-contained prompt for a coding agent. It
carries why this matters, the file pointers, the rejected attempt and why it failed, the
reproductions, the design space and constraints, the acceptance criteria, the known traps — every
one of them measured on `main` while the kit was written — and what the fix PR's description should
contain. Like every prompt in `prompts/`, **it ends by writing a plan and stopping** — a maintainer
agrees the plan before any code is written.

Nothing else has to merge first; the gap is on `main` today.

Comment here to claim it.

csy20:
Fixes #163. First step on #59. Does not close #59.

Agreed plan: https://github.com/redhat-et/ripwire/issues/163#issuecomment-5664830049
Go-ahead: https://github.com/redhat-et/ripwire/issues/163#issuecomment-5700317514
Review: https://github.com/redhat-et/ripwire/pull/244#issuecomment-5702847985

## The rule

A TS/JS member call whose receiver type is syntactically certain is a call on a built-in.

- **Certain:** string / template / array / regex / number / boolean literals, plus chains that stay certain (table in `ingest_binds.h`).
- **Certainty ends** at find / at / pop / shift / reduce / subscript / `!` / `as` / `satisfies` / `<T>x`.
- **Only names that really are members of that built-in** leave the ladder (sorted per-type tables in `model.h`). Bind a JS `Foo.prototype.NAME` polyfill first (TS has no protomethod capture). Else External if the name exists in-repo, Undefined if it does not (`"x".charCodeAt()` is Undefined, not `external=`).
- **Not a member of that type** (`shout`, named `function x()`, `Object.assign`, `defineProperty`, TS `declare global { interface String { loud() } }`) keeps today's path.
- **Scope stamp** only for String / Array / RegExp / Number / Boolean.
- **Out of this PR:** object literals, `this.replace()`, `helpers.transform()`, `JSON.stringify`, typed `text: string`.
- `isMemberAccessNode` stays false for TS/JS. `!ctor.empty()` kept.

## Red first

Pre-change binary (`main` `f8e6087`): every covered literal arm still binds `unrelated.*`; `external=` absent. Seven original controls plus the new prototype-extension controls PASS on that same binary.

After: `bash test/fieldnarrowcheck.sh` ALL PASS. TS/JS/TSX corpora `external=10` (exact, not a substring), `ambiguous=0`, literal corpus warm == cold.

## Measurement

| Corpus | commit | edges | external | ambiguous | unresolved | declined | `--callers=stringify` |
| --- | --- | --- | --- | --- | --- | --- | --- |
| webpack | `a943d69c4` | 23970 → 23823 | 81 → 611 | 3535 → 3524 | 6522 → 6469 | 14954 → 14616 | **361 → 361** |
| node `lib/` | `a093ea5` | 14422 → 14413 | 0 → 10 | 1237 → 1237 | 0 → 0 | 4187 → 4187 | 2 → 2 |
| zod | `59bbc03` | 4457 → 4457 | 112 → 192 | 449 → 449 | 405 → 338 | 1295 → 1282 | (no def) |

`--callers=replace` 154 → 154. `join` 252 → 236. `slice` 632 → 630.

Webpack `--pin-census` join: **139** old-bound sites are now External. Callees: sort 46, repeat 32, join 25, exec 17, test 13, split 3, slice 3. **stringify removed: 0.** Samples: `"=".repeat(50)` had pinned CssSyntax `repeat`; `/^@/.exec(` had pinned DefinePlugin `exec`; `[signal].filter(Boolean)` had pinned streams `filter`. All false.

node `lib/` 10 removals: `regex.test` → test-runner `test`, header `.toLowerCase()` → tls helper, `[].filter(Boolean)` → streams `filter`.

## Versions

`kParserVer` 96 → 97, mirror in the same first commit. `kCacheVersion` stays 22. `qschemetrip.hash` re-pinned. Lands after #231 / #236; not chasing the number against #235.

## Other gates

`chainguardcheck`, `qualnewcheck`, `qextractionkeycheck` ALL PASS. `callformcheck` TS/JS rows PASS (G4 xmllint missing on this machine).

See #71's correction: https://github.com/redhat-et/ripwire/issues/71


joyful-ii-V-I:
Thank you, @csy20. This is the first fix for #59 that decides by the receiver's **type**, which is what the two earlier attempts could not do; the first one lost 1,704 true edges on webpack. You left `isMemberAccessNode` alone, wrote the chain table as data, and showed the arms red before the change.

I re-ran your gate against a binary built from `main` (`f8e6087c`): all 15 literal arms and `external=6` FAIL, and all seven controls PASS. A missing selector prints no `<callees` root, so the JS/TSX arms cannot pass vacuously. I also checked the 46 chain-table rows, every `RecvKind` consumer, the `qschemetrip` re-pin and `kCacheVersion` 22. All correct.

Getting the literal case right is what makes the next steps possible: typed identifiers such as #59's `text: string`, and later `JSON.stringify`. I've approved the CI run for this head, so you'll get platform results in the same round.

**Must fix before merge**

1. **Measure a population.** Use webpack @ `a943d69c4`, one prototype-heavy JS tree (e.g. nodejs/node `lib/`) and one TS tree, each at a named commit.
   - Report `edges= external= ambiguous= unresolved= declined=` before and after.
   - Get the removed edges from `--pin-census=FILE` on both binaries, join on (file, line, callee), and read each row.
   - Include `--callers=` for stringify/replace/join/slice.
   - Main's side predicts about 109 webpack sites and 164 edges go, all of which look false. But webpack has no built-in prototype assignments, so it cannot exercise item 2 or item 3. That is why the other two trees are needed.
2. **True edges are lost on prototype extensions.** Each of these binds on `main`, and this branch vetoes it, so `--callers` reads 0 on live code and dead-code/`--safe-delete` call it dead:
   - `String.prototype.x = function x(){}` (named function)
   - `String.prototype.x = x;`
   - `Object.assign(Array.prototype, { last(){} })`
   - `Object.defineProperty(String.prototype, 'x', { value: function x(){} })`
   - TS `declare global { interface String { loud(): string } }`

   Suggested rule: refuse only names that really are members of the literal's built-in type (small sorted per-type tables), still bind a scope-matched polyfill first, and let every other name keep today's path. That still flips all 15 arms. Add controls for the named-function and TS spellings. The PR body should also say that polyfill binding is JS-only, because TS has no protomethod capture.
3. **Stamp `scope` only for String/Array/RegExp/Number/Boolean** (`ingest_sidecap.h:2002–2006`). Right now every `Foo.prototype.bar` gains `sc=`, which brings a new canonical id (and so moves committed quality baseline and ack keys) and the locality tie-break for calls made from those methods. That changes output outside the literal class; node `lib/` alone has about 163 anonymous protomethods.
4. **No same-named definition anywhere → Undefined, not `external=`** (`graph.h:2299–2314`). When `byName` has no entry, the call currently goes to `vetoExternal`. The Phase-5 veto requires `it != byName.end()`, and Undefined deliberately has no gauge. Otherwise every `"x".charCodeAt()` counts as external: at least 141 webpack sites, so `external=` would go from 81 toward roughly 1,000. Set Undefined and `continue`, and add an arm for it.

**Gate (`test/fieldnarrowcheck.sh`)**

5. Tighten the gate:
   - Add number and boolean receiver arms, plus one chain link past `split` (e.g. `.join()` → String).
   - Anchor `external=` (`grep -q 'external=6'` also matches 60–69), and assert it on the JS and TSX corpora too.
   - Pin the `this.replace()` control to `p="src/user.ts:5"`. Bare `n="replace"` is also satisfied by the object literal at `user.ts:3`, and add a real free `replace` decoy.
   - Add a warm==cold arm on the literal corpus. Today it runs on `$FIX` only, so the new values' cache round trip is never exercised.

**CodeRabbit**

- Number/boolean coverage: agreed, it is item 5.
- Removing the redundant `!ctor.empty()`: please keep it. It is what stops a future literal kind without a constructor name from matching every unscoped method.
- The docstring-coverage warning is not a repo gate, so ignore it.

**Queue and shared products.** This lands after #231 and #236. You don't need to chase `kParserVer`, the `qschemetrip` hash, or the `RecvKind` append order with #235/#241/#243. Several open PRs claim the same values, and we reconcile them when each one lands.

joyful-ii-V-I:
Thanks for the quick turnaround, @csy20. `e0d5b72a` is CONFLICTING against `main`, so GitHub creates no CI run for it at all. #241 merged this afternoon and added a range check on every cached enum byte.

The only conflict is `src/model.h`. `RecvKind` now has a count constant right below it:
```cpp
inline constexpr std::size_t kRecvKindCount = static_cast<std::size_t>( RecvKind::ElixirSelfModule ) + 1;
static_assert( enumCountIsExact<RecvKind, kRecvKindCount>(), "..." );
```
When you merge `main` (a merge commit is fine; please don't force-push), keep your appended `Lit*` values and point `kRecvKindCount` at your LAST new enumerator. The `static_assert` fails the build if it still names `ElixirSelfModule`. `ingest_cache.h` auto-merges.

After that, post a short note with the population numbers from item 1 and I'll approve the CI run. kParserVer and the qschemetrip hash are still ours to reconcile at landing.

csy20:
Merged `main` (`32336a83`, including #241 and #258). Merge commits, no force-push.

`RecvKind` still ends at `LitBoolean`; `kRecvKindCount` names that enumerator. `ingest_cache.h` auto-merged. `kParserVer` stays 97.

Item 1 (same numbers as the PR body; the merge does not change the literal rule):

| Corpus | commit | edges | external | ambiguous | unresolved | declined | `--callers=stringify` |
| --- | --- | --- | --- | --- | --- | --- | --- |
| webpack | `a943d69c4` | 23970 → 23823 | 81 → 611 | 3535 → 3524 | 6522 → 6469 | 14954 → 14616 | **361 → 361** |
| node `lib/` | `a093ea5` | 14422 → 14413 | 0 → 10 | 1237 → 1237 | 0 → 0 | 4187 → 4187 | 2 → 2 |
| zod | `59bbc03` | 4457 → 4457 | 112 → 192 | 449 → 449 | 405 → 338 | 1295 → 1282 | (no def) |

`--callers=replace` 154 → 154. `join` 252 → 236. `slice` 632 → 630.

Webpack `--pin-census` join: **139** old-bound sites now External. **stringify removed: 0.**

joyful-ii-V-I:
Thank you, @csy20. This round answered every point, and the member-table rule is the right shape: a literal's type decides, and a name the type doesn't have keeps today's path.

We built `main` (`32336a83`) and this head (`d4fd5ccb`) and re-measured webpack @ `a943d69c4`. Your numbers reproduce exactly: edges 23970→23823, external 81→611, ambiguous 3535→3524, unresolved 6522→6469, declined 14954→14616, `--callers=stringify` 361→361.
- We diffed `--pin-census` on both binaries and read all 139 removed rows. Every one is a built-in call on an array, string or template literal (or a certain chain) that main had pinned to an unrelated helper: `SortableSet.sort`, `util/fs.js`'s `join(fs, root, file)`, `DefinePlugin.exec`, test-file `test`/`repeat`. join 252→236 and slice 632→630 are those false edges going away.
- The other 391 new `external=` sites were `declined=`/`unresolved=` before, and none of them is a name with no in-repo definition.
- No other census row changed anywhere in the corpus.
- Your five prototype-extension controls, the polyfill bind and the `sc=` restriction all check out on our build too. `fieldnarrowcheck` is 69/69 on this head, and 27 arms are red on main. A cache written by the old binary is rejected.

**Nothing blocking.** One tiny thing we'll fold in at landing, so you don't need to push: the `RecvKind` comment in `src/model.h` still says "anything else is vetoExternal".

**CodeRabbit comments you can ignore:** its numeric/boolean comment is now covered by your new arms, and the `!ctor.empty()` and docstring notes were answered last round.

Queue: #244 lands through the maintainers' integration train that follows the one now being built. That train merges `main` and takes the next free parser version plus the `qschemetrip` re-pin, so there's nothing for you to push or re-number. CI on this head is running and was approved after a check of the changed test files.

joyful-ii-V-I:
This PR is part of integration train 2b, #275, and will show as merged when that train lands.

Issue #157, opened 2026-09-11
A file refused for pathological nesting becomes invisible — no `--skipped` row, nothing on a warm run, and `--match` parses it anyway


joyful-ii-V-I:
## What is wrong

ripwire deliberately refuses pathologically nested JSON, YAML and Markdown before parsing. For YAML
and Markdown the refusal protects memory: those vendored grammars' scanners write past a 1024-byte
buffer at roughly 250 levels deep. The refusal is right. Its visibility is not — it is one line of
stderr on cold runs only, `--skipped` never lists the refused file, and a warm run (the default,
because ripwire caches per root) prints nothing at all. Worse, `--match` parses the refused file
anyway, in the same run that refused it, so the two-layer safety design quietly becomes one layer.
An agent asking "why is this file missing from the map?" gets no answer.

## The evidence

Two gates carry KNOWN GAP blocks pinning exactly this, each on its own generated fixture — a
300-level `deep.yml` and a 300-marker `deepquote.md`. Four arms apiece:

- a warm run's stderr says nothing about the file;
- cold `--skipped` has no row for it;
- warm `--skipped` has no row for it;
- `--match='(block_mapping_pair)'` (and `--match='(block_quote)'` for Markdown) returns hits inside
  it, in the same run whose ingest refused it.

A presence guard — cold refusal note, warm map equal to cold, sibling file indexed — keeps each
block from passing vacuously, and the `--match` arm is judged only on a clean exit so a crash cannot
pass it. All arms pass today. **Flipping them is the finish line.**

Confirm on a plain dev build with `bash test/yamllangcheck.sh` and `bash test/mdsectioncheck.sh`.

## Size

**Medium**, in two parts that land separately: the `--skipped` rows with the warm-cache fix, and a
single shared refusal predicate applied at the `--match`, `--pattern` and span-tier parse sites. The
tricky parts are making an old cache stop hiding refused files, and choosing the disclosure
attribute for `--match`. Expect a cache version bump with its mirror.

## Where to start

`prompts/help-wanted/nesting-refusals-visible.md` is a self-contained prompt for a coding agent. It
carries the file pointers — the three prescans, where they run relative to the cache-hit path, and
every unguarded parse site — plus the reproductions, the design space and constraints, the
acceptance criteria, the known traps, and what the fix PR's description should contain. Like every
prompt in `prompts/`, **it ends by writing a plan and stopping** — a maintainer agrees the plan
before any code is written.

[...]

PR #245, opened 2026-09-16
fix(diagnostics): a notice written in several writes could be split across lines by another thread, and kotlincheck §12 flaked on it

joyful-ii-V-I:
Fixes the kotlincheck §12 flake seen on #235/#240 CI.

## The defect

`Diagnostics::ConsoleLog::handleDegraded` built the `[math degraded]` notice from nine `std::cerr <<` insertions. With stdio sync on (the default), each insertion is its own `fwrite` on an unbuffered stderr, so its own `write(2)`. The assert, panic and thread-violation banners had the same shape: 15 to 21 writes each.

kotlincheck §12's fixture refuses two Kotlin files at once, so two parse workers reach the refusal microseconds apart. One takes the once-per-site latch and starts the nine writes. The other prints its own `[ripwire] …: kotlin string-template nesting > 128 levels` line in one write. When that line lands straight after `[math degraded] `, the arm's one-line grep `math degraded.*kMaxKotlinStringNestDepth` fails with "raised no DEGRADED_PATH_ALERT". The alert is still on stderr, whole, one line further down.

Evidence (root-cause run on f8e6087c, §12's map over its own fixture):
- **18 gate failures in 5,700 runs.** The notice was torn in 32–73% of runs depending on load. The gate only sees a tear at one of the eight insertion boundaries, which is why its failure rate was so much lower.
- **A probe that formatted the notice first and wrote it once: 0 failures and 0 tears in 3,800 runs.** 2,300 of those ran paired against the old code under the same load.
- Ruled out: a second refusal path, the latch (one `_dpaSeen` object, exactly one alert per run in 6,300 runs), and capture loss (every torn stderr was 814 bytes, the size of an intact one).
- CI: the same FAIL line in 8 jobs since Kotlin landed, 3 of them on `main`, all on plain (non-NDEBUG) legs across gcc, clang and appleclang.

## The fix

Every reporter (`handleAssert`, `handlePanic`, `handleThreadViolation`, `handleDegraded`) now calls one writer, `writeNotice`, exactly once. The writer:
- formats the whole notice into a fixed 4,096-byte stack buffer with `rw::formatTo`;
- hands it to stderr in **one stdio call** (`rw::emitRaw`);
- then calls `fflush(stderr)`.

Other properties:
- `noexcept`, `[[noreturn]]`, the trap and the abort are all unchanged. No header changed.
- **No heap.** A reporter may be running because memory ran out. `rw::emitTo` was not used because `std::print` and the `std::format` fallback both build a `std::string`.
- **Byte-identical text** for every notice under the cap. A longer notice is cut at a UTF-8 boundary and ends in ` ... [notice truncated: N bytes]` (a new cap, `kNoticeByteCap`, registered in LIMITS.md as OUTPUT).
- A null C-string argument prints as empty instead of reaching `std::format`.

**Why `emitRaw` and not a raw `write(2, buf, n)`, as asked.** `emitTo(stderr, …)` is one `fwrite`, but it allocates. `emitRaw` is one `fputs`, which holds the `FILE` lock for the whole call, so no other stdio writer in the process can interleave, at any length. Every other stderr writer in the process goes through stdio. Underneath, the unbuffered stderr passed the whole notice to the kernel as one `write(2)`: measured on macOS up to the full buffer (4,093 bytes, one datagram), and glibc's unbuffered path does the same by reading its source. A raw `write(2)` would add nothing in-process, and it would step *outside* the stdio lock, so a long line another thread writes through stdio could be split by the notice instead. Across processes, one write to a pipe is atomic only up to `PIPE_BUF` (512 on macOS), with either spelling.

## Gate: `test/diagnoticecheck.sh` (new), red first

`test/diagnotice_harness.cpp` runs each reporter in a forked child whose fd 2 is an `AF_UNIX SOCK_DGRAM` socket. A datagram socket keeps write boundaries, so the parent counts the `write(2)` calls a notice took. This red has no timing in it and needs no product hook: the real iostream/stdio stack runs.

| arm | old reporters (f8e6087c) | new |
|---|---|---|
| (S) static: each `ConsoleLog::handle*` calls `writeNotice` once, makes no stream call; writer makes exactly one | red ×4 (11–25 stream tokens each) | green |
| (S′) control: real file mutated (writer call duplicated; stream call injected into a reporter) | — | both seen |
| (W) one write, byte-exact, correct ending, 7 cases | red ×7: `writes=9/18/15/15/15/21/18`, **every `exact=1`** | `writes=1 exact=1` ×7 |
| (W′) control: harness rebuilt against the duplicated-writer mutation | — | `writes=2` |
| (L) over-long notice: one line, marker, valid UTF-8, own opening | red (no marker) | green (4,093 B, 1 write) |
| (T) 4 reporter threads × 3,000 notices vs 2 stdio + 2 raw `write(2)` competitors | red in 200/200 harness runs; 2,201 and 11,417 torn lines in two gate runs | `torn=0` |
| (A) no heap: `src/alloccount.cpp` delta, with vs without 110 calls (55 over-long) | green (guards the rewrite, not the tear) | equal (`1 9008` both) |
| (A′) control: writer rerouted through `rw::emitTo` | — | allocs 4 → 169 |
| (Z) W, L, T under ASan+UBSan | red | green |

Final gate against the old `diagnostics.cpp`: 27 FAIL, 2 PASS. Against the new one: 29 PASS, `rc=0`. The same gate is also green with Homebrew clang 22. The harness is also clean under the full G1 flags (`address,undefined,integer,float-divide-by-zero,float-cast-overflow -fno-sanitize-recover=all`).

**Byte identity.** Every (W) case matched text the harness spells out independently, on the old build and on the new one. The reassembled bytes of all seven cases from the two builds are identical under `cmp`. The over-long case's kept text is a byte prefix of the old 9,148-byte notice, and its marker names 9,148.

## kotlincheck §12

- The arm keeps its single-line grep (no newline stripping, which would hide exactly this defect), now as if/else.
- On a FAIL it prints `head -5` of stderr, because none of the eight CI logs could show the torn text.
- Observed red with a shim that tears the notice: exactly one FAIL row, with the torn stderr printed.

## The §12 loop, after the fix

§12's map over its own fixture (`--no-cache`), alternating run by run with the f8e6087c binary, under 4 niced busy loops (PIDs recorded, killed by trap; `ps` afterwards shows none left):

| binary | runs | gate red | torn notice |
|---|---:|---:|---:|
| this branch | 2,100 | **0** | **0** |
| f8e6087c | 2,100 | 6 | 726 |

## Gates run (each as its own script, `RIPWIRE_BIN=./build/ripwire`, binary `built_from=` the head)

All 35 rc=0, 0 FAIL rows, on a binary `built_from=6a8aee482` (HEAD):

| gate | rc | PASS rows | wall |
|---|---|---:|---:|
| kotlincheck | 0 | 104 | 3s |
| diagnoticecheck | 0 | 30 | 14s |
| jsonlangcheck | 0 | 21 | 1s |
| yamllangcheck | 0 | 52 | 1s |
| prconvergecheck | 0 | 35 | 4s |
| estchargecheck | 0 | 214 | 58s |
| gateexitcheck | 0 | 23 | 7s |
| manifestcheck | 0 | 16 | 12s |
| gatecountcheck | 0 | 7 | 1s |
| limitstablecheck | 0 | 24 | 7s |
| capsweepcheck | 0 | 32 | 3s |
| fixedbufsweep | 0 | 11 | 1s |
| infraportcheck | 0 | 12 | 4s |
| readmedriftcheck | 0 | 47 | 14s |
| versioncheck | 0 | 13 | 0s |
| printffmtparitycheck | 0 | 43 | 6s |
| qualitystalecheck | 0 | 38 | 8s |
| churnjoincheck | 0 | 72 | 25s |
| localscountcheck | 0 | 20 | 2s |
| namedfileinputcheck | 0 | 70 | 2s |
| preproccondcheck | 0 | 51 | 106s |
| refusaltailcheck | 0 | 18 | 1s |
| sincecheck | 0 | 77 | 10s |
| w3fixlegendcheck | 0 | 72 | 36s |
| noaliascheck | 0 | 39 | 5s |
| connectcorecheck | 0 | 31 | 9s |
| clonelexcheck | 0 | 27 | 2s |
| columnarcommacheck | 0 | 11 | 3s |
| type3clonecheck | 0 | 41 | 9s |
| xmlwellformed | 0 | 37 | 29s |
| ripwirepubliccheck | 0 | 16 | 41s |
| docanchorcheck | 0 | 29 | 4s |
| forrankordercheck | 0 | 13 | 19s |
| showcasecapturecheck | 0 | 24 | 11s |
| binoverridecheck | 0 | 5 | 112s |

Five SKIP rows, each read:
- estchargecheck #10: tiktoken is not installed.
- gateexitcheck (E): that arm only runs with `GATEEXIT_FULL=1`.
- churnjoincheck G2, preproccondcheck (600-deep guard) and w3fixlegendcheck: each says "alerts compiled out (NDEBUG)" on this PLAIN build. **Pre-existing and not caused by this change.** The same three skips appear with the f8e6087c main binary. churnjoincheck's probe (`--rank-by=churn --since=notadate`) is now refused before any degrade path on both binaries, so it can never see an alert. Those alert arms are not asserting anything on any flavour. This needs its own fix, outside this PR.

quality-delta over `f8e6087c..HEAD` (the merge-base range): `gating="0"`. One row was acked through the binary (`--ack-only=gating`, exactly one `+ack` line). It is a 19-token clone pairing the harness's one-line `expectedDegraded` with `handoff.h`'s `handoffSymsCapClause`, which share only the `return std::format( literal, args )` shape. The remaining row is non-gating: dead-code on `exitOnSignal`, which is reached as a signal handler, not called. An earlier revision of the harness had its own counting `operator new`; that cloned `src/alloccount.cpp`, so it now reuses that instrument.

Not run locally: a Linux leg. The local Docker VM is out of disk space and I did not prune it. The glibc single-write claim is from reading `_IO_new_file_xsputn`; CI's ubuntu legs run the gate.

## Pins moved

- Gate count 618 → 619 (`docs/gatecount_build.py` over README.md, docs/EVALS.md, present/deck5_ripwire_build.js).
- `test/fixedbufsweep.sh` EXPECTED: mentions 328 → 331, calls/sites 221 → 223, rows 94 → 96. Two new not-markup TABLE rows (`notice[4096]`, `marker[64]`), re-read from the diff.
- `docs/LIMITS.md` regenerated: 213 caps (`kNoticeByteCap`, OUTPUT in `docs/limits_classes.tsv`). README "213 compile-time caps".
- `docs/TUNING.md` regenerated: census 131 declarations / 130 names.
- Registrations: `test/regression.sh`, `.github/pargates-shard-weights.json` (25.0 s), `test/binoverridecheck.sh` EXEMPT (the gate binds no ripwire binary).
- `.ripwire_quality_acks`: +1 row (above).
- `test/printf_parity.manifest`: none (notice bytes unchanged).

## Not fixed here

The default map over §12's fixture says `files=4 symbols=3` and gives no sign of the two refused files (no `nest_refused=` on the header). A reader of the map learns about the refusal only from stderr, or from `--skipped` if they think to ask. That disclosure gap belongs to #157 and the DISCLOSE work, not to this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Diagnostic notices and failure reports now appear as complete, non-interleaved lines during concurrent output.
  - Long notices are safely truncated with a byte-count indicator while preserving valid UTF-8.
  - Diagnostic reporting handles missing input text safely.

- **Tests**
  - Added coverage for concurrent output, truncation, write boundaries, allocation behavior, and sanitizer checks.

- **Documentation**
  - Updated gate and configuration-limit counts across the README, guides, changelog, and presentation materials.


joyful-ii-V-I:
Review of 6a8aee48 → fixed in 1e8a6462 (new commits after the reviewed head: 606d47b7 merge of main, 1e8a6462 fix):
1. diagnoticecheck (A) red on ubuntu was the harness, not the reporters: its "alloc calls=" line exceeded libstdc++'s 15-B SSO under g++'s std::format arm. Now formatted into a stack buffer + emitRaw; the gate header prints CXX and the emitter.
2. Merged main in 606d47b7 (CHANGELOG [Unreleased] kept both entries, main's first; `git diff origin/main --numstat -- CHANGELOG.md` is 27 insertions, 0 deletions); generators re-checked.
3. writeNotice now fflushes stdout first, as std::cerr's tie did; byte identity extended to a stdout+notice 2>&1 case.
4. (Z) prints SKIP (not WARN) when the sanitizer build is unavailable; (A)'s row says it counts operator new; truncation marker now "kept K of N bytes".
Declined: CodeRabbit docstring-coverage warning (house style uses // comments).
Verified in review: one write(2) on glibc (writes=1 at 4,093 B), byte identity 7/7, pins match the generators, gate security CLEAN.

joyful-ii-V-I:
Merged `main` in as `77b944e5` so this merges clean after #241 and #247. The conflict was only in CHANGELOG.md; both [Unreleased] entries are kept, main's first. CI was 31/31 green on `1e8a6462` and runs again on the merge.

joyful-ii-V-I:
`8b66f253` moves this PR's CHANGELOG entry to the top of [Unreleased]. Several open PRs were adding their entries at the same line, so every landing re-conflicted the rest. No other change.

PR #246, opened 2026-09-16
ci(macos): the shipped arm64 binary was compiled by an LLVM whose loop vectorizer ignores the no-alias promise, at a minimum macOS nothing pinned

joyful-ii-V-I:
## The defect

The `macos-arm64` release asset and every macOS CI leg were built with Xcode 16.2 on `macos-14`. That Xcode's AppleClang 16 is LLVM 17. LLVM 17's loop vectorizer never reads `__builtin_assume_separate_storage`: llvm/llvm-project#64666, fixed by #76770 in LLVM 18. CI showed this on PR #200, where `noaliascheck` arm 6 stayed at plain on `macos-14`. So a `VERIFY_NO_ALIAS_BUF` promise in the shipped macOS binary removed scalar reloads, but every vectorized loop kept its runtime overlap check and its scalar fallback. The ASSUME survey counted 27 loops that become check-free on Apple clang 21. None of them could become check-free in the binary users download.

Two further problems:
- **No deployment target on the arm64 leg.** #240's report says the pin and the `otool` minos check only ever applied to the x64 leg. clang takes the lower of the runner's macOS and the SDK default. That gave 14.0 on `macos-14`, but a plain runner move would give 26.x and drop macOS 14 and 15 users without a word.
- **`macos-14` is deprecated.** actions/runner-images#13518 schedules brownouts on eight October dates and retirement on 2026-11-02. On those dates every macOS job in both workflows fails.

## The change

- **release.yml, macos-arm64:**
  - Runs on `macos-26` with `developer_dir` Xcode 26.6.
  - `deployment_target: "14.0"`, exported as `MACOSX_DEPLOYMENT_TARGET` before the PGO build and read back off `build/ripwire` with `otool` after PGO staging and before Package. #240 removed those two steps with the x64 leg; they come back for this leg.
  - Both steps run on every macOS leg and fail loudly instead of skipping. A leg with no `deployment_target` fails at the export, and the read-back is an exact string compare, so a missing or doubled `LC_BUILD_VERSION` also fails.
  - A new step records `xcodebuild -version`, `c++ --version` and the paths of `clang++` and `llvm-profdata`. It fails if `DEVELOPER_DIR` is empty (an empty value would turn the `"$DEVELOPER_DIR"/*` prefix test into `/*`) or if either tool resolves outside it. Training, merge and optimize then use one LLVM, with no `.profraw` version skew.
- **ci.yml, TESTED == SHIPPED:**
  - The eight `release` gate shards and the `asan` macOS leg run on `macos-26` with `DEVELOPER_DIR` Xcode 26.6 and `MACOSX_DEPLOYMENT_TARGET` 14.0.
  - `asan`'s `ASAN_OPTIONS` condition moved too; otherwise the macOS leg would have got `detect_leaks=1`.
  - The job env's `matrix.os == 'macos-26' && … || ''` falls back to empty on any other label. So the build matrix's front-end step and the asan macOS configure now fail when `DEVELOPER_DIR` or `MACOSX_DEPLOYMENT_TARGET` is empty, or when `CMakeCache.txt`'s `CMAKE_OSX_DEPLOYMENT_TARGET` is not the exported value.
  - The Release/plain split, the four-shard layout and the plain-leg skiplist are unchanged.
- **Comments** that described the current toolchain as Xcode 16.2 / `macos-14` now say "through 0.6.1": `CMakeLists.txt`, `cmake/PortableFlags.cmake`, `CONTRIBUTING.md`, the header of `test/portablebuildcheck.sh`, and one `optremarkscheck` skip message. Comments that cite a historical CI run on `macos-14` are left as they were.

## Evidence for Xcode 26.6 on macos-26

- **The image.** `images/macos/macos-26-arm64-Readme.md` at image 20260907.0351.1 lists Xcode 26.6 (default), build 17F113, and "Clang/LLVM 21.0.0". It also carries Xcode 26.0.1 through 26.5. `toolset-26.json` has `"default": "26.6"` with `26.6+17F113`. `macos-26` is the arm64 label; `macos-14`'s image still lists Xcode 16.2 (16C5032a) as non-default, with 15.4 the default.
- **The same build is on the dev machine.** `/Applications/Xcode.app` is 26.6 build 17F113, and `xcrun clang++ --version` reports Apple clang 21.0.0 (clang-2100.1.1.101). So every local measurement below ran on the exact compiler the runner image ships.
- **The loop vectorizer reads the hint with no extra flags.** The probe is a `d[i] += 3*s[i]` loop over two `uint32_t*`, built with `-O2 -mcpu=apple-m1 -mmacosx-version-min=14.0` and counted with objdump:

  | build | without the promise | with `__builtin_assume_separate_storage` |
  | --- | --- | --- |
  | default | 64 | **57** |
  | `-mllvm -basic-aa-separate-storage=false` | 64 | 64 |

  Under `-mllvm -vectorize-memory-check-threshold=0`, the plain loop reports `AnalysisAliasing/CantReorderMemOps` (it needs a runtime check) and the promised loop reports `Passed/Vectorized`.
- **`test/noaliascheck.sh` on this compiler:** `CONSUMED_DEFAULT (default: 0, =true: 0, =false: 1)`, `LOOP_CONSUMED (axpyBuiltin 61 vs plain 65)`, and arm 6/8 control `with the analysis off axpyBuiltin is 65`. It printed ALL PASS. On the new macOS legs arm 6 becomes a hard assertion instead of a `LOOP_NOT_CONSUMED` WARN.

## The minimum macOS: 14.0, read off v0.6.1

`gh release download v0.6.1 --pattern 'ripwire-0.6.1-macos-arm64.tar.gz*'` fetched the tarball (7,941,688 bytes) and `shasum -c` passed. The binary was read, never executed. `otool -l` on `ripwire-0.6.1-macos-arm64/ripwire` shows `LC_BUILD_VERSION`, `platform 1`, **`minos 14.0`**, `sdk 15.2`. README's "Prebuilt macOS floor" row already says macOS 14 or later.

`<print>` at that target with Xcode 26.6: `xcrun --sdk macosx clang++ -std=c++23 -mmacosx-version-min=14.0` compiles and runs a TU that `#error`s without `__cpp_lib_print` and calls `std::print`/`std::println`. It reads `minos 14.0 sdk 26.5`. The control at `-mmacosx-version-min=13.0` hits the `#error`, so the macro really is gated on the target.

## Release leg, emulated locally on the same Xcode build

`DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer` (26.6, 17F113) and `MACOSX_DEPLOYMENT_TARGET=14.0`, then `scripts/pgobuild.sh --cmake-extra -DCMAKE_BUILD_TYPE=Release`. It finished rc=0 in 185 s: one `.profraw` was merged by `xcrun --find llvm-profdata`, which sits in the same `XcodeDefault.xctoolchain/usr/bin` as `clang++`. `build_pgo/CMakeCache.txt` shows `CMAKE_OSX_DEPLOYMENT_TARGET=14.0`, `RIPWIRE_LTO=ON`, `RIPWIRE_PGO=use`, `RIPWIRE_CXX_HAS_BASIC_AA_SEPARATE_STORAGE=1` and `RIPWIRE_LD64_HAS_BASIC_AA_SEPARATE_STORAGE=1`.

The release post-steps, run on `build_pgo/ripwire`:
- PGO determinism diff: identical.
- `ripwire 0.6.1 (Release, AppleClang 21.0.0.21000101, emit=std::print, …)`.
- `otool`: `minos 14.0`, `sdk 26.5`.
- Smoke `xmllint --noout`: clean.
- Output byte-identical to the plain `build/ripwire` on `test/fixture`, on the repo map, and on `--for="resolve references into the call graph"`.

The three new step scripts were extracted from `release.yml` and run with `bash -e`, each against good and bad inputs:
- **pin:** writes `MACOSX_DEPLOYMENT_TARGET=14.0` to `$GITHUB_ENV`. An empty `deployment_target` gives rc=1, naming the leg.
- **record:** rc=0 with both tool paths under the Xcode. An empty `DEVELOPER_DIR` gives rc=1.
- **minos** on `build_pgo/ripwire`: rc=0 at `14.0`. It gives rc=1 against `15.0`, against a missing binary (`minos ''`), and against `14`.

## Gate: test/portablebuildcheck.sh #2i (sixteen rows, text-level, any host)

The arm checks the following. The first seven are verdict rows about the release leg:
- the release leg's exact runner, Xcode and quoted `"14.0"`
- a `MACOSX_DEPLOYMENT_TARGET` export before the first configure that refuses an empty target
- one deployment-target source: no `-DCMAKE_OSX_DEPLOYMENT_TARGET`, `-mmacosx-version-min` or second `MACOSX_DEPLOYMENT_TARGET` anywhere in the leg or in the build job's env and steps
- the exact, fail-closed `otool` minos compare between PGO staging and Package
- the toolchain record step ahead of the first build: empty-`DEVELOPER_DIR` refusal, `pwd -P`, outside-prefix `exit 1`
- `DEVELOPER_DIR` wired from `matrix.developer_dir`
- every one of those steps runs on the leg: `if:` absent or `runner.os == 'macOS'`, never `matrix.deployment_target`

In ci.yml it checks:
- every macOS runner label, `matrix.os == 'macos-*'` condition, Xcode path and `MACOSX_DEPLOYMENT_TARGET` line equals the release's
- the `build/` and `asan/` CMakeCache.txt pin guards both exist

Three built-in controls must each be refused by exactly their own row: no minos step, `ASAN_OPTIONS` back on `macos-14`, and a step-level `-DCMAKE_OSX_DEPLOYMENT_TARGET=15.0` on pgobuild. Each way a control can break is its own named FAIL.

- **Red on #240's head `6fcae7bc`** (`git archive` + this gate): 14 FAIL, 2 PASS.
- **Green:** 16 PASS (35 in the gate), under both an unset locale and `LC_ALL=en_US.UTF-8`.
- **Ad hoc mutations**, each red on exactly one row:
  - the export `if: matrix.deployment_target`
  - the record step deleted, or moved after PGO
  - the empty-target guard deleted
  - minos neutered to `= '99.0'`, or reverted to the regex grep
  - a step env `MACOSX_DEPLOYMENT_TARGET: "15.0"`
  - `-mmacosx-version-min=15.0`
  - an inline `MACOSX_DEPLOYMENT_TARGET=15.0 scripts/pgobuild.sh`
  - `pwd -P` removed
  - the empty-`DEVELOPER_DIR` guard deleted
  - either ci.yml cache guard deleted
  - plus the earlier ones: unquoted `14.0`, `"15.0"`, Xcode 16.2, a CI shard on `macos-15`, a CI Xcode path

## The first CI run's red, and the sort-locale sweep

Run 35130688947 (d7541a66) failed one row on both macos-26 shards: `scroundtripcheck (B) cppqualfix: composed ids differ`. The ids were right; the gate harness was the problem.
- **Cause:** the gate sorted its expected list with `/usr/bin/sort` and `cmp`'d it with Python `sorted()`. macOS 26's sort, under the runner's `en_US.UTF-8`, collates case-insensitively (`qual.cpp::a::…` before `qual.cpp::Box::…`).
- **Reproduced** with the unchanged gate: `LC_ALL=C` ALL PASS; `LC_ALL=en_US.UTF-8` gives the CI row, `1,5d0 < qual.cpp::a::scopedTmpl …`.

The same class, swept across every `sort`, `comm`, `join`, `uniq` and `ls` call in `test/`, `test/lib/`, `scripts/` and `bench/`: 296 unpinned sites classified. 27 sites in 20 files compare an order with a different one: Python `sorted()`, a literal, a pinned hash, ripwire's byte-sorted output, or `git status`. All 27 now run under `LC_ALL=C` (7133a681 lists them).
- **Only `scroundtripcheck` fails today.** The pre-fix `columnarcheck` and `mcpforparitycheck` pass in both locales, so those are latent.
- **Every fixed gate is ALL PASS** under both `LC_ALL=C` and `LC_ALL=en_US.UTF-8`.

## Gates run (local, Apple clang 21.0.0 = Xcode 26.6 17F113; plain build, `MACOSX_DEPLOYMENT_TARGET=14.0`)

The branch merged `main` at a55b118e (#240, #237, #239); none of those changed a file this PR touches. It was rebuilt `--clean-first` after the merge and relinked at every later commit: `ripwire 0.6.1 (dev, AppleClang 21.0.0.21000101, emit=std::print, built_from=7b35b96fc)`, `minos 14.0`.

```
portablebuildcheck    rc=0 pass=35 fail=0  ALL PASS   (#2i: 16 rows; unset locale and en_US.UTF-8)
scroundtripcheck      rc=0 pass=24 fail=0  ALL PASS   (LC_ALL=C and LC_ALL=en_US.UTF-8)
releaseinstallcheck   rc=0 pass=55 fail=0  ALL PASS
versioncheck          rc=0 pass=12 fail=0  ALL PASS   (at the head commit)
manifestcheck         rc=0 pass=16 fail=0
gateexitcheck         rc=0 pass=22 fail=0 skip=1 (arm E, ground truth not requested)  ALL PASS
headbinstagecheck     rc=0 pass=26 fail=0  ALL PASS   (parses ci.yml's staging step)
pargatescheck         rc=0 pass=64 fail=0  ALL PASS   (reads ci.yml's --budget-scale)
optremarkscheck       rc=0 pass=30 fail=0  ALL PASS
noaliascheck          rc=0 pass=38 fail=0  ALL PASS  CONSUMED_DEFAULT / LOOP_CONSUMED
qschemetripcheck      rc=0 pass=5  ALL PASS   (main's #237 pin, on the merged tree)
docdemotecheck        rc=0 pass=20  ALL PASS
deckcheck             rc=0  ALL PASS
readmedriftcheck      rc=0 pass=46  ALL PASS
commentcoherencecheck rc=0 pass=8   ALL PASS
ripwirepubliccheck    rc=0 pass=15  ALL PASS
gatecount_build.py --check  rc=0 (loop names 618; no gate added)
limits_build.py --check     rc=0
det diff test/fixture: identical; xmllint --noout: clean
--quality-delta=$(git merge-base origin/main HEAD)..HEAD   (a55b118e..HEAD)
                            regressions="3" minor="3" acked="2" gating="0"
```

`actionlint`, with shellcheck over every `run:` block, reports 0 errors on `release.yml` and `ci.yml`. main had 3 (`property "cmake_extra" is not defined`), left by #240, which removed the only leg that set `cmake_extra` but kept the two expressions that read it. Commit `d7541a66` removes both; they evaluated to the empty string on every remaining leg, so the configure and `pgobuild.sh` commands each leg runs are unchanged.

Two quality-delta rows gated, both verbosity rows for the release guards: `release.yml`'s `build` job (138 to 196 lines) and its `jobs` key (218 to 276). Both are acked through the binary (`--ack-only=gating`), and each ledger row was rewritten in place with its prior reason kept. Three minor verbosity rows on ci.yml's job keys stay visible.

The full suite was not run locally. CI carries it, and the macOS legs are what this PR changes.

## Risks worth watching on the first CI run

- **Runner pool.** `macos-26` is the same free arm64 pool class as `macos-14`, and 0.6.1's macos-x64 release leg already ran there. The starvation trap still applies: nine macOS jobs per run, and a TIMEOUT on a macOS shard is read against `xmlwellformed`'s runner-speed probe, not taken as a code signal.
- **Rosetta arms.** `strkerncheck` probes Rosetta 2 for AVX2. On `macos-14` it read `rosetta_no_avx2` and ran the x86-64 baseline slice. macOS 26's Rosetta runs the x86-64-v3 slice, as the dev machine and the 0.6.1 x64 leg did, so the macOS CI legs now execute the v3 slice under translation. That path was never exercised on a `macos-14` runner.
- **The release leg itself** first runs at the next tag. Its PGO and minos steps were emulated locally on the same Xcode build, above.

Pins moved: none.


joyful-ii-V-I:
Independent review of d7541a66 → fixed in 7b35b96f (7133a681 gates sweep, 4f679ede #2i, 7b35b96f quality re-ack). main a55b118e is merged, in 52403644.

1. **The `scroundtripcheck` (B) failure on macos-26 was the gate harness, not ripwire.** Under a UTF-8 locale, macOS 26's `/usr/bin/sort` collates case-insensitively, and the gate compared its output with Python `sorted()`. The gate now uses `LC_ALL=C sort`. The same pattern was swept across `test/`: 27 sites in 20 files fixed, listed in 7133a681. Only this gate failed today; the other 26 were latent.
2. **#2i now scans the release leg AND its steps for a deployment-target override** (CodeRabbit's finding), with a control.
3. **#2i no longer accepts `if: ${{ matrix.deployment_target }}` as always-run.**
4. **#2i now also asserts the fail-loudly guards exist**, and the stale macos-14 comments are updated.

Verified OK in review:
- The pin reaches both pgobuild configures.
- minos covers the only Mach-O.
- llvm-profdata comes from the same Xcode.
- No macos-14 runs-on/conditions remain.
- strkerncheck is green on macos-26.

joyful-ii-V-I:
Merged `main` in as `b62d1853` so this merges clean after #241 and #247. The conflict was only in CHANGELOG.md; this lane's entry sits between #240's and #241's, a different spot from the other open PRs' entries. CI was 31/31 green on `7b35b96f` and runs again on the merge.

PR #248, opened 2026-09-16
fix(resolve): a member call through a typed parameter was pinned to the caller's own class

joyful-ii-V-I:
## The defect

On `main` (`f8e6087c`), a C++ member call through a function **parameter** was bound to the *enclosing* class's same-named method instead of the parameter type's method. The result was one precise edge, with no `amb=` and nothing disclosed:

```cpp
struct Target { int pick( int n ) { return n; } };
struct Decoy
{
    int pick( int n ) { return n; }
    int plainCaller( Target& other ) { return other.pick( 1 ); }   // --callees: Decoy::pick (c.cpp:4) — wrong
    int localCaller() { Target other; return other.pick( 1 ); }    // --callees: Target::pick (c.cpp:1) — right
};
```

**Cause.** Rule 2 narrowed a receiver only through `LocalBindKind::Type` (typed locals). A parameter's written type (`ParamType`) was captured, but only `collectFieldUseSites` read it. The call fell through to the name ladder, where the S6-C locality tie-break gives the caller's own class the scope credit. `--pin-census` shows the site as `mech=locality`, 2 candidates cut to 1.

## The change

**Rule 2, and CHA-lite through the same lookup (`Narrower::recvVarTypeName`), now reads `ParamType` records lexically.** For every name with a `ParamType` record, `buildScopedRecvDecls` (in `resolve.h`, beside the table type and its reader) lists all of that name's declarations in the definition: each `VarDecl` with its scope span. It then attaches every Type/ParamType record to the declaration whose `VarDecl` shares its record position (`Binding::startByte`, new; it sits in the padding after `kind`, so `sizeof` is unchanged and pinned). At the call site, the innermost covering declaration decides. An untyped declaration, a tie, or no covering declaration answers nothing. Names with no `ParamType` record keep the flat `varType` table and resolve byte-identically.

I measured two limits before choosing them:

- **The obvious fold into the flat per-function table is unsafe.** It fixed the defect but minted **three new precise wrong edges** on the gate fixture (arms 12–14):
  - a range-for variable's type reaching a later `auto` loop of the same name
  - a same-named field read after the loop
  - a parameter hidden by an untyped loop variable

  The flat table's tombstone also left arm 10 on the locality pin.
- **A qualified written type does not narrow.** A written type is recorded as its final segment, and class names carry no namespace. So `const std::map<K, V>& ref; ref.lower_bound( q )` narrowed to an unrelated in-repo `map`. That gave three such edges on a private C++/ObjC++ corpus (129,759 call sites).
  - The qualified text now rides the declaration's Type/ParamType `RawBind` (`importedName`, see `qualifiedNameText`), and the lexical lookup refuses it.
  - That removes all three wrong edges and gives up 11 correct narrows through qualified in-repo types. Those 11 sites keep `main`'s answer.
  - I tried and rejected an include-visibility guard first. Path-precise includes miss include-root spellings (`"LinearMath/btVector3.h"`), so it refused about 150 correct narrows on that corpus to stop the same three.

**`kParserVer` 96 → 97**, with the `quality.h` mirror. The record format is unchanged, but its content is not: a warm 96 blob holds `""` there and would narrow a qualified parameter type.

## Measured

Method: `--pin-census --no-cache`, the `main` binary (`f8e6087c`) against this branch.

| corpus | target changed | label-only (same target, now Rule 2) |
| --- | --- | --- |
| private C++/ObjC++, 129,759 call sites | **587** | 956 |
| this repo `src/` | **53** (all split → one Rule-2 pin) | 158 |

Private corpus breakdown:
- **373 splits narrow:** 300 via Rule 2, 73 via the CHA cone.
- **147 formerly declined calls gain an edge:** `bound=` 80,432 → 80,583, `declined=` 17,552 → 17,401.
- **66 pins or splits that didn't contain the parameter's type move to it.** 40 of these were `unique` pins to the one same-file method of the wrong class, e.g. `body->setGravity()` inside `addRigidBody( btRigidBody* body )`.
- **1 edge is lost:** a friend function ripwire scopes inside its class, which the parameter's type then names as the caller itself.

I sampled every category and read it against the source. Wall time is unchanged within noise: three cold runs of each binary gave 1.66–2.51 s at load ~12. The output is byte-identical between the dev build and the ASan build.

## Gates

`test/narrowcheck.sh` arms 7–18 run on a generated fixture. Red/green evidence:

| binary | result |
| --- | --- |
| `main` | **9 FAIL** — (7) (8) (10)×2 (11) (12) (13) (14) (15 census `mech=locality`) |
| naive flat-table fold | **5 FAIL** — (10)×2 (12) (13) (14) |
| lexical lookup without the qualifier guard | **(17) FAIL** |
| this branch | **23 PASS**, including under ASan |

Five gates had controls built on "a parameter has no binding". They now use an untyped `auto` receiver:
- `narrowcheck`: `control/param.cpp` becomes `control/untyped.cpp`.
- `chacheck`
- `chaconecheck` g5
- `localitycheck`: its call no longer reached the tie-break the gate exists to test.
- `resolverhonestycheck` F9: its `check_signal` row had gone vacuous on a single edge.

`fieldnarrowcheck` (h) moves from 7 to 6 because `shadowParam( Decoy& m_x )` now resolves to the parameter's type. Its (s1) arm now also asserts that `Pool::acquire` is not linked; that assertion is red on `main`.

Targeted runs on the final binary, 37 gates, ALL PASS:
- **resolver:** narrowcheck, chacheck, chaconecheck, localitycheck, chainguardcheck, fieldnarrowcheck, clsrecvcheck, fieldusescheck, fnptrcheck, shadowcheck, pincensuscheck, aritycheck, resolvecheck, resolverhonestycheck, narrowlangcheck, importnarrowcheck, externalvetocheck, qualifiedresolvecheck, objcfieldcheck, usescheck, callerscheck, selectorchaincheck, chainidcheck, cppqualcheck
- **cache / parser version:** cacheidentitycheck, qextractionkeycheck, cachesplitcheck, localscountcheck, moduleconstcheck, metalcheck, cudacheck
- **fixture consumers and registry:** compactlegendcheck, recallevalcheck, xmlwellformed, manifestcheck, gatecountcheck, limitstablecheck

Checks:
- **Full suite #1** (at the ack commit): `gates=632 pass=627 skip=2 fail=3`. All three failures came from this lane, and commit 5 fixes each:
  - **`showcasecapturecheck` (H):** lines added to `graph.h` shifted the zero-slack published seed. The builder moved to `resolve.h` instead of re-recording the release capture.
  - **`qschemetripcheck`:** the `kParserVer` re-pin case.
  - **`gateexitcheck` G2:** two one-line `&& ok || no` verdicts, now if/else.
- **Full suite #2** (at the final commit `eaadb9d0`): `gates=632 pass=630 skip=2 fail=0 wall=893.3s jobs=6 tree_writes=0` — one clean run, no timeouts. The two skips are the environmental `argvdiffcheck` and `editchecknotecheck`; `g1freshcheck` ran against a fresh `asan/` and passed.
- **Generators:** `docs/gatecount_build.py --check` and `docs/limits_build.py --check` both clean (618 gates; 212 caps).
- **quality-delta,** range form `$(git merge-base origin/main HEAD)..HEAD`: `gating="0"`.
  - It first reported `gating="3"`: a `ctorTypeOf|nodeTextOf` clone, and `Narrower`'s constructor at 6 parameters. I fixed both in code; `1508bd10` restructures and is output-identical by `cmp`.
  - The two rows left are the `Narrower` constructor's intended 4 → 5 contract change, acked through the binary (`--ack-only=api-surface`, exactly two `+ack` rows, one per symbol).
- **ASan/UBSan** (`-fno-sanitize-recover=all`): zero reports on the gate, this repo, a census run, and the private corpus.

## Out of scope, stated

- **An abstract parameter type narrows onto its namesakes** (disclosed in CHANGELOG by `186fcb84`).
  - Rule 2 resolves against definitions only. A parameter typed as an interface whose methods are pure-virtual declarations can't narrow to it, and lands on unrelated classes sharing the final name segment.
  - Measured on rocksdb @ `0e2801ac3` (`--pin-census --no-cache`, `main` `f8e6087c` against this branch): **79 sites (88 rows)** through an `Iterator*` parameter now split five ways over the nested `memtable/` `Iterator` classes, and none of the five is right.
  - Before: 25 were a unique override pin, 26 a split over overrides, 28 had no edge.
  - Every one is disclosed (`amb=`, `prov="split"`), but each is a wrong answer, not a missing one. This is the collision typed locals already have on `main`, extended to parameters.
- **Member-initializer lists:** `Decoy( Target& t ) : v( t.pick( 3 ) )` sits outside the parameter's body span and keeps `main`'s answer, which is the locality tie-break's wrong pin.
- **Untyped receivers:** `auto x = make(); x.m()` still reaches the locality tie-break.
- **Typed locals** still read the flat table, qualified-type collision included.
- **The reported arity observation** (`pick(int)` + `pick(int,int)` both kept for a 1-arg typed-local call) is not a receiver-path divergence. B2.2 drops a candidate only when `argCount > params` (a documented decision about default arguments on a separate prototype), and the bare-name ladder keeps both too.
- **Template member calls** (`other.pick<int>( 1 )`) inherit this once lane/template-call-edges lands.
- **`kParserVer` collision:** if that lane also bumps `kParserVer`, the second to land re-bumps (never reuse a number).

## Pins moved

- `kParserVer` 96 → 97 and `kIngestParserVerMirror` 96 → 97
- `test/qschemetrip.hash` re-pinned through `UPDATE_GOLDEN=1`: its manifest hashes the `kParserVer` declaration, and the new pin equals the `current=` hash the suite printed
- `fieldnarrowcheck` (h) `ambiguous=7` → `ambiguous=6`
- no legend, byte, `printf_parity`, capture or gate-count pins moved
  - `src/graph.h`'s diff is line-neutral (five in, five out) so the 0.6.1 capture's `--at=src/graph.h:3406` seed, the first body line of `rankGraphTeleport`, still resolves.

## Commits

1. `fix(resolve)` — the defect, gate arms 7–18, controls moved, `kParserVer` 97, CHANGELOG
2. merge of `origin/main` (`a55b118e`, #240); CHANGELOG keeps both Unreleased entries
3. `refactor(resolve)` — quality-delta's clone, parameter-count and complexity rows fixed in code; output byte-identical by `cmp`
4. `chore(quality)` — the `Narrower` constructor contract change, acked through the binary
5. `fix(lane)` — suite #1's three reds (details below)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect method resolution for calls made through typed parameters and other scoped receiver variables.
  * Improved handling of nested scopes, lambda parameters, range-based loop variables, and reference locals.
  * Qualified receiver types are no longer incorrectly narrowed to unrelated classes.
  * Updated cached analysis data to ensure results remain accurate after this change.

* **Tests**
  * Added coverage for receiver narrowing, lexical shadowing, ambiguity handling, and cache consistency.


joyful-ii-V-I:
Disposition of the review and CodeRabbit items. One push: **`eaadb9d0..186fcb84`** (`186fcb84`, docs only, no code). Every number below was re-measured on this lane's own binaries, not copied from the review.

| item | disposition | evidence |
| --- | --- | --- |
| "ten rows red on main" | **FIXED** in `186fcb84` (CHANGELOG) and in the PR body | `test/narrowcheck.sh` at this branch against the `main` binary (`f8e6087c`): rc=1 with **9** FAIL rows, (7) (8) (10)×2 (11) (12) (13) (14) (15). The fix commit `3d2d642b`'s message says "10 rows" but lists nine; it is pushed history, so it stays as is and this row corrects it. |
| Abstract-parameter floor (rocksdb `Iterator*`) | **DISCLOSED** in CHANGELOG and the PR body | rocksdb @ `0e2801ac3`, `--pin-census --no-cache`, `f8e6087c` against `eaadb9d0`: **79 sites (88 rows)** through an `Iterator*` parameter now split five ways over the nested `memtable/` `Iterator` classes, none right.<br>Before: 25 were a unique override pin, 26 a split over overrides, 28 had no edge.<br>The sampled receivers are all `Iterator* iter` parameters. The review's 32 counts `unique`→disjoint split only, from main `31e788ce` to the #254 tip. |
| Member-initializer floor | **DISCLOSED** in CHANGELOG and the PR body | On this branch alone, `Decoy( Target& t ) : v( t.pick( 3 ) )` keeps main's answer, the locality tie-break's pin to `Decoy::pick`. The honest split the review saw comes from #254's locality cap. |
| CodeRabbit pre-merge check "Docstring Coverage 56.25% < 80%" | **DECLINE** | CodeRabbit's default threshold, not a project rule: `.coderabbit.yaml` sets no docstring option. The house convention (CONTRIBUTING §3) is a `//` intent comment, and every new function carries one. |
| CodeRabbit review | nothing to answer | "No actionable comments"; 0 inline comments. |
| Merge `main` / CHANGELOG conflict | **DEFERRED to integration train 2**, per the CI coordinator | `main` is deliberately not merged on this branch. |
| `kParserVer` 97 / `test/qschemetrip.hash` | **UNTOUCHED**, per the CI coordinator | Both are resolved once at landing (collision with #244). |

joyful-ii-V-I:
This lands through integration train 2, #265, with parser version 97. Resolutions are on that branch, so nothing more is needed here. This PR will show as merged when #265 merges.

PR #249, opened 2026-09-16
fix: a cache blob, a file in the tree or an MCP preview could crash, hang or leak descriptors — six fixes, each red-first, plus static rules for their shapes

joyful-ii-V-I:
Every defect below crashed, hung or starved ripwire from input it does not control: a cache blob, a file committed to the tree, a filesystem timestamp, or an MCP request. Each one was reproduced before it was fixed and has a gate that fails on the base. The commits are one per defect, plus a CHANGELOG commit.

| # | trigger | before | after | gate |
|---|---|---|---|---|
| 1 | checksum-valid qsnap blob, vector count `0xFFFFFFFF` | `reserve(n)` asks for 32 GiB → `std::bad_alloc`, SIGABRT (exit 134) on every `--quality-delta`/`--edit-check` until evicted (Ubuntu) | count bounded by bytes left; blob disclosed corrupt and recomputed | `cachefuzzcheck` Part 2, three vector rows, bounded |
| 2 | checksum-valid qchurn blob, commit or path count `0xFFFFFFFF` | 96–128 GiB reserve → SIGABRT on `--for`/`--metrics`/`--exemplar` (Ubuntu) | same bound; silent miss, walk rewrites the blob | `cachefuzzcheck` Part 3 |
| 3 | a file that reads short (truncated mid-read) | `( got == want ) && fclose(fp)` leaks the FILE; 300/600 short-read streams never closed; under `ulimit -n 200` all 20 ordinary files vanish from `--grep` | `rw::OwnedFile` closes on every path | `crashsweepcheck` B1 (interposed short-read shim) |
| 4 | FIFO, directory, `/dev/zero` or `/dev/urandom` link at `.ripwire_config` / `.ripwire_quality_acks` | `--quality-delta` hangs (FIFO, device links); a directory at `.ripwire_config` is the Linux `LLONG_MAX` allocation shape | `readRegularFile`: `O_NONBLOCK` open, `fstat`, anything irregular reads as absent, disclosed on stderr | `crashsweepcheck` B2, 8 shapes |
| 5 | MCP `edit_check` with `new_body` right after a HEAD move | preview ingests ran outside the ingest lock, racing the prefetch worker; TSan: data race + server abort (exit 134) | the preview ingest takes the process-wide ingest lock | `qsnapprefetchcheck` (f); red needs `RIPWIRE_BIN=tsan/ripwire` |
| 6 | a file dated after 2262 (ext4/XFS/tmpfs/tar) | `tv_sec * 1e9` signed overflow: UB in release, abort in the sanitizer build (Linux tmpfs) | `rw::saturatingNanoseconds` | `crashsweepcheck` B3; the ASan CI job now runs the gate |

**Static rules** (the owner's question: could these have been caught?). `crashsweepcheck` also runs ripwire's own `--match` over `src/`, and each rule is proven live on a probe tree with one violation and one compliant twin:
- **S1:** an allocation sized by a decoded count must be bounded earlier in the function. On the base it finds the three quality.h counts.
- **S2:** every raw `fopen`/`open`/`fdopen`/`opendir`/`open_memstream`/`popen` site is registered with the fact that makes it safe, and a new site points at `rw::OwnedFile`. A close hidden behind `&&`/`||`/`?:` is refused. On the base it finds `readFile`.
- **S3:** every `std::thread` body is `noexcept` or a single try block. On the base it finds eight bodies. They are now `noexcept`; nothing changes at runtime, because a throw leaving a thread was already `std::terminate`.

The gate header states what each rule catches and what it misses.

**Runtime checks.** None of these would have tripped a debug assertion. An input-sized count is not an invariant, and VERIFY on external data is forbidden. A short read is legal, and a FIFO is a legal file. The sanitizers catch three when a test reaches the path: ASan with an allocation cap (1, 2), TSan (5) and UBSan in the G1 build (6). For (3), clang-analyzer's `unix.Stream` checker had already caught this exact shape in docparse.h's twin.

**Verification.** Every arm was run red on the base and green on this branch, on the plain and ASan builds; (f) was also run on TSan. 42 targeted gates passed locally at the PR head: the three owning gates (also on ASan), plus `gateexitcheck`, `manifestcheck`, `gatecountcheck`, `binoverridecheck`, `infraportcheck`, `limitstablecheck`, `ripwirepubliccheck`, `versioncheck` and the neighbours of every touched reader (`sidecarsymlinkcheck`, `registermacrocheck`, `qackconcurrencycheck`, `staleackcheck`, `qualitycheck`, `qddialscheck`, `editpreviewcheck`, `editcheckcheck`, `mcpeditcheck`, `statgatecheck`, `racymtimecheck`, `freshnesscheck`, `mcpincrementalcheck`, `mcpstalecheck`, `mcpwatchercheck`, `qsnapcachecheck`, `qchurncheck`, `qchurnmemocheck`, `textdocscheck`, `githardencheck`, `namedfileinputcheck`, `tornreadcheck`, `savecachecheck`, `cachesplitcheck`, `grepcheck`, `greptiercheck`, `docdriftcheck`, `crossrefcheck`, `fieldidcheck`, `childwalkscalecheck`, `lintrulescheck`). The full suite is left to CI. `--quality-delta=<merge-base>..HEAD` reports `gating="0"`; its five rows are all minor.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented malformed cache data from causing excessive memory allocation or process termination.
  - Improved handling of incomplete file reads to prevent resource leaks.
  - Fixed hangs and crashes when configuration or quality files are FIFOs, directories, devices, or invalid links.
  - Prevented timestamp overflow for dates beyond April 2262.
  - Improved reliability when preview edits occur during background indexing.

- **Tests**
  - Added regression coverage for corrupted caches, unusual file types, timestamp boundaries, and concurrent preview operations.

- **Documentation**
  - Updated documented test-gate totals from 618 to 619.


joyful-ii-V-I:
Disposition: one push, eight new commits on top of `1502e7ff`, covering the CodeRabbit review, the two CI failures and the merge with main.

**CodeRabbit findings**
- `0e759ed6` **fix(quality): the acks ledger is read one line at a time again, through a descriptor-checked stream.** *(docparse.h: bound fixed-name reads.)* Valid for the ledger: the shape fix had turned its streaming read into a whole-file read. Rather than adding a size cap, the O_NONBLOCK + `fstat` check now returns `pathguard::NoFollowRead`, the house line reader (it owns the FILE and reads with POSIX `getline`), and `readAckRecords` streams through it. The house rule for sidecars is streaming, not a cap that would refuse a legitimately large ledger. `.ripwire_config` keeps its whole-text parse, which predates this PR.
- `ab0afb73` **test(crashsweep,cachefuzz): S1 stops excusing a size that merely contains sizeof, and no row passes on a leg that did not run.**
  - *(S1 excused `n * sizeof( T )`.)* Valid. `sizeof` and `.size()` terms are now stripped, and the remaining identifiers still need a bound. The MISSES note names the nested-parentheses residue.
  - *(B2 judged shapes against an empty truth.)* Valid. The shape loop now runs only against a non-empty truth.
  - *(The qchurn row passed without a bounded leg.)* Taken in part. The unbounded row stays, because it proves the blob is refused and rewritten and it goes red if the reader accepts the blob. It is now named `correctness:<row>`, and a missing sanitizer leg is a named SKIP.

**CI failures on `1502e7ff`**
- `67a4e31e` **qschemetripcheck** re-pinned to `37dac79a…`, with a RE-PIN LOG entry. `deserializeSnapshot` only gained a refusal: a count no blob of that length can hold. Layout and meaning are unchanged, so **`kQSnapCacheScheme` stays 12** (no bump). `qextractionkeycheck` is green.
- `3d5da74a` **cppqualcheck** `--uses=readWholeFile` 24 → 22. The two sites removed are this lane's: `readRegisterMacrosConfig` and `ackNothingToAccept` now read through `readRegularFile`. A line-level diff of every call against main shows exactly those two.

**Latent offset defect (peer lane's producer field)**
- `9aca6fda` cachefuzzcheck now verifies the layout before aiming any row at an offset. For qsnap, the u64 at `QSNAP_SHA_OFF` must be `fnv1a64(HEAD sha)`, and `QSNAP_COUNTS_OFF = SHA_OFF + 8`. For qchurn, walking the good blob from the count offset must land on the trailer. A mismatch FAILs loudly.
- `62b60e42` The first red proof (shifting the qsnap offset to 8 and the qchurn one to 12) showed the checks fire, but also that bash 3.2's `set -u` then aborted the gate on the empty mutant table. The gate now reports both layout FAILs and runs every other Part.

**Merges**
- `e0f648f9` main 731d23be (#241): cachefuzzcheck keeps #241's Parts 3 and 4, and this lane's qchurn part becomes Part 5.
- `b6e6b2b6` main 31e788ce (#245, #246): the regression.sh loop is the union of the two lanes' additions (`diagnoticecheck`, `crashsweepcheck`), and `gatecount_build.py` → **620**. This lane's CHANGELOG entry now sits at the top of [Unreleased].

**Verified at the new head.** The plain and ASan builds are from the merged tree. Every gate passed, rc 0:
- the requested set: `qschemetripcheck`, `qextractionkeycheck`, `cppqualcheck` (plain and ASan), `cachefuzzcheck` (plain and ASan, including the bounded ASan rows), `gateexitcheck`, `manifestcheck`, `gatecountcheck`, `limits_build.py --check`;
- this lane's own arms: `crashsweepcheck` (plain and ASan) and `qsnapprefetchcheck` (plain and ASan);
- the neighbours: `sidecarsymlinkcheck`, `staleackcheck`, `qackconcurrencycheck`, `qualitycheck`, `registermacrocheck`, `qddialscheck`, `editpreviewcheck`, `statgatecheck`, `mcpincrementalcheck`, `diagnoticecheck`, `binoverridecheck`.

`--quality-delta=<merge-base>..HEAD` reports `gating="0"`.

joyful-ii-V-I:
Disposition for the one red job on `62b60e42`, `asan (ubuntu-24.04)` → crashsweepcheck `SCANFAIL rc=-6`. One push, two new commits.

**Classification: (b), a real leak in vendored tree-sitter, reachable from any structural query.** It is not a gate-design problem.

- `f08ec8c1` **fix(astquery): a structural query naming a field some grammar lacks leaked its capture quantifiers.** In `ts_query__parse_pattern` (query.c), the field-prefixed branch parses the sub-pattern into `field_capture_quantifiers`, then returns `TSQueryErrorField` without deleting it when the grammar has no such field. `--match` and `--lint-rules` compile every query against every linked grammar. So S2's `right: (unary_expression argument: (call_expression … @c))` leaks 16 bytes for each grammar that has `unary_expression` but no `argument` field, and LSan (`detect_leaks=1` + `abort_on_error=1`) aborts at exit. macOS never showed it because Apple's ASan has no LSan.
  - **Reproduced** with `leaks --atExit` on macOS: the pre-patch build reports `1 leak for 16 total leaked bytes` for that query. All 13 other queries crashsweepcheck issues report 0, before and after, so this was the only leak on the gate's path. Patched: 0 for all 14.
  - **Fix:** the missing delete, recorded as `third_party/patches/tree_sitter/001-field-error-quantifier-leak.patch` with its `RIPWIRE_VENDOR_PATCH` marker. There is no lsan suppression: `lsan_suppressions.txt` deliberately keeps `ts_query_*` visible.
  - **Gate:** `vendorpatchcheck` A–C carry the patch. New arm **L** runs the leaking query under LeakSanitizer on a Linux sanitizer build, or under `leaks(1)` on a plain macOS build with a leak-free control, and SKIPs by name elsewhere. L is red on the pre-patch build (1 leak) and green here. The Linux red→green is this job's own crashsweepcheck step.
- `4599057b` **test(crashsweep): a scan the sanitizer aborts now shows the sanitizer's report instead of a bare rc=-6.** The job's `log_path=/tmp/asanlog` hid the report. The scan's runs now append `log_path=stderr`, and a failed scan prints the report tail under the FAIL row.

**Linux repro.** The `rw-linux-x64` container could not host a fresh ASan build: its overlay filesystem is 100% full, and I did not delete anything in another session's container. Its Aug-10 ASan binary predates the relevant code and does not reproduce. The Linux proof is therefore this push's ASan job.

**Verified locally.** Plain: `vendorpatchcheck`, `crashsweepcheck`, `matchgrammarcheck`, `matchcapturecheck`, `lintrulescheck`, `gateexitcheck`, `manifestcheck`. ASan: `vendorpatchcheck` (L SKIPs on macOS), `crashsweepcheck`, `matchgrammarcheck`. All rc 0.

joyful-ii-V-I:
Disposition for the independent review of `4599057b` (F1, F2, F4, F6). One push; main is merged once (the #250 exception) and there are four new commits.

- `f2786db6` **merge(main): take #250.** This resolves F2.
  - `regression.sh`: the gate loop is the union (#250's `tempfilesymlinkcheck`, this lane's `crashsweepcheck`), and `gatecount_build.py` → **621**.
  - `cppqualcheck` `--uses=readWholeFile` → **19**, re-derived on the merged tree. There are 20 grep lines, which is 19 call sites plus the definition, and `--uses` answers 19. #250's three githarden sites and this lane's two fixed-name readers are disjoint.
  - CHANGELOG merged clean; this lane's entry is still at the top of [Unreleased].
- `8850a60c` **test(crashsweep): S2 recognises #250's owner types, re-registers its writers, and fails on a stale allow row.** This resolves F1.
  - A call passed straight to an owner constructor needs no registry row: `OwnedFd fd( ::open(…) )`, `OwnedFile fp( ::fdopen(…) )`, `return OwnedFile( std::fopen(…) )`. That covers pathguard's `readRegularFileNoFollow` open, `readWholeBeneathNoFollow`'s first open and its file openat, and `openOwnedFile`. The probe tree gains an `OwnedFd( ::open )` twin that must not fire.
  - Registered, each with its reason: the three `fdopen` calls adopting `ExclTempFile`'s released descriptor (`saveOracleCache`, `saveCache`, `publishDocBridgeBlob`), `openExclNoFollow` (transferred into `ExclTempFile`), `randomTempSuffix` (close before the only exit), and `readWholeBeneathNoFollow`'s directory `openat` (adopted by `cur.reset( next )`).
  - The four stale rows are deleted, and so is the dead `saveCache` short-circuit allow row.
  - An allow row in S1, S2 or S3 that excuses nothing now fails the gate. Mutation-checked: re-adding the `saveCache` row turns S2 red.
- `33b2c357` **fix(docparse): a refused fixed-name file is disclosed once per run, not once per read.** This is the F4 NIT (3 identical lines). crashsweepcheck B2 now asserts exactly one line.
- `5daec455` **test(qsnapprefetch): the no-TSan-warning row SKIPs on a non-TSan binary.**
  - **Known gap for fix 5 (edit_check preview race): there is no CI guard.** Arm (f)'s behavioural rows pass on the unfixed plain binary as well, and CI has no TSan job. The red-first evidence is a local TSan run only: data race plus server abort on the base TSan build, clean on the fixed one. This PR claims no CI red for it.
- `43261a16` **docs(patches): the `tree_sitter/001-field-error-quantifier-leak.patch` row.** This resolves F6. The row records what the patch fixes, how it was measured, the gate (vendorpatchcheck L), and that it is an **upstream candidate**: tree-sitter master still returns without the delete at `query.c:2877-2879`.

**Verified at `43261a16`, plain build stamped at that commit, all rc 0:** `crashsweepcheck`, `cppqualcheck`, `gatecountcheck`, `manifestcheck`, `gateexitcheck`, `cachefuzzcheck`, `vendorpatchcheck`, `qsnapprefetchcheck`, `versioncheck`, `tempfilesymlinkcheck`, `sidecarsymlinkcheck`, and `limits_build.py --check`. No local ASan (machine rule).

**Left for the train (F8):** the qsnap-producer-identity lane moves the qsnap sha to offset 24, so cachefuzzcheck's `QSNAP_SHA_OFF` must change with it; the layout pre-check fails loudly until it does. The qschemetrip re-pin should also be done once, on the integrated tree.

PR #251, opened 2026-09-16
fix(regex): a user's regex aborted --graph-query and --arch, --match kept rows no predicate decided, and file() matched the checkout's directory name

joyful-ii-V-I:
## The defects

Only `--regex` screened a user's pattern before handing it to `std::regex`. Three other entry points reached the same engine unscreened, and none of them caught anything at match time.

Repro (Apple libc++, dev build of origin/main `f8e6087c`). The fixture has a directory named with 44 `a` and an include edge `zz/b.c -> aaaa…/aaaa….h`:

```
ripwire rxfix --no-cache --graph-query='file(all,"(a+)+z")'
  rc=134  libc++abi: terminating due to uncaught exception of type std::__1::regex_error: The complexity of an attempted match …
ripwire rxfix --no-cache --arch=rules.txt          # deny path zz/.* -> (a+)+z
  rc=134  same abort (the FROM form `deny path (a+)+z -> .*` aborts too)
ripwire rxfix --no-cache --match='(function_declarator declarator: (identifier) @fn (#match? @fn "(a+)+z"))'
  rc=0    the throw was swallowed; rows the predicate never decided are KEPT
ripwire src --match='… (#match? @n "("))'
  rc=0    a malformed pattern filters nothing: every row printed as the query's answer
ripwire rxfix --regex='(a|a)+z'                    # passes the structural screen
  rc=0    the engine gave up mid-file; the file was skipped under an alert Release deletes, and hits= printed as a count
```

libstdc++ has no match budget, so on Linux each of these backtracks without end.

`file()` also matched the path with the checkout's own directories in front of it. `file(all,"alpha")` selected every symbol in a clone named `repo_alpha` and none in `repo_beta`. Under an absolute root, `file(all,"^src/")` selected nothing.

## The seam

`src/regexguard.h` owns every user-authored regular expression: the structural screen (non-portable escapes and the nested-quantifier family, moved verbatim from `src/search.h`), the compile and the match.

- **One exception boundary.** It is the only place in `src/` that catches `std::regex_error` and `std::bad_alloc`. It catches them by type and turns them into values: `RegexCompile{ regex, refusal, isScreened }` and `RegexVerdict{ Miss, Hit, Exhausted }`. There is no `catch(...)`. The public API is `noexcept`, and `static_assert`s pin that. No call site holds a `try`.
- **Every entry point goes through it:**
  - `--regex`: `regexCompileError` is now a one-line wrapper, and its refusal text is byte-identical.
  - `--graph-query file()`.
  - `--arch` path rules: FROM is compiled and TO is screened when the rules file is parsed.
  - `#match?` / `#not-match?` in `--match` and `--lint-rules`, collected through two opt-in `AstQueryGroup` outputs.
- **A refused pattern** gets a refusal by name at exit 1, worded the way `--regex` already refuses.
- **A match the engine abandons** (`Exhausted`) also gets a refusal by name at exit 1. It is never read as "no match".
- **The built-in lint packs** keep the old keep-the-row fallback for their constant patterns.
- **`file()` matches the root-relative path its own `p=` prints.** This is the rule `--arch` adopted for its rules. Each file is decided once instead of once per symbol.

## Gates, red then green

`test/regexguardcheck.sh` (new):

- **(a)** A catastrophic pattern is refused by name on `--graph-query file()`, on `--arch` TO and FROM, and on `#match?` in `--match` and in `--lint-rules`. `--regex` is included as a regression check. Each entry point has a positive control.
- **(b1)** The fault switch `RIPWIRE_FAULT_REGEX_MATCH=1` (non-NDEBUG only, compiled out of Release) makes every guarded match throw. Each entry point must refuse, whatever the standard library. The build flavour is established with the sibling `RIPWIRE_FAULT_CHARGE_BUFFER` probe.
- **(b2)** `(a|a)+z` runs against libc++'s real engine.
- **(c)** Static check: no `std::regex` construction, match, iterator, result type or `<regex>` include outside the owner, with comments and strings blanked. The allowlist has two rows (`redact.h` and `skillscan.h`, both constant rule tables), each with a reason. Planted-file controls cover both directions, and a stale row fails.
- **(d)** Two clones at different directory names, run with absolute, relative and `./x/` roots, give the same counts. `^src/` works under an absolute root, and every row's `p=` is the path that was matched.

| binary | result |
|---|---|
| origin/main `f8e6087c` | 35 FAIL, 4 of them rc 134 |
| this branch | 63 PASS, ALL PASS (also ALL PASS on the ASan/UBSan binary) |

All five registrations are done: the `regression.sh` loop, the gate count (620 on the merged tree), and a shard weight of 14.8 s. The gate binds `RIPWIRE_BIN`, so it needs no binoverride EXEMPT row, and its header has no absolute paths.

Targeted gates, each run as its own script with its exit code read, on the merged tree `e30fcc49` (dev build): **94 gates, all rc 0, 3,097 PASS rows**.

- **Regex and grep:** regexguardcheck 63, regexbombcheck 65, regexcheck, regexrefusecheck, grepanchorcheck, and the 13 grep gates. grepfastcheck (6) is byte-identical to `RIPWIRE_BASE` on all 13 vectors.
- **Query and arch:** querycheck, graphqueryrefusecheck, layerquerycheck, archcheck, archmetricscheck, baselinecheck, rootrelcheck, rootrelemitcheck.
- **Match and lint:** astqueryregexcheck, lintcheck, lintrulescheck, matchcapturecheck, sarifcheck, patterncheck.
- **Meta and pins:** gateexitcheck, manifestcheck, gatecountcheck, binoverridecheck, printffmtparitycheck, docscommandscheck, limitstablecheck, compactlegendcheck, legendcoveragecheck, ripwirepubliccheck, infraportcheck, qualitycheck, staleackcheck, ackonlycheck, versioncheck, diagnoticecheck.
- **Generators:** `limits_build --check`, `gatecount_build --check` and `docs_commands_build --check` all pass.

Under ASan/UBSan (`-DRIPWIRE_ASAN=ON`, LSan suppressions), checked on `fb7a0533`, before the last merge, which touched no regex code:
- 18 direct runs across every touched verb, including the fault-switch paths, printed no sanitizer line.
- regexguardcheck, astqueryregexcheck, regexbombcheck, querycheck, archcheck, lintrulescheck and grepanchorcheck were ALL PASS against the asan binary.

## Zero cost

**Byte-identical output.** The origin/main binary and this one were compared on one checkout, with stdout, stderr and exit code all diffed. **32 of 32 comparisons are identical.** The cases cover:
- `--grep`: literal, and JSON.
- `--regex`: word, anchored, `--no-prefilter`, alternation, context, compact, unindexed-aux, zero-hit, and the three existing refusals (bomb, malformed, non-portable escape).
- `--graph-query`: `file()` via a relative root, `and()` pushdown, `cx`+`file`, `kind`.
- `--arch`: the parity rules, the sibling path-rules, and path-rules over the whole repo.
- `--match`: anchored, `#not-match?`, and a capture-typed argument.
- `--lint` on `src` and on the repo root, `--lint-rules`, `--lint-rules --sarif`, and the default map.

A three-request `--mcp` grep session is also byte-identical.

The only differences are the fixes, listed separately:
- `file(all,"regex")` under an absolute root whose path contains "regex": 9 → 0.
- `file(all,"^src/chain")` under an absolute root: 0 → 4.
- The malformed `file()` refusal is now named.
- A malformed `#match?` refuses instead of printing 6,945 B of unfiltered rows.

**No compile was added.** A pattern is compiled once per query, once per rule and once per grep worker, as before; the per-edge compile of an `--arch` TO pattern is unchanged. `file()` now runs one `regex_search` per distinct file instead of one per symbol. The screen is a single linear pass per compile. The `try` around a match costs nothing on the non-throwing path, and the fault switch is `constexpr false` under NDEBUG.

**Not measured: release `__TEXT,__text` and objdump.** The Release scratch builds were started twice and cancelled both times: the shared machine held a build queue from other sessions (load 20–40), and the CI-handoff protocol arrived first. The claim above rests on byte-identical output and on unchanged compile placement. No codegen figure is quoted.

## Pins moved

- `test/astqueryregexcheck.sh` C4 and its golden's `match-malformed` section. A malformed `#match?` in a user's `--match` is now refused at exit 1; it used to print every row unfiltered.
- `test/printf_parity.manifest` `help_all`: the `--help` `file()` clause. `docs/COMMANDS.md` is regenerated, and its caveat bullets are unchanged.
- `.ripwire_quality_acks`: one row healed and one row added. Both are api-surface contract changes: `passesPredicates` gains the undecided counter, and `query::Eval` gains the single-root prefix. `--quality-delta=<merge-base>..HEAD` reports `gating="0"`; its two new-symbol rows are the screen, moved verbatim.
- Gate count: 620 after the union with `diagnoticecheck` from main.

## Summary by CodeRabbit

* **New Features**
  * Added consistent protection for user-provided regular expressions across search, graph queries, architecture checks, matching, linting, and regex commands.
  * Unsafe, unsupported, or unfinished regex evaluations are now clearly reported and return exit status 1 instead of producing potentially misleading results.
  * `file()` filters now match root-relative paths consistently, including `^src/` patterns.

* **Documentation**
  * Updated command help, guides, changelog, and presentation materials to describe regex handling and the expanded 620-script gate suite.

* **Tests**
  * Added comprehensive coverage for regex refusal, abandoned matches, and root-relative file filtering.


joyful-ii-V-I:
## Disposition — CodeRabbit review round, the coordinator's items, and the Release performance proof

**New commits on `lane/regexguard`, one push:**

| sha | subject |
|---|---|
| `18cce2ae` | fix(regex): an `--arch` TO template that could never compile was stored; one that its own substitution broke was silently inert; and a 20,000-byte `--regex` died in a 512 KiB grep worker |
| `6f511b52` | fix(skillscan): a fenced `curl curl …` line hung the skill scanner, and an engine throw would have ended wrap's `noexcept` scan |
| `62545d5c` | docs(changelog): the regex entry states what the round found and fixed, the misleading sentence is reworded, and the entry moves to its slot after #245's entry |
| `ed234e19` | docs(readme): the cap inventory sentence said 213; the regex stack bounds made it 215 |
| `68dea35f` | refactor(skillscan,arch): the linear net-exfil scan, the frontmatter table and the `--arch` undecided-rule refusal each read as one small unit |
| `0ed796b9` | docs(changelog): the regex entry's performance sentence quotes the final commit's Release measurement (head) |

main was not merged, per the protocol. Against the merge base `31e788ce`, the CHANGELOG diff has 0 deletions.

### CodeRabbit findings

1. **CHANGELOG sentence ("an alert Release deletes and print `hits=`") — FIXED in `62545d5c`.** It now says the scan skipped the rest of that file behind a `DEGRADED_PATH_ALERT`, which a Release build compiles out, and still printed `hits=` as a complete count.
2. **Reject malformed TO templates while parsing (`src/arch.h`) — FIXED in `18cce2ae`.** `toTemplateRefusal` compiles a copy with a placeholder atom in each `\1..\9`: first `x`, then `9` for an interval position such as `a{2,\1}`. A template that compiles with neither placeholder rejects the rules file, with the line named. The stored template is unchanged, so `substituteBackrefs` still gets the original. Arm (e0) passes `deny path zz/(\w+)/.* -> (\1`: exit 1, `TO path-regex '(\1' refused … rules file rejected`.
3. **Reject overlapping alternatives before matching (`src/regexguard.h`) — DECLINED in this PR.** A structural screen that rejects `(a|a)+z` either rejects valid linear patterns or misses real bombs:
   - `(ab|ac)+` shares a first byte but backtracks linearly.
   - Semantic overlap needs a language-intersection test, not a token scan.

   The honest fix is an engine with a hard budget. That changes the engine, and it needs an owner decision. The measured libstdc++ residual below belongs to the same decision, so I've listed the options there rather than guessing one here. What this PR does guarantee: on libc++ an abandoned match is refused by name on every entry point (arm b2), and arm (b1)'s fault switch proves that disclosure path on every standard library.
4. **Enforce the whole exhaustion contract (`test/regexguardcheck.sh`) — FIXED in `18cce2ae`.** Arms (b1) and (b2) now share one helper, `abandonedContract`. It checks exit 1 (never a signal), `abandoned the match` on stderr, the pattern named on stderr, and no answer element on stdout, each as its own row. `--lint-rules` names the rule ids whose queries carry a `#match?` (the undecided counter is kept per group, so a rule id is the most specific thing it can name). The gate asserts `rx-ok`. The one-line `&& ok ||` verdicts in those helpers are now `if` blocks.

### Coordinator items

- **deckcheck — FIXED in `62545d5c`.** The CHANGELOG no longer names a flag that `--help` doesn't list. deckcheck passes locally.
- **An `--arch` TO pattern that fails only after substitution — FIXED in `18cce2ae`.**
  - `pathRuleForbids` now reports `isRefused` together with the substituted text.
  - The verb refuses at exit 1 and names the rule, the edge and the text, e.g. `a{2,\1}` becoming `a{2,1}`, which is invalid on every standard library.
  - Arm (e) shows the refusal. Its control, `a{1,\1}`, substitutes to the valid `a{1,1}` and is still judged.
- **CHANGELOG slot — DONE in `62545d5c`.** The entry now sits directly after #245's entry and before "Intel macOS binaries end with 0.6.1".
- **Addendum (a), regex stack depth — FIXED for compile in `18cce2ae`; match on libstdc++ is disclosed, not fixed.** Measured with a standalone probe that compiles and destroys one pattern per run and binary-searches for the first signal death:

  | smallest crash | libc++ std::thread (512 KiB) | libstdc++ 13, 8 MiB | libstdc++ 13, 512 KiB |
  |---|---|---|---|
  | nested groups | 3,392 deep | 15,616 deep | 960 deep |
  | literal pattern | 16,896 B | 58,368 B | 3,648 B |
  | `(a\|b)*c` **match**, subject length | no crash (slow at 32 KB) | **13,312 B** | 816 B |
  | `a*b` / `.*x` **match** | no crash (slow at 65 KB) | **26,624 B** | 1,632 B |

  - **Reproduced:** `--regex=<20,000 bytes of a>` passed the main thread's probe compile and then died with SIGBUS (rc 138) in a grep worker.
  - **The bound:** the screen now refuses a pattern over **2,048 bytes** or nested **64 groups** deep, by name. Both limits sit below every measured compile crash; the tightest margin is libstdc++ on a 512 KiB stack, 1.78× on bytes and 15× on depth. A pattern exactly at each limit still compiles, and arm (g) checks both sides. Both limits are classified BOUNDARY in `docs/limits_classes.tsv`, and LIMITS.md and TUNING.md are regenerated.
  - **Why not compile on the calling thread:** it would not reach the worker-side destruction, the astquery compile threads, or the prefetch thread.
  - **Residual, disclosed in the CHANGELOG and not claimed fixed:** libstdc++'s DFS matcher recurses once per consumed character. On Linux with the default 8 MiB stack, `--regex='a*b'` crashes on a 26,624-byte matching line, and `#match?` on a large captured node or a skill-file line with a long whitespace run can do the same. A pattern bound cannot reach this. **Owner decision needed, three options:**
    1. Refuse by name when a subject exceeds about 8 KB. Portable, but `--regex` would refuse on any tree with one long line.
    2. Use libstdc++'s `regex_constants::__polynomial` BFS executor. Its recursion is bounded by pattern size and it is polynomial time, so it would also bound finding 3 on Linux. But it rejects back-references, and it picks the leftmost-longest match rather than ECMAScript's leftmost-first. Grep hit sets and `--arch` captures could then differ between libc++ and libstdc++.
    3. A budgeted linear engine, either vendored or hand-rolled.
- **Addendum (b), skill scanner — FIXED in `6f511b52`.** The constant patterns turned out to be worse than able to throw: one could hang.
  - **Hang:** `EXFILTRATE:net-exfil`, `(\b(curl|wget|nc)\b.*(\$VAR|base64))|(…)`, is quadratic in the line. Measured on the release binary: a 20,000-byte fenced `curl curl …` line took 5.9 s, and a 200,000-byte line was still running at 60 s.
  - **Termination risk:** `hasTransmitVerb` is `noexcept`, and `wrapScanSkillDir` (`noexcept`) scans through these regexes, so an engine throw would have terminated the process.
  - **Now:** every pattern compiles and matches through regexguard.h. An undecided match fails closed, as a CRITICAL `SCAN-INCOMPLETE:regex-abandoned` finding with exit 2. `net-exfil` is decided by `hasNetExfilShape`, a linear scan derived exactly from the regex. The static arm's skillscan.h allowlist row is removed.
  - **Arms:**
    - (f1) The linear decision agrees with an independent Python oracle (with `.` written as `[^\r\n]`) on 1,200 generated fenced lines, 184 of them positive. The origin/main binary, still running the regex, also passes (f1), which is what validates the oracle.
    - (f2) The 200,000-byte line scans within 20 s, and a `$SECRET` placed at its end is still caught.
    - (f3) The fault switch produces the CRITICAL finding.

### Release performance proof

origin/main `31e788ce` vs `68dea35f`, both `-DCMAKE_BUILD_TYPE=Release` in scratch build directories, niced. Numbers only, no budget gate.

- **`__TEXT,__text`:** 8,684,280 → 8,704,196 B (+19,916 B, +0.23%). The growth is the refusal texts, the skill-scanner scan and the template validation. The first push `e30fcc49` measured 8,695,396.
- **`grepScanText`:** an out-of-line symbol in both builds. It went from **443 instructions / 1,772 B to 437 / 1,748 B**.
  - It makes the same call set: 3× `basic_regex::__search`, the same length_error and bad_array_new_length paths, and the same PLT stubs.
  - The lane adds one cold `__clang_call_terminate` landing pad, from `forEachMatch` being `noexcept`.
  - The `try` moving from per-file to per-line adds no instructions on the non-throwing path.
- **Instructions retired** (`/usr/bin/time -l`, whole process, median of 5, interleaved main/lane, warm cache per binary). The machine's load average was 16–28 throughout, so instructions retired is the load-independent figure:

  | corpus | case | main | lane | Δ |
  |---|---|---|---|---|
  | llvm-project (8,837 C/C++ files) | `--regex=getOperand\w*\(` | 216.55 G | 211.28 G | −2.43% |
  | | `--regex=^#include` | 204.28 G | 199.24 G | −2.47% |
  | | `--regex=\bconst\b --no-prefilter` | 216.01 G | 208.79 G | −3.34% |
  | | `--grep=getOperand` (control) | 25.50 G | 25.75 G | +0.99% |
  | | map (control) | 23.20 G | 23.28 G | +0.33% |
  | ripwire | `--regex=getOperand\w*\(` | 12.18 G | 12.22 G | +0.33% |
  | | `--regex=^#include` | 21.61 G | 21.52 G | −0.43% |
  | | `--regex=\bconst\b --no-prefilter` | 39.76 G | 39.94 G | +0.43% |
  | | `--grep=getOperand` (control) | 1.83 G | 1.85 G | +1.63% |
  | | map (control) | 1.10 G | 1.12 G | +1.20% |

  - **Wall/CPU**, median of 5 on `e30fcc49` at the same load: llvm `--regex` CPU −4% to −14%; ripwire `--regex` wall between −7% and +15%. The `--grep` literal control moved ±4–8% on an unchanged code path, so the wall numbers are noise at this load.
  - **Verdict: no measurable cost** from the per-line `try`, so no restructure. Every timed run's stdout was byte-identical between the two binaries.
- **Byte-identical output:**
  - The CLI is **45 of 45** identical against the `31e788ce` dev binary. The comparisons cover `--grep`/`--regex` (every mode and the three existing refusals), `--graph-query`, `--arch` (including a backreference TO rule), `--match`, `--lint`, `--lint-rules` including SARIF, `--scan-skills` over `test/skillfix` and `skills/`, `--scan-skill` over all 10 fixtures, and the map.
  - A three-request `--mcp` grep session (2,624 B) is also identical.
  - The only differences are the 4 intended fixes.

### Gates on `0ed796b9`, the head (dev build)

- `regexguardcheck`: **94 PASS**. Red on origin/main `31e788ce`: **55 FAIL**, seven of them a signal death (rc 134/138).
- 98 targeted gates were run, each as its own script with its exit code read, all rc 0 (3,262 PASS rows); they are listed in the lane report. Two gates first went red on generated counts, and both are fixed in this push: `limitstablecheck`/`capsweepcheck`/`readmedriftcheck` (regenerated) and `deckcheck`.
- `--quality-delta=31e788ce..HEAD`: **gating="0"**. The two gating rows the round created were fixed, not acked, in `68dea35f`: a clone of `accessShapeSpecs` and one nesting level in `runArchViews`, along with a clone of `namesplit::isIdentChar` that the fix itself had introduced.
- Per the machine rule, no local ASan rebuild this round; the CI asan legs run these gates. The earlier ASan pass on `fb7a0533` was clean.

joyful-ii-V-I:
## Disposition — independent review of #251 @ 0ed796b9 (F1–F4)

**New commits on `lane/regexguard`, in one push:**

| sha | subject |
|---|---|
| `99548d47` | fix(match): an undecided capture-typed `#match?` blamed "the regex engine abandoned the match" when nothing was abandoned, and named neither the text nor where it was |
| `6ecf2a6b` | fix(arch): an allow path-rule the engine could not finish refused an edge that no deny rule could forbid |
| `769003de` | docs(regex): the allowlist comment still named skillscan.h, and the libstdc++ residual quoted a probe number as if it were this tool's (head) |

main was not merged, per the train protocol. Against merge base `31e788ce`, the CHANGELOG diff is 62 insertions and 0 deletions; the entry stays in its slot after #245's entry.

### F1 — undecided capture-typed `#match?` gave the wrong cause and named nothing: FIXED in `99548d47`
- **Counter:** the single counter is replaced by `AstRegexUndecided`, a sink that workers write to only on the undecided path. It is guarded by a mutex; a query whose predicates all decide never takes the lock.
- **Cause:** `evalMatchPredicate` returns the outcome by cause: `TextScreened`, `TextUncompilable` or `Abandoned`.
- **First site:** `noteUndecidedPredicate` records the first site as the lowest fileId, then the lowest byte, so the named site is the same on every run.
- **Wording:** `--match` and `--lint-rules` print a count per cause, then that site with its text and reason. The reviewer's fixture now reads:
  > ripwire: --match: 2 #match?/#not-match? evaluation(s) could not be decided (1 captured text(s) the structural screen refused as a pattern, 1 captured text(s) that do not compile as a pattern); the first, at a.c:4, used the captured text '"(a+)+z"' as its pattern, which the structural screen refused: catastrophic backtracking: …

  "The regex engine abandoned the match" now appears only for a real abandon: "ran the pattern '…': the regex engine abandoned the match …".
- **Unchanged:** the built-in packs pass a null sink, and the legacy "filter nothing" behaviour stays for them.
- **Arm (h), one run per cause:**
  - screen-refused text: two files; a.c:4 is named ahead of b.c; the refusal is byte-identical run to run
  - text that does not compile
  - the same through `--lint-rules`, which names the rule `rx-capture`
  - an abandoned match via `RIPWIRE_FAULT_REGEX_MATCH=1`

  Each run asserts the cause, the site and text, that no cause is claimed that didn't happen, and that no answer element is printed. The previous head's Release binary fails 9 of these rows.

### F2 — allow rules decided first, so an allow the engine could not finish refused a determinable answer: FIXED in `6ecf2a6b`
`pathRuleForbids` now looks for the first matching deny before consulting any allow:

| Situation | Result |
|---|---|
| No deny matches | Permitted; no allow is run |
| A deny is undecided before any deny matches | Refuses, naming that deny, unless an allow matches |
| A deny matches and some allow matches | Permitted, even after an undecided allow |
| A deny matches, no allow matches, an allow was undecided | Refuses, naming that allow |

Every decided edge gets the old answer and the old label (the first matching deny).

**Arm (i):**
- Portable: the reviewer's exact rules file, `allow path zz/.* -> (a|a)+z` plus `deny path nomatch/.* -> .*`, answers at exit 0 on every standard library, because the allow is never run.
- On libc++ only: when a deny fires, an abandoned first allow followed by a matching second allow is permitted.
- On libc++ only: when a deny fires and the abandoned allow is the only allow, the run refuses and names that allow.
- Red rows on the previous head: 2 FAIL.
- Red rows on origin/main: 3, all rc 134.

### F3 — allowlist comment: FIXED in `769003de`
`src/regexguard.h` now says the allowlist is one row (redact.h), and says why skillscan.h left it: skill files are untrusted, so an abandoned match must fail closed.

### F4 — "26,624 bytes": FIXED in `769003de`
The CHANGELOG now reads "from about 26 KB, build-dependent". It gives the standalone probe's 26,624 bytes (13,312 for `(a|b)*c`) next to the reviewer's runs of this tool's own Linux builds: between 27 KB and 35 KB for the gcc dev build, and between 35 KB and 45 KB for the clang Release build. The entry's gate sentence also lists arms (h) and (i), and the red-on-origin/main count is recounted with them.

### Reviewer notes, recorded here without code changes
- **Skill-scanner rewrite, independently verified:** 400,000 random lines with 74,723 positives showed 0 mismatches. An exhaustive run over 597,870 strings of up to 6 tokens also showed 0 mismatches.
  - The only divergence the reviewer found was a `'\n'` inside the line, which `scanSkillText` never passes.
  - `769003de` makes `hasNetExfilShape` end a segment at `'\n'` as well as `'\r'`, so it equals the regex on any input.
- **`SCAN-INCOMPLETE` cannot be forced by an attacker:**
  - On libc++, the budget throws only after more than 4,096 × (remaining bytes) steps in one attempt, and every remaining constant pattern is linear per attempt.
  - libstdc++ never throws.
  - A skill's author controls the verdict anyway, and `ripwire wrap --force` bypasses it.
  - Only the fault switch or a `bad_alloc` reaches that finding.
- **Out of this lane, for the owner's decision on the libstdc++ matcher:**
  - `redact.h` is allowlisted as "never user text", but its subject is repo text, and a long token run reaches the same per-character recursion. It belongs to the same engine decision, not to this PR.
  - Policy question: #252 treats a skills walk that stopped early as WARN, while #251 treats an undecided line as CRITICAL. It may want one policy.
  - Train note (#253): both lanes make `file()` root-relative. The resolution at train time keeps #251's guarded per-file loop with `rootRelPath(ing, fileId)` as the subject and drops #251's `Eval` constructor parameter and its call site.

### Evidence at `769003de`
- **Gate (dev build):** `regexguardcheck` **115 PASS**; origin/main `31e788ce` gets **72 FAIL**, ten of them a signal death (rc 134/138).
- **Byte-identical:** **45/45 against `31e788ce`**, rerun on the `769003de` binary itself. The only differences are the 4 intended fixes.
- **`--quality-delta=31e788ce..HEAD`:** **`gating="0"`**, 4 new-symbol rows.
- **Targeted gates:** 98 run, each as its own script with its exit code read; **all rc 0** (3,283 PASS rows). They include regexguardcheck, astqueryregexcheck, archcheck, archmetricscheck, lintrulescheck, skillscan, deckcheck, limitstablecheck, readmedriftcheck, gateexitcheck, manifestcheck, gatecountcheck and printffmtparitycheck.
- **Not rebuilt:** Release and ASan were not rebuilt this round. None of the three commits touches the grep scan path, and CI's asan legs run these gates.

joyful-ii-V-I:
This PR is part of integration train 1b, #277, and will show as merged when that train lands.

PR #252, opened 2026-09-16
fix: a deep or odd-shaped argument, source file or skills tree could crash or stall a verb — six fixes, each red-first

joyful-ii-V-I:
This is the second half of the crash sweep, split from the cache/sidecar/file-handle PR by area. Each defect below takes ripwire down, or holds it for minutes, when a CLI argument, a repository file or an MCP request is nested or shaped past what a recursive or throwing path survives. Each was reproduced on the base, and each fix extends the gate that already owns the verb. The commits are one per defect, plus a CHANGELOG commit.

| # | trigger | before | after | gate |
|---|---|---|---|---|
| 1 | `--graph-query` nested 50,000 levels | recursive-descent evaluator overflows the stack: SIGSEGV (exit 139) | refused before evaluation past 256 levels, exit 1 with the reason | `graphqueryrefusecheck` 6 |
| 2 | a `--layout` extent: `#define` 200,000 `(` deep / `(0-2^40)*2^23/(0-1)` / `2^40*2^40` | SIGSEGV (139) everywhere / SIGFPE (136) on Linux x86-64 / signed overflow (abort under ASan) | checked arithmetic, `INT64_MIN/-1` refused, parens bounded at 64: an unknown extent with its caveat | `layoutcheck` §12 |
| 3 | `--eval-skills` over a skills dir with a `SKILL.md` link loop, a directory link loop, a mode-000 skill | uncaught `filesystem_error` from the throwing overloads: SIGABRT (134) | `error_code` overloads and `increment(ec)`; skipped entries are named on stderr | `skillevalcheck` |
| 4 | `ripwire wrap <agent>` in a repo whose `./skills` nests past the path-name limit | `operator++` throws inside a `noexcept` function: `std::terminate` (134) | `increment(ec)`; the early stop is disclosed and scored WARN; the recipe still prints | `codexwrapcheck` |
| 5 | `--match` (or a `--lint-rules` query) nested 4,000 levels | `ts_query_new` recursion on a 512 KB worker stack: SIGBUS (138); 2,000 levels ran past a minute | refused before any compile past 256 levels, exit 1 | `matchgrammarcheck` 6 |
| 6 | `--slice` / MCP `slice` on a function of 4,000 chained `if (x)` | cubic walk: 1,000 → 5.7 s, 2,000 → 48 s, 4,000 → killed at 2 min (server wedged) | syntax depth measured with a cursor first; past 512 levels, refused | `slicecheck` (15) |

The four new bounds are registered in `docs/LIMITS.md` as BOUNDARY, since each one refuses an input and truncates nothing.

**Runtime checks.** Only (2c) trips a sanitizer that a test reaching the path would have caught (UBSan in the G1 build). (1), (2a) and (5) are stack exhaustion, which ASan reports only as the crash itself. (3) and (4) are throws from `std::filesystem`'s throwing overloads, and clang-tidy's `bugprone-exception-escape` flags (4) statically, because the throw is inside `noexcept`.

**Verification.** Every new arm was run red on the base binary and green on this branch, which is merged with main 31e788ce, on both the plain and ASan builds. These gates also passed, rc 0: `gateexitcheck`, `manifestcheck`, `gatecountcheck`, `limitstablecheck`, `limits_build.py --check`, `querycheck`, `mcpslicecheck`, `sliceflowcheck`, `slicediffcheck`, `matchcapturecheck`, `lintrulescheck`, `skillevalsplitcheck`, `wrapverbscheck` and `opencodewrapcheck`. `--quality-delta=<merge-base>..HEAD` reports `gating="0"`: 11 minor rows, and one non-gating dead-code row that is the `for` loop's own iterator. No new gate, and no kParserVer, cache or qsnap-scheme change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented crashes and hangs caused by deeply nested graph queries, match specifications, slice targets, and layout expressions.
  - Invalid arithmetic and excessive parentheses in layout expressions are now safely rejected.
  - Skill discovery now skips unreadable, looping, or non-regular entries instead of aborting.
  - Skill-directory scans stop safely on traversal errors while still producing available output and warnings.
  - Excessively nested slice definitions are refused with a clear error.

- **Documentation**
  - Updated documented safety limits for query, layout, and slice processing.

- **Tests**
  - Added regression coverage for hostile inputs, filesystem errors, overflow, and nesting limits.


joyful-ii-V-I:
Disposition for the 12 red jobs on `6369f88e`: one push, two new commits.

1. **`graphqueryrefusecheck` arm 6 — "Argument list too long" (ubuntu), fixed in `94e34236`.** Linux caps one argv string at MAX_ARG_STRLEN (128 KiB), and the 50,000-level `kind(` chain was about 400 KB. `--graph-query` has no file or stdin form, so the arm now uses the form that nests the most frames per byte, `cx(…all,1)`: 20,000 levels in 120 KB. It runs under `ulimit -s 2048`, so without the bound the depth overflows regardless of a build's frame size; the fix refuses before evaluating.
   - Red on the unfixed evaluator: exit 139 on macOS (dev, 8 MB and 2 MB stacks) and in the `rw-linux-x64` container, clang release and gcc builds, 8 MB and 2 MB.
   - Green here: exit 1 with the bound named.
2. **`codexwrapcheck` — "the early-stopped skill scan was not disclosed" (ubuntu), fixed in `94e34236`.** The PATH_MAX fixture was macOS-only. The trigger is now the descriptor limit: 64 nested directories under `ulimit -n 24`. The iterator holds one open directory per level on both standard libraries, so it runs out of descriptors the same way everywhere.
   - Red on the unfixed walk: exit 134 on macOS (libc++), and on Linux clang and gcc (libstdc++: "cannot increment recursive directory iterator: Too many open files").
   - Green here: exit 0, the early stop disclosed, recipe printed. The arm does not depend on root, because EMFILE is not a permission check.
3. **`readmedriftcheck` L2, fixed in `dea01caf`.** `limits_build.py` does not write that README sentence, so it was updated by hand to 217 caps. `limits_build.py --check` agrees (217 caps, 7 parameters).
4. **`capsweepcheck` (C), fixed in `dea01caf`.** Ran `python3 bench/capsweep/capsweep.py emit`; `docs/TUNING.md` goes from 131 to 134 declarations.
5. **`recentscopecheck` arm 10e: unrelated flake, left unchanged.** The arm runs `--rank-by=churn-decay --in … --exclude … --since … --limit=39 --offset=1` and asserts the scoped block's `next=` cap. This branch touches none of that path: its source changes are the query.h evaluator, layout.h, skilleval.h, wrap.h, the astquery nesting check, slice depth, and the matching refusals in verbs_navigate.h and mcpverbs.h. The arm passed on the other 7 legs, and 4 of 4 local runs of the whole gate on this branch passed.

**Verified locally at `dea01caf`, all rc 0:** `limits_build.py --check`, `capsweepcheck`, `readmedriftcheck`, `graphqueryrefusecheck`, `codexwrapcheck`, `gateexitcheck`, `manifestcheck`, `recentscopecheck` (×4). Per the machine rule, no local ASan build for this PR; CI's ASan legs cover it.

joyful-ii-V-I:
Disposition for the independent review and the three unanswered CodeRabbit threads at `dea01caf`. One push, six new commits, no main merge.

**CodeRabbit threads**
- **layout.h:713, `parens` never restored.** Valid, and it was a regression: 66 sibling parenthesised terms sized to 268 on main and came back unknown here. Fixed in `b8bd2463`: the counter is now `parenDepth`, raised on entry to a `(` level and restored when it closes, so 64 bounds the nesting depth. layoutcheck §12 adds WideExtent (a 264-byte field, `modeled="1"`), red at `dea01caf` and green here.
- **skilleval.h:186, an `is_directory` error skipped silently.** Valid. Fixed in `05701a5e`: the entry's error is named on stderr before the skip, and an unlistable root now says "cannot list". skillevalcheck asserts both ends of the directory link loop are named; red at `dea01caf`.
- **wrap.h:510, `skip_permission_denied` hides a sealed subtree.** Valid; it predates this PR and is fixed now per the fix-now rule. Fixed in `f7149be5`: a folder the scan cannot enter is named as a WARN on stderr, the score rises to at least WARN, and its siblings are still scanned. codexwrapcheck seals a folder holding injection text next to a readable one and requires exit 0, the named WARN and the recipe (SKIP by name as root); red at `dea01caf`.
- **codexwrapcheck.sh:42, the path fixture never trips on Linux.** Already fixed in `94e34236` (the EMFILE fixture), red on macOS and on Linux clang and gcc.

**Slice bound (i): chose the linear walk, not a raised cap.** Fixed in `52359644`.
- **Cause:** every occurrence climbed to its statement anchor, scope and classifier shape through `ts_node_parent`, which descends from the tree root on each call. That made the walk cubic.
- **Change:** `sliceScanDefinition` builds a parent table in one cursor pass (keyed on `TSNode::id`, the address `ts_node_eq` compares) and installs it through an RAII scope. The 19 parent calls go through `sliceParent`, which falls back to `ts_node_parent` for any node outside the table. The statement anchor is memoized per node.
- **Measured, plain build:** 2,000 / 4,000 / 8,000 nested ifs in 0.05 / 0.06 / 0.08 s (the old walk took 48 s at 2,000 and did not finish at 4,000). The 808-level chain takes 0.06 s (11.6 s on main).
- **Parity:** output is byte-identical to main on 17 slices of this repo's functions, 7 of them with `--slice-flow=both`, and on the chains.
- **Why a bound remains:** the walks still recurse once per level, so it is a named **stack** guard at 4,096 levels. 8,100 nested ifs overflowed a 2 MB stack and fit in 4 MB, so 4,096 sits inside half the default 8 MB main-thread stack, with room for sanitizer frames. The deepest function in the reviewer's 47,795-file survey is 808. Both refusal messages now say "would exhaust the stack"; the cube claim is gone.
- **Gate:** slicecheck (15a) 3,000 nested ifs are answered inside 30 s; (15b) an 806-name chained assignment (the CPython shape, ~808 levels) is answered, the must-answer arm; (15c) 6,000 nested blocks are refused by name. (15a) and (15b) are red at `dea01caf`; (15a) is also red on main.

**Found on main (j), fixed in `bac39c46`.** `--layout` treated any statement containing `(` as a member function. So `char a[(4)];` (and, found while fixing it, `int x = (3);` and `int x{ (3) };`) dropped the field while still reporting `modeled="1"` and a short size. A `(` now opens a parameter list only before the first `[`, `=`, `{` or bitfield `:`, and `operator` members stay functions. layoutcheck §13 covers ParenExtent, ParenInit and ParenBrace (2 fields, size 8), plus an OperatorAssign control; red on main and at `dea01caf`.

`e90643f1` updates the CHANGELOG entry. `docs/LIMITS.md` is regenerated for the 4,096 value; the README cap count stays 217.

**Verified on the tree identical to `e90643f1`, all rc 0:** `layoutcheck`, `slicecheck`, `mcpslicecheck`, `sliceflowcheck`, `sliceflowsenscheck`, `slicediffcheck`, `skillevalcheck`, `skillevalsplitcheck`, `codexwrapcheck`, `wrapverbscheck`, `readmedriftcheck`, `capsweepcheck`, `gateexitcheck`, `manifestcheck`, `limitstablecheck`, and `limits_build.py --check`. No local ASan (machine rule).

joyful-ii-V-I:
Disposition for the re-review at `e90643f1` (slice stack safety). One push, one new commit: `1d93429b`.

**`kMaxSliceDepth` 4,096 → 2,048** (coordinator call).
- **Why:** the re-review measured about 4,085 levels needing 3,660 KB for nested loops on plain arm64 (~870 B a level). Every slice path runs on the ~8 MB main thread, ASan frames are 2–3× wider, and CI's asan job doesn't run slicecheck, so 4,096 levels wasn't shown to be ASan-safe.
- **At 2,048:** the worst shape needs about 1.8 MB plain. That still leaves the bound 2.5× the deepest real function (808). The refusal wording is unchanged ("would exhaust the stack"). `docs/LIMITS.md` and `docs/TUNING.md` are regenerated.

**slicecheck (15)**, all four red at `dea01caf` and green here:
- (15a) 2,000 nested ifs: answered.
- (15b) **new:** 2,040 nested `for` loops, the widest frame per level, must be answered. 2,043 are answered and 2,044 are refused, so the arm sits at the guard's edge. It needs 2 MB of stack on this build (1.5 MB overflows).
- (15c) the CPython-shaped ~808-level chain: must answer.
- (15d) 6,000 blocks: refused by name at 2,048.

**Allocation note (optional, done):** the parent and anchor tables are reserved from the definition's `ts_node_descendant_count`, and the anchor climb reuses one scratch vector. Slice output is still byte-identical to main (9 `--slice-flow=both` slices).

**Verified, all rc 0:** `slicecheck`, `mcpslicecheck`, `sliceflowcheck`, `sliceflowsenscheck`, `slicediffcheck`, `readmedriftcheck`, `capsweepcheck`, `limitstablecheck`, `gateexitcheck`, `manifestcheck`, `limits_build.py --check`. No local ASan (machine rule); running slicecheck with `RIPWIRE_BIN=asan/ripwire` is what would prove (15b) under sanitizer frames.

The `--layout` field drops on main (`alignas`, `__attribute__((aligned))`, `decltype`, `std::function<void(int)>`) are noted as the separate follow-up lane, not this PR.


---

**Addendum, `2acfe5c9` (the ASan run):** `RIPWIRE_BIN=asan/ripwire test/slicecheck.sh` at `1d93429b`, with CI's sanitizer options minus `log_path`, **failed**. (15a) and (15b) aborted with an AddressSanitizer stack-overflow in `SliceRdWalker::loop → stmtC → stmt`. ASan frames are ~10× plain, not 2–3×. On the main thread, 2,040 nested loops need 18.8 MB, ifs 14.0 MB and else-if chains 8.5 MB. At 8 MB, loops overflow past ~880 levels; at 5.3 MB (a 1.5× margin), past ~570. A ~560 guard would refuse the 808-level CPython chain (15c). So `kMaxSliceDepth` stays at 2,048, and a definition deeper than 256 levels now walks on a joined thread with its own 64 MB stack. Measured on that thread: loops need 18.6 MB under ASan (3.5× margin), ifs 13.8 MB (4.7×), plain loops 1.8 MB (37×). If the thread can't start, the slice is refused by name. (15b) now runs with the caller's stack held to 1 MB, where a plain build of the old walk overflowed. slicecheck passes on plain and on ASan (44 PASS). The gates listed above were re-run at `2acfe5c9`, all rc 0.

PR #253, opened 2026-09-16
fix(resolve): the same tree answered differently under `ripwire .` and `ripwire "$PWD"`

joyful-ii-V-I:
Refs #228

Thanks to @hnipps for the report. This is the root-spelling half of it; the checkout-shape half (export-ignore, submodules, sparse checkouts, skip-worktree, one-sided `--no-ignore`) stays open for the help-wanted kit.

### The defect

The crawl stores every path with the root exactly as it was typed, and the include/import index and the path predicates read that spelling raw. So how you spelled the root changed the answer:

- **Python root-relative imports** joined onto an empty base, which is the crawl root only under `ripwire .`. Under `"$PWD"` (every MCP session, and the `--quality-delta` HEAD side, which always ingests at an absolute temp root) `from pkg.store import load` stopped resolving and the name ladder bound a same-directory `load`. `ripwire . --quality-delta` gated on an unchanged four-file tree.
- **A root typed `../repo`** stored `../repo/x.h`, which `lexicalNormalize` refuses, so every index key was empty and every include/import in every path-resolved language (C/C++, Python, TS/JS, Rust, Bash, Ruby, Lua) went unresolved.
- **Directories above the root decided predicates.** `isTestPath`, `isFixturePath`, `builtinLayer` and `pathTierOf` walked the typed prefix: a checkout living under `tests/` or `fixtures/` was all tests (`layer="test"`, dead-code exempt, `--affected` seeds, demoted in ranking) under an absolute root and none of that under `.`.
- **Displays and byte charges.** Route anchors printed `/.../x.h` under an absolute root; `--pack-task` callee rows printed the whole path under `"$PWD/"`; the `--lint` cap and the map's byte model charged the typed root on every path, which flipped a `--top-k=300` Django map to `important-last(auto:fill)` under `"$PWD"` alone.

### The fix

One seam. `ingest()` records the root once (`IngestResult::crawlRoot`), and `rootRelPath` (`src/model.h`, beside `relForHash`, which moved there) is the root-relative view: a prefix compare, no allocation, no syscall. The precise include adjacency, the decl-to-def include proof, the JS import / Python external-veto indexes, the module vocabularies, the test/fixture/layer/tier predicates, the mention and stack-trace suffix matches, the lint cap and the byte models read it. Stored paths, disk opens, cache keys and printed `p=` are untouched; `root=` still prints as typed. `rootRelativeUri` trims a trailing `/`. `kQSnapCacheScheme` 12 → 13 so a pre-fix HEAD Snapshot is never served.

Rather than canonicalising the root to an absolute path (which would make every spelling agree on the *wrong* Python answer), every spelling now gets what `ripwire .` always gave.

### Sweep

Differential over the 161 committed fixtures × 27 verbs × 7 spellings (`.`, `./`, `"$PWD"`, `"$PWD/"`, symlink, `$PWD` from another cwd, `../name`), outputs compared after normalising only the printed root (and `est_tokens=` / expand's `file NB < bundle NB`, which price the root= bytes):

| class | verbs that diverged on main | languages | root cause |
| --- | --- | --- | --- |
| Python root-relative import | deps, impact, callers, affected, quality-delta, map | Python | empty-base probe matches only under `.` |
| `../name` root | deps, impact, callers, callees, uses, map, metrics, seams, safe-delete, grep, test-gate, dead-code, quality-delta | C/C++, Python, TS/JS, Rust, Bash, Ruby, Lua (Elixir resolves by module name: unaffected) | `lexicalNormalize` refuses the `..` prefix; every key empty |
| directory above the root | map/around/metrics (`layer=`), affected/test-gate/exercises/seams/safe-delete (test seeds), dead-code/quality-delta (fixture exemption), for/pack-task (tier) | all | predicates read the typed prefix |
| anchor display | for, pack-task | all | `routeAnchorPath` elided the absolute prefix |
| byte charges | lint (cap), map (fill order) | all | raw path length |

main 31e788ce: 465 diverging (fixture, verb) pairs. This branch: 13, all accounted for — 11 are a pre-existing stderr-only `DEGRADED_PATH_ALERT` in `--for` under `.` (stdout identical, same on main), 2 are specifier text in source (`./a/b`) that the normaliser itself rewrote. A second sweep with the fixtures staged under `tests/fixtures/bench/static/docs/captures/min/migrations/` went from 199 to 0 real divergences.

Classified and left alone (invariant by construction): cache keys and baseline keys (already `relForHash`), `isTestPartnerOf`/`runnerVerb`/`baseNameOf` (basename or extension only), the S6-C locality tie-break and trace-locus depth (a shared prefix shifts every candidate equally), multi-root (labels are already root-relative), Go/C#/Java/Swift/PHP/Dart (no path-resolved tier).

### Measured

Django shallow clone 2b30f62 (3,449 indexed files), `--no-cache` map header:

| spelling | main edges / ambiguous / declined | this branch |
| --- | --- | --- |
| `.`, `./` | 74,972 / 5,958 / 40,681 | 74,972 / 5,958 / 40,681 |
| `"$PWD"`, `"$PWD/"`, symlink | 62,591 / 3,135 / 49,153 | 74,972 / 5,958 / 40,681 |
| `../dj` | 63,357 / 3,424 / 50,269 | 74,972 / 5,958 / 40,681 |

The six maps are one document after root normalisation, byte-equal to main's `.` map. Under `"$PWD"` that is 12,381 more (caller, callee) edges, and `ambiguous=` rises with them because more calls reach a definition set at all. `--quality-delta` with a fresh cache: main gated 14 rows under `.`/`./`, 8 under `../dj`, 0 under the absolute spellings; this branch reports 0 under all six. As a consequence of the predicate fix, a `TMPDIR` inside a `fixtures/` directory no longer exempts the HEAD tree (research repo: 4 gating rows → 0).

Before/after under the absolute spelling across the fixture sweep: 4,038 (fixture, verb) outputs identical, 309 changed, and every one changed to main's `.` answer (153 `--for` + 153 `--pack-task` anchors, pyimportprecisefix `--deps`/`--impact`, expandmodefix `--lint` cap); 0 changed otherwise. `est_tokens=` on the map and pack-signatures also moves under an absolute root, because the byte model no longer charges the typed root per path.

### Gates

New `test/rootspellingcheck.sh` (registered in regression.sh, gate count, shard weights): six spellings byte-identical for map / --deps / --callers / --impact / --affected / --quality-delta on the committed four-file repro (`test/rootspellfix`) and eight language import fixtures, plus --dead-code / --for / --pack-task on the repro and a cross-directory C++ header selector; positive controls (named edges, `graph_ambiguous="0"`, exit 0 with `regressions="0"`); a real-edit sensitivity arm (re-pointing the import gates, exit 2, under `.` and absolute); a placement arm under `tests/fixtures/bench/`; and a scheme-upgrade arm (`RIPWIRE_PREFIX_BIN=<pre-fix build>`: the old binary writes the qsnap blob, the new one must not serve it).

- origin/main: 57 FAIL / 29 PASS. This branch: 88 PASS (86 + the 2 upgrade rows). The upgrade arm was RED on the fix before the scheme bump; the C++ header arm was RED on this branch's first commit, which keyed the decl-to-def index root-relative but resolved the includer as typed.
- Full suite on the merged tree: `gates=634 pass=631 skip=3 fail=0` (argvdiffcheck, editchecknotecheck, g1freshcheck).
- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: `gating="0"`.
- ASan over the touched paths (58 runs, 4 spellings × 12 verbs): no reports.

### Pins moved

- `test/golden.xml`: the six `layer="test"` attributes are gone (the `test` was the directory the root `test/fixture` sits in); est 895 → 863.
- `test/fillordercheck.sh` 895 → 863 (same reason).
- `test/printf_parity.manifest`: flagless, expand, around, pack_signatures, pack_task, seams (all `test/fixture`, same reading; `UPDATE_GOLDEN_EXPECT` matched exactly).
- `test/forrankordercheck.sh` fixture/geometry 3319 → 3494, attributed three ways in its RE-PIN note.
- `test/partitioncheck.sh` arm 8: N 4 → 5 (K<N held at 4 only because `test/` above the root tiered every file as a test).
- `test/verifycheck.sh` layer arm and `test/regression.sh` arm 3l: moved to `test/archfix`, whose layer directories are inside the tree.
- `test/candheadcheck.sh`, `test/defoverdeclcheck.sh`, `test/routecheck.sh` (f1, f4): anchors root-relative (`candheadfix/a_gold.hpp` → `a_gold.hpp`); choices and `+N` unchanged.
- `test/qschemetrip.hash` re-pinned with a RE-PIN LOG entry; `kQSnapCacheScheme` 12 → 13.

## Summary by CodeRabbit

* **Bug Fixes**
  * Results are now consistent regardless of how the crawl root is written, including relative, absolute, trailing-slash, symlinked, and parent-directory paths.
  * Improved import, include, module, anchor, test, fixture, layer, and file-path handling for roots outside the current directory.
  * Output sizing and quality classifications now use displayed root-relative paths.
  * Updated cached quality results to prevent stale data after these changes.

* **Tests & Documentation**
  * Added comprehensive root-spelling coverage and updated regression fixtures.
  * Updated documented test-gate counts from 619 to 620.


joyful-ii-V-I:

PR #254, opened 2026-09-16
fix(resolve): an explicit receiver of unknown type was pinned to the caller's own class, and a std:: type narrowed to an in-repo namesake

joyful-ii-V-I:
**Stacked on #248** (merged in at `186fcb84`) (`lane/param-receiver-binding`), which this PR targets; the diff below is only this branch. It closes the two holes #248 left open, and **it narrows #248's qualifier guard**: measured on rocksdb, refusing every qualified parameter type refused ~94 correct narrows, so a parameter now refuses only a `std::` type, the same rule locals get here. Arm 17's fixture moves from `ext::map` to `std::map`, and `ext::map` becomes arm 24, a pinned stated floor.

## The defects

1. **An explicit receiver of unknown type was pinned to the caller's own class.** `auto other = make(); return other->pick( 1 );` inside `Decoy` answered one precise edge to `Decoy::pick`, no `amb=`. The S6-C locality tie-break excluded field-of-this, field-of-var and `super()` receivers, but kept every NamedVar on the premise "a typed var already narrowed above". A receiver no rule typed never narrowed, so the scope-segment credit handed the tie to the caller's own class.
2. **A `std::`-qualified local type narrowed to an in-repo namesake.** `std::map<int, int> table; table.find( k )` pinned an in-repo `map::find`: Rule 2's flat per-function table records a written type's final segment, and class names carry no namespace. Constructor-inferred locals (`auto t = std::map<…>()`) and C++ assignments (`t = std::map<…>()`) did the same. The assignment record did not carry the qualified text at all.

## The change

- **(1)** A NamedVar receiver whose type no receiver rule (2, 2c, 2b) established keeps the tie-break credit only up to the end of the caller's FILE segment (`resolve.h receiverLocalityCap`). Same-file and same-directory locality still count; the class segment does not. `this->` and bare calls are unchanged.
- **(2)** `resolve.h namesStdType`: a type written in namespace `std` never narrows. `std` is reserved to the implementation, so no in-repo class *is* a `std::` type. The lexical (parameter) lookup refuses it; the flat table tombstones the variable, since it cannot tell which declaration is in scope. **Every other qualifier narrows**, parameters included. The C++ assignment now records its constructor's qualified text (`pushTypedBind` + `declaredTypeOf`), so **kParserVer 97 → 98** (+ quality.h mirror; format unchanged).
- **(3) Disclosure (owner decision).** Every edge a qualified, non-`std` written type chose by its last name — Rule 2's narrow (parameter or local) or CHA-lite's cone prune by that type — carries **`prov="final-segment"`**. The qualifier was never checked against the class's namespace, so arm 24's wrong edge and arm 23's correct one both read as a guess, not as a uniquely resolved name. The flat table records whether a declaration wrote the type qualified (`resolve.h FlatRecvType`), `Narrower::finalSegmentTypeAt` answers it per site, and `resolve.h edgeProvenance` now owns the whole scip > binding > import > split > final-segment precedence. Both map legends define the value: `final-segment(last-name-match;namespace-unchecked)`.
- graph.h stays line-neutral against #248 (+32/−32), so the 0.6.1 showcase seed `--at=src/graph.h:3406` still resolves to `rankGraphTeleport`.

## Measured first (`--pin-census --no-cache`)

**Design choices, each measured on the #248 tip with an env switch before the code was written:**

| choice | result |
|---|---|
| (1) cap at the file segment **vs** skip the tie-break for these receivers | 0 target differences on all 7 corpora; skipping relabels every {caller itself, X} tier `locality` → `unique` and drops `lpin=` (262 rocksdb, 53 private, 24 django, 101 rails). **Cap kept.** |
| (2) refuse ANY qualifier on locals (the literal ask) | refused 424 narrows on rocksdb, 27 private, 9 src/. Joined to each receiver's qualified text, every rocksdb/src/ one was a non-`std` qualifier (`ROCKSDB_NAMESPACE::Status`, `rw::notes::NoteIndex`) and every sampled one was correct. It also minted new wrong ladder pins (`value_slice.data()` → `MultiGetJNIKeys::data`). The only wrong narrows it removed on the three corpora were 6 `std::map` locals. **Rejected.** |
| (2) refuse only `std::` | removes exactly those 6; moves 0 sites on rocksdb and src/. **Chosen.** |
| #248's parameter guard: any qualifier **vs** `std::` only | `std::` only resolves 94 more rocksdb sites (e.g. `WriteBatch::Handler* handler; handler->MarkCommit( xid )` had been pinned to `WriteBatchInternal::MarkCommit`), 12 private, 5 src/; all sampled correct but ONE (an `ankerl::unordered_dense::map<…>&` alias template → the in-repo `map`: the stated floor, arm 24). |

**Final binary vs the #248 tip:**

| corpus | sites changing target | breakdown |
|---|---|---|
| rocksdb | 211 | 117 locality decisions → splits (14/14 sampled pins wrong, mostly delegation: `rep_->Name()` in `Wrapper::Name`); 94 in-repo qualified parameters narrow. `bound=` 200,009 → 200,036 |
| private C++ corpus | 88 | 70 locality → splits (14/16 sampled wrong; one right pin is now a 3-way split that keeps it); 6 `std::map` locals stop narrowing; 12 qualified parameters narrow (1 is the floor) |
| django | 72 | locality → splits (`old_ids.add( obj )` had pinned `ManyRelatedManager::add`) |
| rails | 145 | locality → splits (`conditions.freeze` → `Callback::freeze`) |
| src/ | 5 | splits → Rule-2 pins through `notes::`/`rw::quality::` parameters (all correct) |
| vue-core, go/net | 0 | — |

**Byte cost of the disclosure** (rocksdb @ 0e2801ac3; cfa5792f's binary vs the final one; a normalised diff shows the attribute, the legend term and `est_tokens` are the only differences):

| rocksdb output | before | after |
|---|---|---|
| default map | 31,366 B | 31,711 B (+345, **+1.10%**; 14 marked edges) |
| `--for="write batch handler mark commit"` | 8,746 B | 8,746 B (no `prov=` in bundles) |
| whole graph, `--top-k=100000` | 5,406,160 B | 5,413,666 B (+7,506, +0.14%; 355 marked edges) |

A private corpus's default map pays only the legend term (+51 B). The first legend reading cost +1.29% on rocksdb; it was shortened on the coordinator's call.

**A name-only collision guard was measured and rejected.** The Iterator-shaped collisions #248 states (a `rocksdb::Iterator*` narrowing onto five unrelated `memtable/…::Iterator` classes) remain `amb=`-disclosed splits. The guard moved 967 rocksdb rows off those namesakes, but it declined 58 correct platform-alternate splits (`port::Mutex::Lock` over posix and win) and 8 correct `log::Writer` narrows, minted 10 new wrong unique pins (`cv.Wait()` → `AsyncFileReader::Wait`), and declined 132 correct `Template.render` splits on django. The real fix is the namespace chain in `Symbol::scope`; the evidence and a narrower pure-virtual variant are handed to a follow-up lane.

The assignment capture moved no site on these corpora; arm 21 is its only witness. Wall time on rocksdb is within noise (3 cold runs each at load avg 42: 1.69–2.19 s before, 1.73–2.80 s after). Determinism (map + census twice, `cmp`), `xmllint`, and warm == cold on rocksdb all check out.

## Red first

- `test/localitycheck.sh` arms 5–9 (generated fixture): **5, 6, 7 RED on the #248 tip** (untyped local / member of an unreadable type / typed with no such method — each a locality pin to the caller's class). **8 RED on the skip-the-tie-break variant** (pins the cap design). 9 is the Rule-2 control.
- `test/narrowcheck.sh` arms 17–24: **19, 20, 21, 23, 24 RED on the #248 tip**; **21 RED without the assignment capture**.
- `test/narrowcheck.sh` arm 25: **(25a) arm 24's wrong edge, (25b) arm 23's correct narrow, (25c) arm 22's local twin, (25f) map legend, (25g) compact legend — RED before `prov="final-segment"` existed**; (25d) an unqualified narrow and (25e) a uniquely named call carry no `prov=` (controls).

## Gates moved for the reason the fix exists

- `clsrecvcheck` (B)/(C)/(E) asserted the caller-own-class pin **stands** for `item.validate( v )`. They now assert the honest two-way split (the contrast with the firing route is unchanged); (F) reads `ambiguous=3` with no locality pin.
- `localitycheck`'s HIGH-1 probe was an untyped local that no longer earns scope credit, so it could not tell a byte-prefix tie-break from a segment-aware one. It is now `this->go()` in a class template with a dependent base, with a census liveness row (`pre=2`).
- `canoncheck`: the comment describing that fixture.
- `chacheck`, `chaconecheck`, `resolverhonestycheck`, `fieldnarrowcheck` pass unchanged: their untyped controls are scope-less free functions, which never reach the tie-break.

## Pins moved

- `test/qschemetrip.hash` → `8ca11f67…` (kParserVer 98; the gate's own re-pin case, equal to the red run's `current=`; RE-PIN LOG entry added).
- No gate count change, no printf parity rows. The map legend and compact legend each gain the `final-segment` reading; compactlegendcheck, legendcoveragecheck, legendcostcheck, legenddriftcheck and printffmtparitycheck pass unchanged. kParserVer and the qschemetrip hash are untouched since cfa5792f (train 2 assigns them).

## Stated floors (not closed here)

- A qualifier that is neither `std` nor the class's own namespace still narrows by final segment (arm 24). Closing it needs the namespace chain in `Symbol::scope`.
- The ladder's tier 1 still admits same-file candidates alone, so a delegation whose true target lives in another file lands on a same-file namesake before the tie-break runs.
- The same `std::` collision exists in Rule 2b's field-type table (`std::string name_; name_.size()` → an in-repo `string::size`, confirmed). Skipping the field at capture was built and measured: it un-tombstones same-named classes (`StringSource#contents_` on rocksdb: `std::string` in test_util/testutil.h vs `Slice&` in db/log_test.cc) and minted 4 wrong edges. It is handed off as its own task.

## Verification

- Full suite at `0585b2f3` (the head), one run, serialised behind another session's suite: `gates=632 pass=629 skip=3 fail=0 wall=937.9s jobs=6 tree_writes=0` — skips: `argvdiffcheck`, `editchecknotecheck`, and `g1freshcheck` (no local ASan tree, per the machine rule; CI's asan legs cover it). The earlier head `cfa5792f` ran `pass=630 skip=2 fail=0` with a fresh ASan tree.

- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: `gating="0"` (1 non-gating row: #248's `pushTypedBind` arity; the `buildGraph` complexity row is gone since `edgeProvenance` took the precedence chain).
- `docs/gatecount_build.py --check` and `docs/limits_build.py --check`: clean.
- ASan (`-fno-sanitize-recover=all`) at `cfa5792f`: localitycheck, narrowcheck, clsrecvcheck, plus map+census over src/ and every probe fixture; 0 reports. Not re-run locally for the later commits (machine rule); CI's asan legs run them.


joyful-ii-V-I:
**Disposition — pushed `cfa5792f..0585b2f3` (fast-forward, one push).** This applies the owner decisions relayed by the CI coordinator on the independent review of #248/#254.

| sha | what |
|---|---|
| `095fdd2d` | **F3, relax but disclose.** Every edge that a qualified, non-`std` written type chose by its last name (Rule 2, parameter or local, or CHA-lite's cone prune) carries `prov="final-segment"`, defined in both map legends. `narrowcheck` arm 25 was red before (the attribute on arm 24's wrong edge, arm 23's correct one, arm 22's local twin, and both legends); an unqualified narrow and a uniquely named call carry no `prov=`. graph.h stays line-neutral (+32/−32 vs #248), so the showcase seed still resolves. |
| `16007542` | **F4 and F5.** The CHANGELOG publishes the dynamic-language precision of the removed locality pins (django 8 of 14 wrong, rails 8 of 12; the right target stays inside every split) and states that this entry supersedes #248's parameter rule by name. |
| `3b38bfc2` | Merge of #248's `186fcb84`. The CHANGELOG is a union: #248's floors paragraph closes its own entry, and none of its lines is removed. |
| `0585b2f3` | The shorter legend reading, taken on the coordinator's call to bring byte cost down. Cost on rocksdb vs `cfa5792f`'s binary: default map +345 B (+1.10%, 14 edges), `--for` bundle 0 B, whole graph +7,506 B (+0.14%, 355 edges). Only the attribute, the legend term and `est_tokens` differ. |

**Not landed, on measurement:** a name-only same-final-name collision guard for the Iterator shape. It moved 967 rocksdb rows off the wrong `memtable` namesakes, but it declined 58 correct platform-alternate splits and 8 correct `log::Writer` narrows, minted 10 new wrong unique pins, and declined 132 correct `Template.render` splits on django. The evidence and a narrower pure-virtual variant go to a follow-up lane; the namespace chain is the real fix.

**Untouched on purpose:** kParserVer (98) and `test/qschemetrip.hash`, which train 2 assigns. main is not merged.

**Verified at `0585b2f3`:**
- Full suite `gates=632 pass=629 skip=3 fail=0`. The skips are argvdiffcheck, editchecknotecheck, and g1freshcheck (no local ASan tree).
- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD` is `gating="0"`.
- The legend/output gates (compactlegend, legendcoverage, legendcost, legenddrift, printffmtparity, docscommands, mcpclidiff, mcpattrparity) all exit rc=0.

**CodeRabbit:** #254 never got a review of its own, because its base was #248's branch. #257's full-range review covered this code with 0 actionable comments.

joyful-ii-V-I:
This lands through integration train 2, #265, with parser version 98. Resolutions are on that branch, so nothing more is needed here. This PR will show as merged when #265 merges.

joyful-ii-V-I:
Landed on `main` through integration train 2, #265 (merge commit `bcd3b016`). This PR's head `0585b2f3` is an ancestor of `main`. GitHub can't mark it merged because its base was #248's branch, and it refuses to retarget to `main` when there are no new commits, so it's closed here with its content on `main`.

PR #255, opened 2026-09-17
fix(quality): --quality-delta answered from a HEAD snapshot another build computed

joyful-ii-V-I:
## The defect

`--quality-delta` caches the Snapshot it computes for `HEAD` (the `qsnap` blob). The blob's dead set depends on **call resolution**: `quality.h`'s `isDeadCandidate` reads `g.inEdges`, and `pythonDispatchedMethodIds` reads the graph. The filename key folded the qsnap scheme, the extraction identity (`kCacheVersion` + `kParserVer` mirrors), the excludes, the file-size ceiling and the HEAD sha. A resolution-only change moves none of those, and by rule it bumps none, because a resolver runs over cached ingest facts. So two builds that resolve calls differently, sharing one cache dir on one repo HEAD, served each other's dead set.

**Verified with two real builds before any code:** main `a55b118e` as built, and the same tree with `graph.h`'s `keepStdQualifiedCandidates` (d39554dd's `std::` guard) switched off. Fixture: two files, where `std::launder( &v )` may or may not bind an in-repo `Pool::launder`. Shared `XDG_CACHE_HOME`:

| run | result |
| --- | --- |
| guarded build, cold cache | `regressions="0"`, exit 0 |
| guarded build after the unguarded build warmed the cache | gating `dead-code` row on the **untouched** `Pool::launder`, exit 2 (phantom) |
| unguarded build, working tree deletes the only call, cold | the real gating `dead-code` row, exit 2 |
| same, after the guarded build warmed the cache | `regressions="0"`, exit 0 (**real regression hidden**) |

Both warmed runs found exactly one qsnap blob in the shared dir, so one key produced two answers. This happens after upgrading ripwire on a repo whose HEAD has not moved, or with an installed `ripwire` and a local `./build/ripwire` on one checkout. d39554dd's own message reasons "no graph or edge blob is cached", so the dependency was missed after the author checked for it.

## The change

The qsnap and qbody keys, and the blob header they share, now carry a **producer identity**. It is the SHA-256 over every file under `src/` and `queries/`, derived on each build by `cmake/source_identity.cmake`. It is compiled in as one generated definition (`build/generated/source_identity.cpp`, declared by `src/sourceidentity.h`). `serializeSnapshot` writes `fnv1a64` of it after the extraction identity, and `deserializeSnapshot` refuses a mismatch: the same two guards P0.2 gave the extraction identity. `qheadsnap` (the ingest blob) keeps its key, since it holds extraction facts only.

**Why derived, not a `kResolverVer` to bump.** This cache's record: B10.1a's `isDeadCandidate` exemption (retired late, at v3) and 28c7d32's `kParserVer` bump (P0.2) each changed what a blob means without the bump that would retire it. None of the twelve scheme versions cites a resolution change. A derived identity renames the blob on any source change, whether or not anyone remembers what the change meant. That also covers semantics the `qschemetrip` manifest never listed (clone identity, `isPublicApi`, `canonicalIdRelTo`).

- **A generated `.cpp`, not a header.** A changed identity recompiles one tiny object instead of `main.cpp` and `ingest.cpp`.
- **`src/`, not `src/infra/`.** The header names this build, and `infraportcheck` (C) forbids that in the vendorable layer.
- **Header shape changed.** `kQSnapCacheScheme` 12 → 13, `kQBodyCacheScheme` 3 → 4 (the P0.2 precedent).
- **Out of scope, stated in the script:** `third_party/` (grammar changes move `kParserVer`, already in the key) and the compiler (the determinism contract).

## Measured

- **Identity script:** 49 ms per build, 191 files; equal to an independent Python derivation.
- **Cost of the rename** (ripwire's own tree, private XDG cache, five runs each, stdout asserted identical across all ten runs): warm `--quality-delta` median **3.04 s** (2.78–3.21); qsnap recomputed over a warm ingest median **5.12 s** (4.85–5.34). So an edit to a source file costs about 2.1 s once, on the first delta after the rebuild.
- **Two-build experiment re-run on the fixed tree** (fixed build vs fixed build with the guard off): one qsnap blob per build in the shared dir, both orders byte-identical to a cold cache, no corrupt alert (a clean miss).

## Gates

**New: `test/qsnapproducercheck.sh`** (5.2 s standalone). It forges a real warmed blob by *structure*: the header ends at the `fnv1a64(HEAD sha)` field, and a parse that does not land on the trailer is a FAIL. **F** and **G** are matched pairs that differ in one thing:
- dead keys dropped (phantom) or added (hidden) with the producer bytes kept: a control that must change the answer;
- the same forgery with those bytes flipped: must be refused and match a cold cache byte-for-byte.

**A–E** pin the key (and that `qheadsnap` stays off it), the header, the derivation from source content (path-independent copy: same identity; one appended line in `graph.h`: different), the blob's producer slot, and warm reuse.

- **Red first** (commit `03de352d`, main's code): 8 FAIL / 5 PASS. Both controls fire (F1 exit 0 → 2, G1 exit 2 → 0); F2 has no producer bytes to flip; G2 hides the regression (cold exit 2, forged exit 0); A/B/C/D find nothing.
- **Green** (`eb55bbc6`, and again on the merged `ae7e1e01`): `qsnapproducercheck: ALL PASS` (16 rows).
- **ASan:** the gate through a wrapper that tees every child's stderr: ALL PASS, 0 sanitizer reports. `cachefuzzcheck: ALL PASS`, including `qsnap ASan sweep: no sanitizer report`.
- **Targeted batch, plain build, all rc=0:** infraportcheck, qschemetripcheck, qextractionkeycheck, qsnapcachecheck, evictioncheck, qsnapprefetchcheck, editcheckcheck, g1freshcheck, manifestcheck, gatecountcheck, limitstablecheck, gitstampcheck, selfcontainedcheck, xmlwellformed, binoverridecheck, clonededupcheck, fixedbufsweep, portablebuildcheck, g1configcheck, headsnapcachecheck, qualityexcludecheck, qualitycheck, gateexitcheck, pargatescheck, readmedriftcheck, deckcheck, noaliascheck, crossdirincludecheck, ripwirepubliccheck (after commit), versioncheck (after rebuild).
- **Full suite** on `ae7e1e01` (main `31e788ce` merged), one single run, no concurrent suite: `gates=634 pass=632 skip=2 fail=0 wall=875.1s jobs=6 tree_writes=0` — the two skips are the environmental `argvdiffcheck` / `editchecknotecheck`; `g1freshcheck` ran (ASan tree fresh) and passed.
- **Quality delta** `--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: `gating="0"`. Two minor rows, both on `deserializeSnapshot` and both already past their bars before this change: complexity 36 → 38, verbosity 107 → 112 (the producer check).
- **Registered:** `test/regression.sh`, `.github/pargates-shard-weights.json`, and the generated gate count (620 after merging main's `diagnoticecheck`). Both generators `--check` clean.

## Pins moved

- `test/qschemetrip.hash`, re-pinned with a log entry (serializeSnapshot/deserializeSnapshot changed, scheme bumped in the same diff).

## Found on the way (not this change)

- **Same class, narrower window.** The `--quality-baseline` sidecar stores `dead` records stamped only with `head <sha>`, so a baseline pinned by one build and read by another at the same HEAD compares two resolvers' dead sets. The fix is a policy choice (fall back or disclose), so it is filed separately, not folded in.
- **Stale object with no error.** During the experiment, restoring `graph.h` landed in the same second as `main.cpp.o`'s write. Apple's make 3.81 compares whole seconds, so a "successful" incremental rebuild kept a guard-off `main.cpp.o` (`--callers=launder` count 1, not 0). Deleting the object fixed it. This is the stale-object class `CLAUDE.md` documents, reached without any concurrent build.

## Summary by CodeRabbit

- **New Features**
  - Quality-analysis caches now automatically separate results when the source code, queries, repository state, parser, or configuration changes.
  - Cache data is validated against the build that produced it, preventing stale or incompatible results from being reused.

- **Bug Fixes**
  - Improved protection against incorrect quality-analysis results caused by differing producer builds.

- **Tests**
  - Added coverage for cache reuse, invalidation, source identity changes, and corrupted data handling.
  - Expanded regression coverage to include null-byte validation.

- **Documentation**
  - Updated gate-suite counts and documented the revised cache behavior.


joyful-ii-V-I:
## CodeRabbit round: disposition

One review on `ae7e1e01`, one actionable finding. It has been checked against the code and is fixed.

| Finding | Verdict | Commit |
| --- | --- | --- |
| `CHANGELOG.md:83`, Minor: the entry said the cache key and blob header "carry" the SHA-256 source identity | **Valid, fixed.** `qsnapExclHex` / `qbodyExclHex` fold the full 64-hex identity into the material `exclConfigHex` hashes (`fnv1a64`) for the filename key. `serializeSnapshot` writes `producerIdentityHash()`, the `fnv1a64` of that identity, which `deserializeSnapshot` must match. The entry now states both encodings. Wording only, no code change. | `8fb4dcd9` |

**New head: `8fb4dcd9`** (was `ae7e1e01`). No other commits in this round. The full-suite and ASan evidence in the PR body stands, because the change touches only `CHANGELOG.md`.

**Landing:** this PR goes to main through an integration train. The train resolves the `kQSnapCacheScheme` number, re-derives the `test/qschemetrip.hash` pin, and reconciles the gate loop and count and the `test/cachefuzzcheck.sh` header offsets on the merged tree. The branch is deliberately not re-merged with main here, which is why GitHub shows a conflict.

@coderabbitai resolve

joyful-ii-V-I:

PR #243, opened 2026-09-16
fix(ingest): a C++ member call with explicit template arguments minted no call reference

joyful-ii-V-I:
## Defect

A C++ member call whose callee carries explicit template arguments minted **no call reference at all**:
`r.get<K>( 1 )`, `p->get<K>( 1 )`, `x.template get<K>()`, `this->template get<K>()`. The callee parses as
`field_expression field: (template_method …)` — inside a `dependent_name` behind the `template` keyword — and no C++
reference pattern bound either shape. The drop happened at extraction, before `ambiguous=`/`unresolved=` could count
it.

Measured 2026-09-16 on main `3bf884e2`, a four-line repro:

```cpp
struct R { template<class E> E get( int n ) { return E( n ); } int plain( int n ) { return n; } };
enum class K : unsigned char { A, B };
int useTemplate( R& r ) { return int( r.get<K>( 1 ) ); }
int usePlain( R& r ) { return r.plain( 1 ); }
```

`--callers=get --no-cache` → `count="0"`; `--callers=plain` → `count="1"`. Every verb reading the call graph
under-counted these calls (`--callers`, `--uses`, `--impact`, `--safe-delete`), and `--quality-delta` could call a
method reached only this way `kind="dead-code"`.

Two more wrong answers in the same family, on the **qualified dependent** form (which did extract):
- `X::template make<int>()` minted a reference **named** `template make`, which resolves to nothing.
- `X::template Rebind<int>::f()` keyed its qualifier as `template Rebind`, so the canonical tier missed and the call
  split over a same-named definition in another scope.

Free `f<T>( x )` and qualified `ns::f<T>( x )` (the `std::get<0>( t )` shape) were already bound; the gate keeps them
as controls.

## Change

- `queries/cpp/tags.scm` — one alternation pattern binds `field: (template_method name: (field_identifier) @name)` and
  `field: (dependent_name (template_method …))`. One node matches one branch, so each call is minted once.
- `src/ingest_binds.h` — `receiverOf` reads `calleeAccessParent`, which steps over `template_method` and
  `dependent_name` to the `field_expression`. Without it the new references classified as BARE calls, and the
  enclosing-class rule bound `other.pick<int>()` to the caller's own same-named method.
- `src/ingest_names.h` — `skipTemplateDisambiguator` (the keyword plus whitespace/comments), applied to the qualifier
  inside `immediateScope` and to the name in `cppResplitRefName`. `cppResplitRefName` is the H4 re-split block moved
[...]

PR #256, opened 2026-09-17
fix(ingest): a class template's out-of-line member keyed a second identity with the template arguments in its scope

joyful-ii-V-I:
## The defect

A C++ member defined out of line on a class **template** kept the template-argument list in its scope:

```cpp
template <class T> struct Box { void grow(); };
template <class T> void Box<T>::grow() {}
void use( Box<int>& b ) { b.grow(); }
```

This produced `<s t="method" n="grow" sc="Box&lt;T&gt;">` next to the declaration's `sc="Box"`, so one member had two identities. `--callers=Box::grow` resolved to the declaration and answered `count="0"`, and `--impact` and `--uses` missed the call the same way.

Other symptoms of the same raw text:
- A multi-line argument list put its line break into the `--pin-census` id.
- A list holding `::` was cut inside it: `template<> void Slot<std::string>::clear()` was scoped `string>`.
- `Factory<int>::make()` qualified as `Factory<int>`, which keys nothing, so the call split onto an unrelated `Decoy::make`.

## The change (third revision, `8464f3be`)

Revision history:
- **`d42f3639`** stripped the arguments from every template scope. An independent review measured 58 precise edges becoming splits and a lost delegation (APSInt.h:371). The owner decided to split the two halves.
- **`a329e0ff`** did that, but its template-family fallback ignored **inherited** members. A re-review found llvm Casting.h:548 `isa` pinned precisely to one rare specialization, and 9 `list_storage` calls pinned to `<DataType, bool>`.
- **`8464f3be`** fixes that.

**Scopes (ingest):**
- **A primary template's out-of-line member keys the bare template name** when its template-id names exactly the parameters its own `template <…>` introduces (`src/ingest_names.h` `templateIdNamesItsParameters`).
- **A specialization (explicit or partial) keeps its canonical template-id.** `canonicalTemplateIdText` drops whitespace and comments except between identifier characters and writes `, ` after a comma.
- **References keep the template-id they write,** 3+-segment spellings included (`cppRefQualifierText`).
- **Specialization headers' base clauses are now read** (new in `8464f3be`). `queries/cpp/tags.scm` captures `class_specifier name: (template_type)` as `@definition.specialization`, and `captureSpecializationHeader` emits only its base clause as inherit refs, with the canonical template-id as the derived name in `qualifier` (the Rust impl pass's convention). No symbol is minted and the Lego view is unchanged; `--uses=Base` now lists the site.

**Canonical tier (`src/resolve.h` `appendCanonicalCandidates`):**
- **Exact template-id match:** that specialization, precise (`Traits<int>::encode( 1 )`, the APSInt → APInt delegation).
- **No exact match:**
  - If the id names an **existing** specialization (it has members, or its header has a base clause) that doesn't define the name, the answer is what that specialization **inherits**, or no answer.
  - Otherwise, the primary's own **or inherited** member (`Narrower::methodOnTypeOrBases`) plus every specialization's own or inherited member. Each member reached through a base is widened to that base template's specializations (`ImutContainerInfo<T>` → `ImutProfileInfo` and its specializations). More than one is a disclosed split.
  - With **nothing visible from the primary,** only a split of **2+ specializations** answers, e.g. a traits template whose primary defines nothing (`MappingTraits`, `DenseMapInfo`). A lone specialization beside an invisible primary (`list_storage`, whose primary members extract under `cl`) gets no answer.
- **No answer leaves the bare-name ladder unchanged.** A family answer never reaches a same-named definition outside the template.
- `graph.h` stays line-neutral above `rankGraphTeleport` (3404).

**Locality (`resolve.h` `localityRank`):** for an **unqualified** bare or `this->` call, a candidate declared in the caller's own scope beats one nested inside it (`Outer::start` over `Outer::Inner::start`). Explicit receivers are excluded.

## Measured — `--pin-census --no-cache`, frozen corpora, **id join**

**Join method.** Sites are keyed on (caller symbol id `#N`, callee, line), and census rows split by a multi-line id are rejoined before parsing. Symbol ids are identical across the binaries on llvm: all 17,838 ids that parse on main name the same file, kind and line.

The earlier "54 none→precise / 6 none→external" came from a (caller file, callee, line) occurrence join over main's census, where 28 multi-line ids had broken their rows. The id join gives 12 / 0, which is what the reviewer found, and it is the method used below.

| corpus | edges | ambiguous | notes |
| --- | --- | --- | --- |
| llvm `ADT`+`Support`+`lib/Support` (590 files, 37,055 calls), main `31e788ce` | 45,768 → **45,001** | 7,247 → **7,237** | `--callers=lib/Support/APInt.cpp:getHashValue` = **5** |
| dgl `f0b7cc9`, main `b1489df4` (same resolver) | 20,829 → **20,733** | 1,891 → **1,882** | |
| this repo | identical | identical | 0 call sites changed |

**llvm, main → this head:** split→precise 37 · none→precise 12 · external→precise 12 · precise retargeted 5 · precise→split 14 · precise→none 1 · none→split 4 · external→split 9.

**The review's 74 new or retargeted precise edges (at `a329e0ff`):**
- **64 unchanged**, the ones the review read as correct.
- **isa (Casting.h:548)** is now a split: `CastInfo<…!is_simple_type…>::isPossible`, **`CastIsPossible::isPossible`** (inherited), `CastIsPossible` specializations, and `ConstStrippingForwardingCast::isPossible`.
- **8 `list_storage` calls** are splits again, as on main.
- **CommandLine.h:1760** `list_storage<DataType, StorageClass>::clear()` is back to **main's own** locality pin to `list::clear`. It is not introduced here: that primary's members extract under `cl`, and the call is qualified, so the own-scope rank doesn't apply.

**New precise edges in this revision:** 2, both correct. They are `cast` and `dyn_cast` (Casting.h:579 and :663) through `CastInfo<To, std::unique_ptr<From>>`, a specialization that only inherits `UniquePtrCast`.

**Family splits that widened (each new member checked against the base clauses in the source: `FoldingSetTrait : DefaultFoldingSetTrait<T>` at FoldingSet.h:255; `CastInfo<To, std::unique_ptr<From>> : UniquePtrCast` and `CastInfo<To, std::optional<From>> : OptionalValueCast` at Casting.h:533/539):** the `cast`/`dyn_cast`/`cast_if_present` families now include members that specializations inherit (`ConstStrippingForwardingCast`, `OptionalValueCast`, `UniquePtrCast`, `DefaultDoCastIfPossible`). `FoldingSetTrait` includes the inherited `DefaultFoldingSetTrait::Profile`. `ImutKeyValueInfo::Profile` splits over the `ImutProfileInfo` family; main split over the unrelated `ImmutableMap::Profile`.

**The 58 (main precise → first version split):** 44 precise to main's target. 14 split, as classified before: 10 main pins on the wrong class, `RHS.branched()`, and 3 `dominates` overloads.

**dgl, main → this head:**
- 11 split → precise.
- 12 precise retargeted, all from a specialization's own `Call` to the `_Sum`/`_Max`/`_Min` base it calls.
- 3 dependent template-id calls split over the primary and its specialization.
- 1 split → none (a dispatcher's call into its own next instantiation).

**Review probe `probe-family/f.cpp`:** `Info<int>::isPossible` splits over `PossibleBase::isPossible` (inherited) and the defining specialization, with no `Unrelated` (main included it). `MapInfo<Key>` and `numeric_limits<int>` match main.

## Gates

**`test/cpptmplscopecheck.sh`, 64 checks.** Corpora are generated in `mktemp`.
- **New arms in `8464f3be`:**
  - `Caster<int>::isPossible` must be the split {inherited `PossibleBase::isPossible`, defining specialization, inheriting `Caster<T*>`'s `PtrBase::isPossible`} with `mech=split`;
  - `Hasher<char>::charHash` resolves to what the specialization inherits (`CharBase`);
  - `Hasher<long>` splits;
  - `Storage<D, S>::reset()` with a declared-only primary is not a lone pin;
  - `Storage<D, bool>::reset()` is precise;
  - `MapInfo<Key>` splits (default argument);
  - `Mapping<T>::mapFields` (a primary that defines nothing) splits over its two specializations, with no stray;
  - `--uses=CharBase` lists the specialization's extends site.
- **Kept:** the non-template control and template twin (byte-identical), the primary shapes, the three specialization forms, the review's `Traits` probe, the APSInt-shaped delegation, the partial specialization's own-class call, the decoy, and the tie twins.
- **Red first:** main `31e788ce` fails 42/64; `aa69e66f` fails 6/64 (the inherited-member arms); the build before the lone-specialization rule fails 1/64. This head passes all 64.

**Targeted at this head's code:** 49 gates, 0 failed. They include cpptmplscopecheck, stdqualcheck, cppqualcheck, decltodefcheck, callformcheck, chacheck, chaconecheck, legocheck, legobundlecheck, usescheck, typerefcheck, querycheck, queryfilescancheck, locality, lpin, chainguard, rustqual, rustanc, rubyscope, declinecheck, resolverhonesty, fnptr, narrow, fieldnarrow, cuda, metal, pincensus, mapdiff, identity, scipjoin, showcasecapturecheck, forrankordercheck, readmedriftcheck, printffmtparitycheck, qextractionkeycheck, qschemetripcheck, limitstablecheck, manifestcheck, gatecountcheck and xmlwellformed.

**Full suite:** CI on this PR, per the CI coordinator.

**`--quality-delta=$(git merge-base origin/main HEAD)..HEAD`:** gating=0, acked=2 (unchanged). Minor, non-gating rows:
- `captureTagsFacts` complexity 375 → 377 and LOC 406 → 410 (one `if` plus its body);
- `captureSpecializationHeader` params = 8 (new symbol; the whole precondition is tested inside so the measured function spends one branch).

## Merge with #243 (template-call-edges), verified on a scratch merge

#243 `6b234717` merged into `8464f3be` with rerere disabled (local commit `e24f6c51`, never pushed; no recorded resolution replays). It builds and passes cpptmplscopecheck 64, cppqualcheck 96, callformcheck 199, stdqualcheck 50, nestedqualcheck 9, decltodefcheck 255, qextractionkeycheck, localitycheck and chacheck.

The recipe is unchanged:
- In `src/ingest_sidecap.h`, take #243's deletion of the H4 re-split block.
- Inside #243's `cppResplitRefName`, use `cppRefQualifierText( … )` for both qualifier lines.
- Move `cppResplitRefName` below `cppRefQualifierText` in `src/ingest_names.h`.
- In `cppRefQualifierText`, wrap the last segment in `skipTemplateDisambiguator( … )`.
- Pins (kParserVer and its mirror, `qschemetrip.hash`, the re-pin log, CHANGELOG) are for the train.

## Shared products

- `kParserVer` 96 → 100 with its mirror. **The train assigns the final number.** `d42f3639`, `aa69e66f` and this head all say 100 with different extraction, so use `--no-cache` or clear local caches when comparing.
- `test/qschemetrip.hash` is unchanged since `c4a1a887`; re-derive it at landing.
- Gate count 621 on the tree merged with `105666c1`.
- `docs/LIMITS.md` unchanged.

## Pins moved

- `test/qschemetrip.hash` (parser version).
- `test/stdqualcheck.sh` is not moved (restored to main).

## Follow-ups, not in this lane

- llvm `list_storage`'s primary members extract under `cl` (a class-template parse gap), which leaves CommandLine.h:1760 on main's pin.
- A class specialization's own class is still not a `cls` symbol; only its members and base clause are read.
- An in-class member-template declaration is extracted as `t="fn"`.
- A typed local of a nested class (`const_iterator I`) doesn't narrow.
- `--uses=Box::grow` answers `count="0"` on the non-template control too (#231).


joyful-ii-V-I:
**Disposition at head `aa69e66f`.** One push on top of `d42f3639`, adding two commits:
- `a329e0ff`: specializations keep their own identity, template-family fallback, own-scope locality tie-break;
- `aa69e66f`: the parser-version note and its re-pin log corrected to match.

No merge of main; the parser version stays 100 for the train to assign.

### CodeRabbit

1. **`src/ingest_names.h:317`, "Remove the silent qualified-scope cap": declined.**
   - The bound is not new. It is the single `kMaxQualifierHops` (32), already recorded in `docs/LIMITS.md`. `innermostQualifiedName` has always used it; this PR hoisted it so the chain walkers share one cap. `limits_build --check` still reports 213 caps.
   - Past the cap nothing becomes wrong. The unwalked links keep their written, canonicalized text, which is main's reading, so at worst precision degrades. No edge is invented or dropped.
   - A 33-link qualified class name is not written C++.
   - Removing the bound from one walker would leave the three chain walkers inconsistent. Changing the cap is a whole-cap decision, not an edit to one loop.
2. **Docstring Coverage 40% < 80%: declined.** That threshold is CodeRabbit's default, not a project rule (CONTRIBUTING §3). Every new helper carries a `//` comment stating its contract and the reason it exists.

### Independent review (relayed by the CI coordinator)

- **T1, the specialization join: fixed, per the owner's decision.**
  - A primary template's out-of-line member keys the bare name; a specialization keeps its canonical template-id; a call keeps the id it writes. When nothing is keyed by that id, the call falls back to the template's family, never to the bare-name spray.
  - llvm subset: `--callers=lib/Support/APInt.cpp:getHashValue` is 5 again. Of the 58 precise→split sites, 44 are precise again, each to main's own target. The other 14 are itemised in the body: 10 main pins on the wrong class, 1 correct pin now a split that contains it, 3 overloads now one identity.
  - Edges 45,768 → 44,935; ambiguous 7,247 → 7,230.
  - Your `Traits` probe now matches main on `<int|bool|long>`, and is precise where main split: `Traits< int >` and `ns::Info<char>`.
- **T2, "no call site can reach a `Box<int>` identity": removed** from the code comment, the CHANGELOG and the body. The claim is no longer made, and the design now depends on the opposite.
- **T3, nested-class locality tie: fixed for bare and `this->` calls** (`resolve.h` `localityRank`, gate §6 twins).
  - On main, the non-template twin already split and the template twin pinned the nested class's member.
  - The four llvm `IntervalMap::operator=`/`overlaps` sites are explicit receivers (`It.start()`, `It.stop()`, `I.start()`, `RHS.branched()`). `It` and `I` are `const_iterator`s, so 3 of main's 4 pins were on the wrong class. They stay 2-way splits that contain the right target; the tie-break deliberately ignores explicit receivers.
- **T4, #243 interplay: verified on a scratch merge.** #243 `1171f775` merged into `a329e0ff` as local commit `ffa17084`, never pushed. It builds, and cpptmplscopecheck 55, cppqualcheck 90, callformcheck 199, stdqualcheck 50, nestedqualcheck 9 and decltodefcheck 255 all pass. The resolution recipe (the `ingest_sidecap.h` conflict and function order in `ingest_names.h`) is in the body. The scratch rerere resolutions were forgotten so they cannot replay into the train's merge.
- **T5, recorded outputs:** for the train. The parser version and its mirror, `qschemetrip.hash`, the gate count and shard weights are re-derived there.
- **T6, re-key and MAX-masking:**
  - Specializations no longer share an id with the primary, so quality-snapshot MAX-masking across them no longer arises.
  - Only a primary template's out-of-line member re-keys once, from `Box<T>` to `Box`. The CHANGELOG says so.

**Gates:** the new gate (55 checks) fails 38 on main and 17 on `d42f3639`, and passes here. 42 targeted gates pass at the head, and the merge-base quality-delta reports gating=0. The full suite is CI.

joyful-ii-V-I:
**Disposition at head `8464f3be`.** One push on top of `aa69e66f`, adding one commit. No merge of main; the parser version is still the train's to assign.

### Re-review of `aa69e66f`

**R1: the family fallback pinned a precise wrong edge when the primary inherits the member. Fixed.**
- **Specialization headers are now read.** `template <> struct X<…> : Base {}` is captured as `@definition.specialization`. It is never a symbol; only its base clause becomes inherit refs, with the canonical template-id as the derived name.
- **New family rule (`resolve.h` `appendCanonicalCandidates`), applied when no definition is keyed by the written id:**
  - If the id names an *existing* specialization that doesn't define the member, the answer is what that specialization inherits, or no answer.
  - Otherwise the answer is the primary's own **or inherited** member (`methodOnTypeOrBases`) plus every specialization's own or inherited member. A member reached through a base is widened to that base template's specializations.
  - With nothing visible from the primary, only a split of 2+ specializations answers.
  - Anything else leaves the bare-name ladder exactly as before.
- **llvm Casting.h:548 `isa`** is now `mech=split` over `CastInfo<…!is_simple_type…>::isPossible`, **`CastIsPossible::isPossible`**, its specializations, and `ConstStrippingForwardingCast::isPossible`.
- **Your probe** `Info<int>::isPossible` splits over `PossibleBase::isPossible` and the defining specialization, with no `Unrelated`.
- **A second shape the same gap produced, found here:** `template <> struct Info<char> : CharBase {}` (no members of its own) used to pin `Info<char>::hash` to the primary's own `hash`. It now resolves to `CharBase::hash`.
- **Gate arms added:** Caster (CastInfo shape), Hasher (a specialization that only inherits), Storage (list_storage shape), MapInfo (default argument), Mapping (a primary that defines nothing), and the `--uses` extends site. `cpptmplscopecheck` has 64 checks; `aa69e66f` fails 6 and main fails 42.

**R2: `list_storage`. Fixed for 8 of 9; the ninth is main's own answer, disclosed.**
- CommandLine.h:1731/1733/1737/1761/1762/1784/1791/1793 are splits again.
- CommandLine.h:1760 (`list_storage<DataType, StorageClass>::clear()`) is back to main's own locality pin to `list::clear`. The primary's members extract under `cl`, which is a pre-existing parse gap, so the template supplies nothing visible and the ladder decides.

**R3: "54 none→precise / 6 none→external" did not reproduce. Corrected.**
- Those came from a (caller file, callee, line) occurrence join, and main's census had 28 multi-line ids that broke their rows.
- With your id join (caller symbol id, callee, line; symbol ids verified identical), I get 12 / 0, matching yours.
- **main → this head on llvm:** split→precise 37, none→precise 12, external→precise 12, retargeted 5, precise→split 14, precise→none 1, none→split 4, external→split 9.
- **Edges and APInt callers:** edges 45,768 → 45,001, ambiguous 7,247 → 7,237, APInt `getHashValue` callers 5.
- **Your 74 precise edges:** 64 unchanged; isa and 8 `list_storage` are splits; 1760 is back to main's pin.
- **New precise edges:** only 2, `cast`/`dyn_cast` through `CastInfo<To, std::unique_ptr<From>>`, which inherits `UniquePtrCast` (Casting.h:533).
- **dgl:** 20,829 → 20,733 edges, 1,891 → 1,882 ambiguous. This repo is identical to main.

**R4: process.**
- CI still has no run: the PR conflicts with main only on recorded outputs, and main is merged on the train.
- **#243 recipe re-verified on a scratch merge:** #243 `6b234717` into `8464f3be`, with rerere disabled (local `e24f6c51`, never pushed; 0 replays). It builds, and cpptmplscopecheck 64, cppqualcheck 96, callformcheck 199, stdqualcheck 50, nestedqualcheck 9, decltodefcheck 255, qextractionkeycheck, localitycheck and chacheck pass.

### Checks

- **Targeted gates:** 49 pass at this head's code, including chacheck, chaconecheck, legocheck, legobundlecheck, usescheck, typerefcheck, querycheck, queryfilescancheck and showcasecapturecheck.
- **Merge-base quality-delta:** gating=0 (acked=2, unchanged). Minor, non-gating rows: `captureTagsFacts` complexity +2 and LOC +4; `captureSpecializationHeader` has 8 parameters, because it takes the whole precondition so the measured function spends one branch.

joyful-ii-V-I:
This PR is part of integration train 2b, #275, and will show as merged when that train lands.

PR #257, opened 2026-09-17
fix(resolve): a member field typed std::string narrowed to an in-repo class named string

joyful-ii-V-I:
**Stacked on #248 and #254; review only the last commit (50129f8c).** It lands in the coordinator's integration train after #248 and #254, which is where main gets merged in, kParserVer gets its final number and the qschemetrip hash is re-derived.

## The defect

`struct Record { std::string name_; int nameLength() { return name_.size(); } };` pinned `name_.size()` to an in-repo `string::size`: census `mech=receiver-rule`, one precise edge, no `amb=`. The field capture keeps a qualified type's final segment. All three readers of that record took `string` for any in-repo class with that name:
- Rule 2b's field-type table.
- The HAS-A edges (`Record → string`).
- The member index (`--uses=string.len` pinned `name_.len` and `this->name_.len`).

The previous commit on this stack refused a `std::` written type for locals and parameters; fields are the third place a written type is read.

## The change

- **Capture:** a field's compose record carries the namespace its type was written in as its `qualifier`, the same (name, immediate qualifier) pair a call carries. kParserVer 98 → 99 as committed, plus the quality.h mirror (the final number is assigned in the train); the cache format is unchanged.
- **Readers:** `resolve.h fieldTypeWrittenInStd` is the one predicate. `buildFieldNarrowTables` records `""` for such a field, and the HAS-A loop draws no edge. The member index reads the same table.
- **Other qualifiers:** they narrow as before. `store::Text body_; body_.size()` still reaches `Text::size`.
- **Why `""` and not a skip:** rocksdb's two `StringSource` classes share the key `StringSource#contents_` (`std::string` in testutil.h, `Slice&` in log_test.cc). Skipping the std record un-tombstones that key, and four testutil.h `contents_.size()` calls then pin `Slice::size`.

**Audit.** Every reader of `Reference::qualifier` excludes compose refs:
- the resolve loop and census (`isResolvableCallReference`);
- the model.h `isCompose` filters;
- the inherit-only loops;
- the Python/Rust/Elixir-only readers.

The other `ing.references` loops that don't filter `isCompose` never read `qualifier`, and the ref sort key excludes it.

## Measured

`--pin-census --no-cache` and the default map, cfa5792f vs this commit:
- **Corpora:** rocksdb, a private C++ corpus and `src/` (frozen copy) are **byte-identical**, and so is rocksdb's `--metrics`.
- **Why:** none has an in-repo class that shares a `std::` field type's name and defines the called member. Fields written `std::X`: 1,258 on rocksdb, 1,987 on `src/`, 143 on the private corpus (counted with `--match`).
- **Probe:** rocksdb plus one in-repo `shim::string` defining `size`, `empty`, `data`, `c_str`, `clear`, `append` and `len`.
  - Pins to `shim::string`: 384 before → 7 after. The 7 are gtest `const ::string&` parameters (global namespace, not std).
  - 357 sites move: 279 land on the unmodified rocksdb census row, and 78 are `c_str` calls the probe's own second `c_str` makes ambiguous.
  - `--metrics`: 455 class rows lose a composed type.

## Gates

- **`test/fieldnarrowcheck.sh` arm q:** 44 PASS.
  - Red on cfa5792f: q1, q3, and both q5 rows.
  - Red on a refuse-every-qualifier variant: q2, q4, q6 and all four q7 rows.
  - Red on the skip-at-capture variant: exactly the four q7 rows.
- **Targeted gates, all green:** narrowcheck, qextractionkeycheck, qschemetripcheck (re-pinned with a dated note), composelangcheck, fieldusescheck, stdqualcheck, cacheidentitycheck, chacheck, chaconecheck, resolverhonestycheck, localitycheck, clsrecvcheck, shadowcheck, fnptrcheck, externalvetocheck, manifestcheck, gatecountcheck, limitstablecheck, showcasecapturecheck (graph.h is line-neutral), gateexitcheck.
- **ASan+UBSan+LSan:** clean on the fixtures and on the rocksdb probe; the sanitizer build's census is identical to the plain build's.
- **Determinism and xmllint:** clean on the map, `--around`, `--uses` and `--for`.
- **Full suite, local, one serialised run on 50129f8c:** `gates=632 pass=629 skip=2 fail=1`. The two skips are the environmental ones (argvdiffcheck, editchecknotecheck). The one failure was `versioncheck`, a true positive about the binary's stamp: it read `built_from=e20790c92+dirty` because I built before amending the commit, and the amend changed only `test/qschemetrip.hash` and `test/qschemetripcheck.sh`. After a relink, `versioncheck`, `fieldnarrowcheck` and `g1freshcheck` pass standalone, and the relinked binary's probe census and map are byte-identical to the binary the suite ran.
- **quality-delta, merge-base range:** gating=0.

## Pins moved

- `test/qschemetrip.hash`: the parser version moved, and the new pin equals the gate's printed `current=` on cfa5792f + this commit. It gets re-derived on the train's merged tree.

## Summary by CodeRabbit

- **New Features**
  - Improved call resolution using declared types for parameters, range-loop variables, and local references.
  - Added support for qualified in-repository types while preserving correct behavior for external standard-library types.

- **Bug Fixes**
  - Prevented standard-library types from incorrectly matching same-named project types.
  - Improved handling of unknown or conflicting receiver types, preserving honest ambiguity instead of applying incorrect locality hints.
  - Corrected declaration position tracking for more reliable analysis.

- **Documentation**
  - Updated changelog and evaluation notes with the revised resolution behavior.


joyful-ii-V-I:
This lands through integration train 2, #265, with parser version 99. Resolutions are on that branch, so nothing more is needed here. This PR will show as merged when #265 merges.

PR #259, opened 2026-09-17
fix(quality): a --quality-baseline pin another build computed was the floor, dead set and all

joyful-ii-V-I:
**Stacked on #255 (`lane/qsnap-producer-identity`). Review only `5db5f9b9` (gate), `c0c1125f` (fix) and `3189ccb6` (a one-line follow-up the full suite caught). `2a31bc7b` is the merge of #255 at `ae7e1e01`, which carries `origin/main` `31e788ce`.** This branch lands in the same integration train as #255, right after it. The coordinator resolves the shared products (gate loop, count, shard weights, CHANGELOG order, `COMMANDS.md`/`printf_parity` pins) once, on the merged tree.

## The defect

`--quality-baseline` writes `.ripwire_quality_baseline`, and its `dead <key>` records depend on **call resolution** (`isDeadCandidate` reads `g.inEdges`). The sidecar was stamped only with `head <sha>`, and `selectBaseline` compared only that. So at one HEAD, a floor pinned by one build was compared against another build's working-tree dead set. It's the qsnap cache defect again, this time in a file the user writes on purpose. The window is ordinary: pin with the installed `ripwire`, then run the delta with `./build/ripwire` (or after an upgrade) before the next commit.

Measured with two real builds of `d8c225a2`: one as built, one with `graph.h`'s `keepStdQualifiedCandidates` forced to `return true`. Fixture: `src/pool.cpp` `struct Pool { int launder( int x ) {…loop…} };` and `src/use.cpp` calling `*std::launder( &v )`.

| direction | pinned by | delta by | before | same-build answer |
| phantom (tree untouched) | unguarded | guarded | `sidecar`, gating dead-code on `Pool::launder`, **exit 2** | exit 0 |
| hidden (the `std::launder` call deleted) | guarded | unguarded | `sidecar`, **exit 0** | exit 2 (the real regression) |

## The fix

- **Format v6**: the sidecar carries `producer <64 hex>`, the build's source identity (`producerIdentity()`, from the qsnap lane).
- **`BaselineSource::Foreign`**: a pin at the current HEAD whose producer is missing or names another build is not the floor. `--quality-delta` falls back to the HEAD tree this build computes and reports `baseline="git-HEAD (foreign sidecar ignored)"`. The CLI prints one stderr line and the full legend defines the marker. The pin is **never unlinked** on either arm, since the build that wrote it can still use it. A root with no git exits 1 naming the foreign pin. A pin at another sha is still stale first and keeps its self-heal.
- **Owner's call, recorded**: fall back and keep the file, instead of keeping the pin and demoting dead-code rows. Demoting rows can't surface a regression whose row never appears, and another build can compute any kind differently.
- **Old sidecars**: all existing sidecars are v5. Only a pre-stamp build can write one, so the version rule refuses them (owner's call: bump and refuse, per the v4/v5 precedent). That refusal used to be reported as absent: `baseline="git-HEAD"` plus a stderr line saying "no .ripwire_quality_baseline" right under the refusal naming the file (measured on `d8c225a2`). It now reads `git-HEAD (sidecar unreadable)`. In a root with no git, both arms' fatal messages now say "unreadable" instead of "no <file>". Older binaries refuse v6 the same way.
- `--help=all` states the rule and names the marker. `test/qrevtokencheck.sh`'s hand-written sidecars move to the v6 header so its hostile head stamps still reach the head-stamp path.

On the fixed tree, the same two builds give the same-build answer in all four cross-build runs: phantom exit 0, hidden exit 2 with the real row, both marked foreign, sidecar still on disk.

## Gates

- `test/qbaselineproducercheck.sh` (new): **26/26 PASS**. Against `d8c225a2`'s binary **12 FAIL**: format, both flipped-producer arms, unstamped, v5, no-git, MCP, legend, help. The controls pass there. Matched pairs over a real pin: the controls, with the producer kept, must change the answer and do; the flipped-producer forgeries must give the no-sidecar answer and leave the file byte-identical.
- Targeted set (34 gates, including printffmtparity, docscommands, donelegend, legendcoverage, compactlegend, mcpattrparity, mcpclidiff, sidecarsymlink, qualitystale, qualitysignal, qoriginoracle, qsnapproducer, gateexit, manifest, gatecount, limitstable): all green after one fix. `gateexitcheck` (G2) caught a one-line `&& ok || no` in the new gate.
- ASan (built before the coordinator's sanitizer rule): the new gate, `qrevtokencheck` (25) and `qualitystalecheck` (37) pass with zero sanitizer logs, on the pre-merge and merged trees. Positive control for log capture: a deliberate heap overflow built by the same clang wrote its report to the same place.
- Determinism and well-formedness on the foreign marker: two runs byte-identical and well-formed for full XML, compact XML and `--json`.
- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: **gating="0"**. Five minor rows: `readBaseline` ccx 35→36, `emitQualityDeltaLegend` 30→31, `kHelpHead` LOC +2 (this lane); `deserializeSnapshot` ccx +2, LOC +5 (qsnap lane).
- Full suite, one local run on `2a31bc7b` (the merged tree, before `3189ccb6`): `gates=635 pass=632 skip=2 fail=1`. The skips are the environmental two (`argvdiffcheck`, `editchecknotecheck`); `g1freshcheck` ran against a fresh ASan tree and passed. The one failure, `adaptivecheck` arm (c), was **caused by this lane** and was deterministic (red twice standalone). `c0c1125f`'s `mcpNoBaselineMessage` declared `const std::string file( … )`, which extraction indexed as a symbol named `file`, so the gate's broad live-tree query `--for="file"` routed name-exact and kept 1 of 40. `3189ccb6` renames the local and uses `=` initialization. The probe routes `subtoken+body:broad` again (kept 40 of 40, no cliff), and `adaptivecheck`, this lane's gate, `mcpflagshipcheck`, `mcpclidiffcheck`, `mcpcontractcheck`, `mcpframehonestycheck` and `gateexitcheck` pass. **No local full suite has run on `3189ccb6`**: per the coordinator's plan for a loaded shared machine, CI's legs are that run.

## Pins moved

- `test/printf_parity.manifest`: `help_all` only (`UPDATE_GOLDEN_EXPECT=help_all` matched). Still green on the merged tree with no re-record.
- `docs/COMMANDS.md`: regenerated (the `--quality-baseline` and `--quality-delta` rows).
- Gate count: generated, 621 after merging main (`diagnoticecheck` + this lane's gate).

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented quality snapshots and cached results from being reused across builds with different source content.
  - Improved handling of incompatible, unreadable, or foreign quality-baseline sidecars, with clearer fallback behavior and diagnostics.
  - Updated baseline metadata validation to avoid accepting artifacts from incompatible builds.

- **Documentation**
  - Updated command help and documentation to explain build-specific baseline handling and fallback behavior.
  - Updated documented gate counts from 619 to 621.

- **Tests**
  - Added coverage for cross-build snapshot and baseline compatibility scenarios.


joyful-ii-V-I:

PR #261, opened 2026-09-17
test(gates): three degrade-alert arms asserted nothing on the plain build; pargates now fails that skip

joyful-ii-V-I:
## The defect

A gate that asserts a `DEGRADED_PATH_ALERT` has to know whether the binary can print one, because Release compiles the alert out. Three gates found out by running `--rank-by=churn --since=notadate` and looking for an alert:

- `test/churnjoincheck.sh` G2 (the NFD join loss)
- `test/preproccondcheck.sh` (the 600-deep guard stack)
- `test/w3fixlegendcheck.sh` arm 6 (the `--since` alert's wording)

e7688981 (M8) made an unresolvable `--since` a refusal that exits 1 before any degrade path runs. Since then that probe prints no alert on any build, so all three printed their NDEBUG skip on the **plain** build too. That is the leg CI keeps to prove degrade paths (CONTRIBUTING.md §5). Measured on a55b118e with a `dev` binary: one SKIP each, zero failures, every gate green.

The suite could not see it. An arm-level skip inside a passing gate counts as a pass, so `skip=` never moved.

## The change

**The three arms** read the build type `--version` names, the way kotlincheck §12 and estchargecheck already do. Release, RelWithDebInfo and MinSizeRel skip with the flavour named; every other flavour asserts.

- **preproccondcheck** asserts the exact depth-bound alert text.
- **churnjoincheck G2** asserts `noteGitJoinDegradeOnce`'s alert on the line right after the NFD disclosure. Any `[math degraded]` line used to satisfy it, so an unrelated degrade on the same run could stand in for the join loss.
- **w3fixlegendcheck arm 6** lost its subject: M8 deleted the alert whose wording it pinned. It now pins the refusal that replaced it: exit 1, no document, one stderr line, and on a non-NDEBUG build no alert. An absent alert only proves something on a binary that prints alerts, so the arm first checks that an undecodable `--scip` index does raise one.

**`test/pargates.py`** reads the binary's build type once. On any type other than the three NDEBUG ones, every skip marker row (`_SKIP_RE`, never narration) whose reason says NDEBUG or "compiled out" becomes a FAIL row. The row names the build type and quotes the skip. This covers both whole-gate and arm-level skips. A skip for any other reason is untouched. A binary whose `--version` names no build type disarms the check, and the summary says so.

## Evidence

**Red first**, on the plain build:

| gate | before (pass/fail/skip) | after |
|---|---|---|
| churnjoincheck | 71 / 0 / 1 | 72 / 0 / 0 |
| preproccondcheck | 50 / 0 / 1 | 51 / 0 / 0 |
| w3fixlegendcheck | 71 / 0 / 1 | 74 / 0 / 0 |

**Each arm can fail.** Checked with wrapper binaries; only the targeted arm moved in every case.
- Alerts stripped from stderr: each arm FAILs.
- `--version` rewritten to `Release`: each arm SKIPs.
- An alert appended to the `--since` refusal: arm 6 FAILs.
- The join alert swapped for an unrelated one: G2 FAILs.

**The harness rule** is gated by `test/skipclassifycheck.sh` arm (I): two fake binaries that differ only in the build type they name.
- On the old harness, 4 of its 6 rows are red. It covers both skip shapes on `dev` and on `Release`, a control whose skip blames something else, the FAILURES report, and the DISARMED disclosure.
- Run against HEAD's three unfixed gates in a scratch worktree, the new harness fails all three and quotes each gate's own skip row. The fixed gates pass under a `dev` binary and under a `Release`-reporting wrapper.
- A search of `test/*.sh`, `test/lib` and the python helpers found no other gate that can print such a row on a non-NDEBUG build. On current main, the five gates #258 re-points print such a row only when the build type names NDEBUG, or, in qoriginoraclecheck's case, after an arm that has already failed.

**Gates on the merged tree** (a clean rebuild at the lane head, which includes main 105666c1): churnjoincheck, preproccondcheck, w3fixlegendcheck, skipclassifycheck, pargatescheck, gateexitcheck, manifestcheck and gatecountcheck all exit 0, and `docs/limits_build.py --check` exits 0. gateexitcheck's opt-in full sweep (E) was not run; no arm here changes how a gate records failures. `--quality-delta` shows 0 gating rows: `run()` stays at its old complexity because the rule lives in `fail_ndebug_skips()`.

## Not in this PR

- The same dead probe in five more gates is #258, which touches none of these files.
- **Shared products: none.** A CHANGELOG entry is written and held for the next train's single CHANGELOG reconciliation instead of riding here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of invalid `--since` values, ensuring they fail clearly without producing misleading fallback or degradation messages.
  * Build-specific alert checks now correctly account for configurations where diagnostic alerts are unavailable.

* **Tests**
  * Expanded test coverage for build-type detection, skipped diagnostics, failure reporting, and alert behavior across development and release builds.
  * Added checks to distinguish valid skips from unexpectedly disabled diagnostics.


joyful-ii-V-I:
**CodeRabbit disposition (the one round): no fix push. The single finding is declined here, and the fix belongs in the CMake code that writes the token.**

**`test/pargates.py:840`** — on a multi-config generator (Xcode, Ninja Multi-Config), `CMAKE_BUILD_TYPE` is empty, so `--version` names `dev` for a `--config Release` binary that defines NDEBUG.

Verified: for that setup the finding is real. It is not a defect this PR introduces, and the rule fails in the safe direction there:

- **No build of this project uses a multi-config generator.** The dev tree is `cmake -S . -B build`, CI configures with `-DCMAKE_BUILD_TYPE=Release` or nothing (`.github/workflows/ci.yml:258`, `:457`), and `CMakeLists.txt` has no `CMAKE_CONFIGURATION_TYPES` or `$<CONFIG>` handling.
- **The failure is loud, never silent.** On such a binary the rule prints a FAIL row naming build type `dev` and quoting the skip. It cannot turn a missing assertion into a pass, which is the direction this PR exists to close.
- **The token is shared.** kotlincheck §12, estchargecheck, localscountcheck and qualitystalecheck already decide from the same `--version` token, so a multi-config Release binary already fails those. A pargates-only workaround would leave the harness and those gates disagreeing about one binary.

The fix belongs where the token is written: `CMakeLists.txt:294` (`_ripwire_build_type_str`) should name the active configuration on a multi-config generator. That change needs its own gate, so it goes in a separate follow-up rather than this test-only lane.

PR #262, opened 2026-09-17
ci(nightly): ThreadSanitizer runs against main once a day and reports to one tracking issue

joyful-ii-V-I:
ThreadSanitizer had a build mode (`-DRIPWIRE_TSAN=ON`) and one gate written for it (`qsnapprefetchcheck` arm (e)), but no job ran either. A data race could reach a tag unless someone happened to build TSan locally in between. This adds `.github/workflows/nightly.yml`, which runs it against `main` at 07:17 UTC, on `workflow_dispatch`, and on any pull request that edits the workflow (this one included).

**Why not per PR.** TSan builds already run often on contributors' and maintainers' own machines, and every PR already waits on the macOS runners. A TSan leg per push would load CI more than it adds coverage. Once a day against `main` covers what a local run cannot: someone ran it on what is about to be tagged.

**What runs**
- `changed`: a scheduled run skips the heavy job when `main`'s HEAD is the `head_sha` of the last successful scheduled run and no `nightly-failure` issue is open. If the API read fails, the checks run.
- `tsan` (ubuntu-24.04, clang, its own tree): before any gate, the job proves two things. Every object of the `ripwire` target references `__tsan_func_entry` and the binary carries `__tsan_init`. And the report check can fail: a planted race must fail the wrapper with a data-race report (within three attempts), while its atomic twin must pass.
- Ten gates then run, each through a wrapper that fails on a non-zero exit OR on any per-gate `log_path` report file: `qsnapprefetchcheck`, `mcpeditracecheck`, `mcpwatchercheck`, `mcpstalecheck`, `qackconcurrencycheck`, `cacheisolationcheck`, `det-gate.sh`, `grepfastcheck`, `docdriftcheck`, `mergescoutcheck`.
- Why the wrapper: a gate's own verdict cannot see a report from a process whose stderr it discards, and with `log_path` set, `qsnapprefetchcheck`'s own stderr grep reads nothing. Every gate step runs even after an earlier gate fails, so one night reports all red gates. The job comment gives the reason for each gate.
- `report-failure` / `report-green`: the only jobs that write, and only to issues.
  - A failing run on `main` opens "Nightly checks failing on main" or comments on the open one. The text gives the failing jobs and steps, the sha, the run URL and the head of the first report.
  - The next green scheduled run comments "green again at <sha>" and closes the issue.
- A placeholder comment marks where the Windows full-suite job lands (D3 of #44).

**Security.** The top-level token is `contents: read`. Each job widens only what it needs: `actions: read` and `issues: read` for `changed`, `issues: write` for the two report jobs. No secret beyond the automatic token. Values reach the shell through `env:`, and the report excerpt travels base64 into a body file. Actions are pinned the way `ci.yml` pins them.

**Gate.** `test/g1configcheck.sh` gains six rows read off the workflow's structure by a PyYAML-free scanner:
- the schedule, dispatch and self-filtered PR triggers
- the skip probe, and every heavy job gated on it
- the TSan configure, `halt_on_error`, the positive control, and every gate going through the wrapper
- the permission scoping
- the report-job conditions
- no secret but `GITHUB_TOKEN`

Each row has a mutated copy that turns exactly that row red. No new gate file, so the gate count is unchanged.

**Measured locally** (Apple clang 21, TSan build of 105666c1):
- All ten gates passed through the wrapper with no report file, in 4 s to 404 s each.
- The wrapper went red on the planted race, and on a race whose exit code the command swallowed.
- The control step, extracted from this YAML, went red against a runtime silenced with `report_bugs=0`.
- TSan missed the first version of the planted race (one store per loop after folding) 1 run in 300. The control now uses a `volatile` counter (0 misses in 600) and allows three attempts.
- actionlint: 0 findings.

**Not verified here:** the Linux job itself. Its first run is this PR's.

## Summary by CodeRabbit

- **New Features**
  - Added daily nightly quality checks using ThreadSanitizer to detect concurrency and race-condition issues.
  - Added validation for deterministic behavior across ten thread-related checks.
  - Added automatic tracking of nightly failures through labeled issues, with follow-up updates when checks recover.

- **Documentation**
  - Added contributor guidance for nightly checks and local reproduction steps.
  - Documented the workflow, reporting behavior, and current Windows coverage placeholder.

- **Tests**
  - Added structural validation to ensure the nightly workflow remains correctly configured.


joyful-ii-V-I:
Commits after the head CodeRabbit reviewed:
- `26601c57`: comment correction in `nightly.yml`. The plain-`int` race control missed 8 in 900 runs, not 1 in 300. The workflow uses the `volatile` control, 0 misses in 600, with up to three attempts.
- `6ad005db`: the PR's own tsan run stopped at `sysctl vm.mmap_rnd_bits` with `permission denied`, because the key is root-readable on the hosted runner. The read now uses `sudo`, like the write. actionlint is clean and `g1configcheck` passes.

joyful-ii-V-I:
`a20f2947` answers both CodeRabbit findings:
- `nightly.yml:154`: `actions/checkout@v4` now sets `persist-credentials: false`. No later step uses git with the token, so gate scripts can't read it.
- `CONTRIBUTING.md:531`: the skip condition now says a scheduled run checks again while a `nightly-failure` issue is open.

actionlint is clean and `g1configcheck` passes.

PR #266, opened 2026-09-17
fix(slice): a definition deeper than 256 levels walks on its own 64 MB stack, so the 2,048-level guard holds under ASan

joyful-ii-V-I:
Follow-up to #252, which merged at `1d93429b` before this commit was pushed.

**What was wrong.** #252's 2,048-level `--slice` guard was sized for plain stack frames. An ASan run of `slicecheck` at `1d93429b` aborted with a stack overflow in the reaching-definitions walker, on (15a) 2,000 nested ifs and (15b) 2,040 nested `for` loops. Under ASan the per-level frame is about 10× a plain build's. Measured with `ulimit -s` bisection:
- 2,040 loops need 18.8 MB;
- 2,040 ifs need 14.0 MB;
- nested loops overflow the default 8 MB main-thread stack past about 880 levels.

Plain release builds need 1.8 MB at the guard, and CI's asan job doesn't run `slicecheck`, so `main` is not red. But the guard was not what held the stack.

**The fix (`2acfe5c9`).** A definition nested deeper than `kSliceCallerStackDepth` (256) is walked on its own thread with a 64 MB stack. Shallower definitions, which is nearly all real code, run on the calling thread as before, with no thread started. The guard stays at 2,048, so the 808-level CPython `test_traceback.py` shape is still answered.
- Margin on that thread under ASan: loops 18.6 MB (3.5×), ifs 13.8 MB (4.7×), else-if chains 8.3 MB (7.9×). A plain build needs 1.8 MB (37×).
- If the thread can't start, the slice is refused with a named reason. The CLI and MCP share one helper.
- The new constant is registered as BOUNDARY in `docs/LIMITS.md`, with `docs/TUNING.md`, the README cap count and CHANGELOG updated.

**Gates.**
- `slicecheck` passes on the plain build and under ASan (44 PASS). (15b) now runs with the caller's stack held to 1 MB (`ulimit -s 1024`): the old walk crashed there (rc 139), and the new one passes down to 256 KB.
- `mcpslicecheck`: 2,040 loops are answered with a 1 MB stack; 6,000 blocks come back as a JSON-RPC error and the server stays up.
- `sliceflowcheck`, `sliceflowsenscheck`, `slicediffcheck`, `readmedriftcheck`, `capsweepcheck`, `limitstablecheck`, `gateexitcheck`, `manifestcheck`, `versioncheck` and `limits_build.py --check` all pass.

Known, not in this PR: `--slice --since` still labels an older revision whose function nests past the guard as `unparsed_at_rev`, which is the wrong reason. Fixing it needs a new status value.

## Summary by CodeRabbit

- **Bug Fixes**
  - Deep `--slice` operations now handle definitions nested beyond 256 levels using a dedicated stack, improving reliability under constrained caller stacks.
  - Slice operations now provide consistent refusal messages when nesting is too deep, scanning fails, or the dedicated walk cannot start.
  - Seeded slice runs now correctly detect and report scan failures before continuing.

- **Documentation**
  - Updated cap inventories, limits, tuning guidance, and verification counts to reflect the latest classifications and measurements.


joyful-ii-V-I:
Closing without merging. The owner prefers the real fix: make the recursive reaching-definitions walker (`SliceRdWalker::loop → stmt → …`) iterative with its own work stack, so nesting depth doesn't touch the thread stack and the 2,048 guard only has to bound time and memory. That lands as its own PR with the same slicecheck arms: 2,000 ifs, 2,040 loops under a 1 MB caller stack, the CPython shape, and 6,000 blocks refused. `main` is unaffected in the meantime: plain builds need 1.8 MB at the guard, and CI's asan job doesn't run slicecheck.

PR #267, opened 2026-09-17
Ruby: a constant receiver pins the call instead of splitting it across every same-named method

andriytyurnikov:
## The gap

`resolve.h`'s Rule 2c already says *"the receiver token IS the type"* (`docs/EVALS.md` "Phase 4b").
It could never fire for Ruby: `ingest_binds.h::classifyReceiver` accepted a receiver node of kind
`(identifier)` only, and Ruby's class/module receiver is its own node kind — `(constant)` for
`Calc`, `(scope_resolution)` for `Outer::Engine` and `::Top`. Every `Cls.m(…)` call therefore
classified `RecvKind::None`, `receiverOf` stamped it `FieldOfVar` with an empty `recvVar`, and the
resolver fell through to the §2a name spray. Ruby's one call form that carries a type was the one
the type rule never saw:

```ruby
class Calc;  def self.add( a, b ); a + b; end; end
class Tally; def self.add( a, b ); a - b; end; end

Calc.add( 1, 2 )   # before: two <c n="add" prov="split"/> edges, amb="1"
                   # after:  one edge, to Calc::add
```

Measured on the four Ruby corpora before the change: activerecord 8.1.3 `lib` = 9,116 edges with
`ambiguous=1,496` and `declined=4,576`; a 4,683-file Rails app = 23,784 edges with `declined=12,485`.

## The rule

* The receiver's **final constant segment** is the type name (`Outer::Engine` → `Engine`, `::Top` →
  `Top`). That is the same final-segment convention Rule 2's type bindings already use (`ns::Foo` →
  `Foo`), and the one that meets `Symbol::scope`, which is the IMMEDIATE enclosing name by design.
* A `(scope_resolution)` whose `name:` child is not a `(constant)` is not a constant receiver.
  `Outer::run( 1 )` never arrives as one — tree-sitter-ruby parses it as an ordinary `(call)` with a
  `(constant)` receiver, exactly like `Outer.run( 1 )`, so both spellings narrow through the same arm.
* A Ruby **module** joins Rule 2c's class-name set. `Util.format` is a class-method call through a
  module; `@definition.module` maps to `SymKind::Other` (`ingest_crawl.h::defKind`) for every
  language rather than to a kind of its own. Restricted to Ruby, where the implication runs both
  ways: `queries/ruby/tags.scm` emits class, module, method and constant, and the other three have
  kinds of their own, so a Ruby `SymKind::Other` symbol IS a module.
* A miss never deletes an edge. A receiver naming no in-repo definition (`Time.now`), or naming one
  that does not define the callee, degrades to the unchanged honest ladder.

## Measured, `--no-cache`, before → after

| corpus | files | edges | ambiguous | declined |
| activesupport 8.1.3 `lib` | 290 | 3,868 → 3,912 | 468 → 434 | 1,022 → 985 |
| activerecord 8.1.3 `lib` | 398 | 9,116 → 9,152 | 1,496 → 1,479 | 4,576 → 4,497 |
| actionpack 8.1.3 `lib` | 157 | 3,151 → 3,140 | 390 → 364 | 943 → 923 |
| Rails app A (private) | 4,683 | 23,784 → 24,376 | 1,328 → 1,263 | 12,485 → 11,624 |
| Rails app B (private) | 2,174 | 14,859 → 15,257 | 275 → 431 | 3,264 → 3,050 |

`declined` falls on all five — those are call sites the resolver refused to guess at and now has
evidence for. Two numbers that look wrong and are not:

* **actionpack's edges fall.** A pinned call is ONE edge where a two-way split was two.
* **App B's `ambiguous` rises while its `declined` falls by 214.** A receiver naming two
  same-final-segment classes that both define the callee produces an honest split where there was
  previously no edge at all — floor (b) below, not a regression.

**No collateral movement.** The default map is byte-identical to `main` (316dbf26, built in a
scratch worktree) on this repo's `src/`, npm, a Clojure project, CPython 3.14's stdlib and this
whole repository; this repository's `--report` totals are unchanged at 2,052 files · 18,979 symbols
· 22,529 edges; `--deps` is byte-identical on activerecord (this round is call graph only, and does
not touch the constant-directive work from #57/#65/#139).

## Floors, each pinned by an arm of the gate

* **(a) Ruby feeds no class-hierarchy edges.** `ingest_relations.h::captureBases` has no Ruby arm,
  so a method inherited from a superclass does not narrow: `Child.build` stays an honest split.
  This is what holds the gem numbers down — a gem's class methods are reached up an
  `ActiveRecord::Base` hierarchy — and it is the obvious next round (it would also give Ruby the
  `--lego` inheritance view it has never had).
* **(b) Matching is by final segment**, so `Left::Shared` and `Right::Shared` both keep their `go`.
  Narrowing to one of them needs the Ruby constant index from #57, and a claim in the gate.
* **(c) Variable and chained receivers are untouched** — `c.scale` and `Calc.new.scale` are
  unchanged two-way splits.

## Gate

`test/rubyrecvnarrowcheck.sh` (43 arms) was written and committed RED first: **29 passed and 14
failed** against the pre-change binary, and every control and floor arm was live and green before
the fix. Besides the narrowing claims it pins: a miss degrading to the ladder, `Time.now` minting
nothing, determinism, warm == cold, `xmllint`, and a mutation arm that moves the receiver from
`Calc` to `Tally` and requires the edge to follow it. `test/regression.sh` lists it; the generated
gate count moves 621 → 622 through `docs/gatecount_build.py`.

Also green locally: the eight other Ruby gates, plus `clsrecv`, `narrow`, `narrowlang`,
`chainguard`, `externalveto`, `resolve`, `resolverhonesty`, `shadow`, `decline`, `fieldnarrow`,
`qualifiedresolve`, `qextractionkey`, `version`, `cachehash`. ASan (`-fno-sanitize-recover=all`,
committed LSan suppressions) is clean on activerecord and on the 4,683-file Rails app, and the gate
passes under the ASan binary. The full sequential battery is running; I will post its verdict as a
comment on this PR.

## Versions

`kParserVer` 96 → 97 — record layout is unchanged (`kCacheVersion` stays 22), but the VALUES of
`recv`/`recvVar` move, so Ruby extraction facts must be re-parsed. `quality.h`'s
`kIngestParserVerMirror` moves in the same commit and `test/qschemetrip.hash` is re-pinned. The
branch is based on 316dbf26, where `main` is also at 96 — if `main` spends 97 before this merges,
the number renumbers on the way in.

## Summary by CodeRabbit

* **New Features**
  * Improved Ruby call resolution for constant and module receivers such as `Calc.add`, `Outer::Engine.run`, and `Util.format`.
  * Calls now more accurately connect to the intended class or module method instead of producing multiple ambiguous matches.

* **Bug Fixes**
  * Preserved existing behavior for variable and chained receivers, unresolved calls, and same-name collisions.

* **Tests**
  * Added coverage for Ruby receiver resolution, cache consistency, determinism, and inheritance-related edge cases.

* **Documentation**
  * Updated gate-suite references to reflect 622 available checks.


joyful-ii-V-I:
Thanks for this, @andriytyurnikov. Treating Ruby's `(constant)` / `(scope_resolution)` receivers as a type is exactly the gap Rule 2c was waiting on. A first pass on CI before the full review:

1. **`gateexitcheck` (G1)**: the new `test/rubyrecvnarrowcheck.sh` defines `ok(){ echo "  PASS  $1"; return 0; }`. The house idiom records a failed write so a lost PASS line can't slip through silently:
   ```bash
   ok(){ echo "  PASS  $1" || { fail=1; echo "  FAIL  could not write the PASS line for: $1"; }; return 0; }
   ```
   Any existing gate (e.g. `test/nongitqmetricscheck.sh`) shows the full pattern.
2. **`showcasecapturecheck` (H)**: your `graph.h` insertions move `rankGraphTeleport` off the published capture seed (`graph.h:3406`). **You don't need to chase this.** We regenerate the capture, and assign `kParserVer` plus the `qschemetrip` re-pin, when the PR lands, because several open PRs move the same pins.

Please push only the `ok()` fix. An independent review follows, and its findings will come in one batch.

joyful-ii-V-I:
Thanks @andriytyurnikov. All three first-pass items are in at `b3270335`, and this is ready to land.

**What we checked:**
- `gateexitcheck` and `manifestcheck` are clean on the new gate.
- `rubyrecvnarrowcheck` is 43/43 on this head. On the true merge-base it goes red, 29 pass / 13 fail.
- We drew an independent sample from a local Rails checkout (a newer snapshot than your 8.1.3, on purpose), graded against the source: **20 of 20 sites better**, 0 same, 0 worse. That's a Wilson 95% interval of [83.9%, 100%].
- No perf regression, and output is byte-identical across runs.

**Two things beyond what the PR measured:**
1. **A wrong→right fix.** `activesupport/.../date_and_time/calculations.rb:247`: `Date.beginning_of_week` used to land on the caller's own `Calculations#beginning_of_week`, a silent self-referential edge from the old bare-name fallback. It now resolves to `Date`'s method.
2. **A floor worth naming.** `action_dispatch/journey/parser.rb` defines both `def self.parse` and `def parse`. `Journey::Parser.parse` can only reach the singleton, but it now gets an honest two-way split that also lists the instance method. That's a split, not a wrong pin, so it doesn't block this PR. `method` and `singleton_method` are already distinct captures in tags.scm, so a later round could separate them. We'll add a floor (d) sentence next to your (a)–(c) when this lands, so it reads as a decision rather than a surprise.

**Landing:** this goes in with our next integration train. The conflicts are all shared pins (`kParserVer`, `test/qschemetrip.hash`, the gate count, CHANGELOG), and we reconcile those on the merged tree, so please don't rebase or resolve them.

joyful-ii-V-I:

joyful-ii-V-I:
Landed via #281 (integration train 3), which merged your commits `e6a13c12`, `2688caae` and `b3270335` unchanged, so they're on main now. GitHub keeps this PR open only because the later `3579a39f` main-merge commit isn't in main. That commit has no changes of its own to carry. Credit is in the CHANGELOG, together with the floor (d) note. Thank you, @andriytyurnikov: constant receivers are a real step for Ruby call accuracy.

PR #268, opened 2026-09-17
fix(resolve): a call through an interface pointer landed on unrelated nested classes of the same name

joyful-ii-V-I:
> **Base is `main`** (`bcd3b016`, train 2 / #265 merged in once as `9ac655ed`). `kParserVer` and `test/qschemetrip.hash` are untouched: class identity reads facts ingest already extracts.

## The defect

On rocksdb, `void AssertItersEqual( Iterator* iter1, Iterator* iter2 ) { … iter1->key() … }` answered with five edges to the **nested** `Iterator` classes inside `memtable/`'s skip lists, and none of them was right.

- Rule 2 keys a receiver's type by its final class-name segment.
- A nested class keeps only that segment: `SkipList<Key, Comparator>::Iterator::key` has scope `Iterator`.
- `rocksdb::Iterator`'s methods are pure-virtual *declarations*, which the definitions-only `canonByName` never holds.
- So the only `Iterator::key` Rule 2 could find were namesakes.

#248 disclosed 79 such parameter sites. Typed **locals** (`Iterator* iter = db->NewIterator( … )`) have the same shape, and there are more than a thousand of them.

## The change

Rule 2 now reads the call through **class identity**, rebuilt from facts ingest already has. The index is `ClassIdentity` in `resolve.h`:
- byte spans give each class its enclosing class and each member its owner;
- inherit references give the class graph;
- include targets and namespace scopes supply evidence.

`IdentityNarrower` (resolve.h) owns the narrowing, called from `Narrower::rule2RecvVarType`:

1. **Keep the hits whose owner the written type can name.** This is C++ lookup outward from the caller, or a qualifier naming the enclosing class. When every hit is kept, `own` itself is returned, byte-identical.
2. **With nothing left, and one claimable class, take the shallowest ancestor that defines the method.**
3. **If the ancestry only *declares* it, resolve to its definitions in the class's real subclasses.** This is the dispatch split a virtual call through an interface is.

Steps 2–3 are claims, flagged so the ladder keeps them whole (`graph.h`, line-neutral: `rankGraphTeleport` stays at 3404). A claim is never trimmed to the same-file override.

**Guards.** Each one exists because an intermediate build made a wrong edge on a real corpus:

| guard | the wrong edge it stops | arm |
|---|---|---|
| A forward declaration (`class Iterator;`) is not a class | rocksdb has five at namespace scope; they made every bare `Iterator` look ambiguous and refused the dispatch | 31 |
| Namespace-level namesakes are told apart by include evidence: path-precise, else an include-root target read as a **path suffix**, else a shared namespace | `IntrinsicInst* II; II->getType()` landed on `sandboxir::Value::getType` (llvm); a basename was not enough, since three `Value.h` hold three `Value`s | 33 |
| A nested class visible through the caller's includes is never dropped | `using NodeSet = MachineGadgetGraph::NodeSet;` (llvm X86 LVI): the alias is invisible to the index and the fix dropped the right class for an unincluded namesake | 32 |
| A claim through a namespace qualifier needs that namespace's evidence in the class's file | a namespace-qualified third-party `map<K, V> m; m.emplace( … )` became a different vendored library's same-named container's `emplace` (private corpus) | — |
| Exclusions don't need namespace evidence; claims do | `port::Mutex` dropped its `port_win.h` platform variant | — |
| **Explain or keep:** identity replaces an answer only with one it can explain; two same-named classes defining the method at one ancestor level refuse | `MultiGetRange::Iterator mget_iter` (`using MultiGetRange = MultiGetContext::Range`) lost its edge (rocksdb) | 29, 30 |
| An inherited body is claimed where the ladder declines | cross-directory candidates, no include narrow: 397 rocksdb calls had no edge | 34 |

## Measured

Method: `--pin-census --no-cache`, comparing the lane's pre-merge base (`50129f8c`, which already carried #248/#254/#257's target-changing commits) against the lane head `6a32bd5c`. The two binaries differ by this lane's two commits alone, so the numbers are this change's own; they predate the merge with main. Every change bucket on the three external corpora was sampled and read against source on the final build.

| corpus | call sites changing target | `bound=` |
|---|---|---|
| rocksdb @ `0e2801ac3` | 2,659 | 200,036 → 201,085 |
| llvm-project @ `4d5358b1d` | 37,949 of 1,790,841 | 1,126,051 → 1,153,808 |
| private C++/ObjC++ corpus | 177 | 80,582 → 80,646 |
| this repo `src/` | 0 | unchanged |

**rocksdb** (largest buckets):
- 995 namesake splits become the real `Iterator` implementations: DBIter, ArenaWrappedDBIter, ModelIter, …, 11 for `key`.
- 638 declined calls gain a dispatch split: `Statistics::getTickerCount`, `DB::DefaultColumnFamily`.
- 397 declined calls gain their inherited body: `IOStatus io_s; io_s.ok()` → `Status::ok`.
- 276 partial splits complete: `Comparator::Compare`, 6 → 26 targets.
- 88 wrong unique pins become the interface's implementations: `env->DeleteFile` had pinned an unrelated file system.

**llvm-project:**
- 21,408 declined calls gain their inherited body: `LD->getAlign()` → `MemSDNode::getAlign`, `e->getRHS()` → `BinaryOperator::getRHS`.
- About 8,800 wrong or partial answers move to the right base: `FD->getType()` → `ValueDecl::getType`, `E->getType()` → `Expr::getType`, `Val.getBitWidth()` → `APInt::getBitWidth`.
- The remaining splits are genuine const/non-const overload pairs.

**Cost** (llvm-project, two cold runs each):

| | user time | peak RSS |
|---|---|---|
| tip | 52.4 s / 52.9 s | 2.47 / 2.37 GB |
| this branch | 50.9 s / 55.9 s | 2.40 / 2.36 GB |

Output is byte-identical run to run on llvm-project, rocksdb and the private corpus, and `xmllint` is clean.

## Sampled correctness (direction, not just counts)

**Sample.** A seeded uniform sample (seed 268): 30 of rocksdb's 2,659 retargeted call sites and 30 of llvm-project's 37,949.

**Method.** Each site was graded against source, independently of the new answer:
1. find the receiver's declared static type;
2. apply C++ name lookup and, for a virtual call through a pointer or reference, the overriders;
3. grade the old and new answers each as RIGHT / PARTIAL / WRONG / NONE (no edge).

| corpus | before: RIGHT / PARTIAL / WRONG / NONE | after: RIGHT / PARTIAL / WRONG / NONE | better / same / worse |
|---|---|---|---|
| rocksdb | 9 / 0 / 9 / 12 | 28 / 1 / 1 / 0 | 28 / 1 / 1 |
| llvm-project | 1 / 0 / 14 / 15 | 27 / 1 / 2 / 0 | 28 / 1 / 1 |

Overall, on the lane before the two fixes below: 56 of 60 better (Wilson 95%: 84–97%) and 2 of 60 worse (0.9–11%).

After the specialization fix below, the llvm worse site holds its right target inside a disclosed split (NONE → PARTIAL). The same 60 then read 57 better, 2 same and 1 worse, and the one worse is the stated floor below.

**Better, typically:**
- `Iterator* it; it->Seek(k)`: rocksdb's unrelated memtable `Iterator` classes → all 11 real overriders.
- `IOStatus s; s.ToString()`: no edge → `Status::ToString`.
- `LoadInst& LI; LI.getType()`: four unrelated file-local `getType` → `Value::getType`.
- `DIType* T; T->getTag()`: unrelated AsmPrinter classes → `DINode::getTag`.

**Worse (2): one fixed, one measured and kept as a stated floor.**

- **llvm, `SmallVectorImpl<FunctionDecl *>& v; v.push_back( FD )`: fixed** (`83c1718f`, arm 38).
  - What went wrong: the answer was the primary `SmallVectorTemplateBase::push_back`, but pointer `T` instantiates `SmallVectorTemplateBase<T, true>`. That specialization has no class symbol, and its members are scoped `SmallVectorTemplateBase<T, true>`, so the walk never saw them.
  - Fix: a defining level now adds its template's specialization-scoped definitions (`ClassIdentity::specializationDefs`). A primary template's own out-of-line `DominatorTreeBase<NodeT, IsPostDom>::verify` bodies are included.
  - The CHA-lite cone prune now skips an identity claim, as the ladder and locality already do. The cone cannot name `TBase<T, true>` and was dropping exactly these bodies.
  - This site is now `split` {primary ×2, `<T, true>`}: it holds the right target.
- **rocksdb, `BackupEngine* e; e->RestoreDBFromLatestBackup( RestoreOptions, dir, dir )`: a stated FLOOR** (arm 37).
  - `BackupEngineReadOnlyBase` pairs a pure-virtual overload with an inline compatibility overload, and the claim takes the compatibility body.
  - I built the fix (`83c1718f`): join the receiver's subclass definitions when a level declares more bodiless overloads than it has bodies.
  - A build with only that join disabled isolates its effect: exactly 7 rocksdb sites and 0 llvm.
  - Graded against source, those 7 read 2 better and 5 worse. The 5 call the compatibility overload, which was already right, and the join added an override of the other overload. Arity cannot tell the two overloads apart; only argument types could.
  - The join is removed (`e5ca33b5`) and the limit is stated.

**What the specialization fix changes** (census, pre-fix merge `9ac655ed` vs `e5ca33b5`):

| corpus | sites changing target |
|---|---|
| llvm-project | 366 |
| rocksdb | 0 |
| private corpus | 0 |
| `src/` | 0 |

The head's census is byte-identical to the "fix minus the join" build on both rocksdb and llvm.

Of llvm's 366:
- 169 had no edge;
- 21 unique and 11 split answers go from an unrelated class to the template member (`DT.verify()`: `LoopFuse`'s `FusionCandidate::verify` → `DominatorTreeBase<…>::verify`);
- 120 `push_back`/`pop_back`/`contains` splits gain their specialization or out-of-line overload.

A seeded sample of 20, graded against source, reads 20 better and 0 worse:
- 12 are now RIGHT (from 10 NONE and 2 WRONG): non-virtual template-base members such as `DominatorTreeBase::recalculate`, `GenericDomTreeUpdater::applyUpdates` and `LoopInfoBase::verify`.
- 8 are WRONG → PARTIAL: `SmallVectorImpl<T>::push_back` where `T` is trivially copyable, now a split holding the `<T, true>` body next to the primary's pair.

**With these changes, the original 60 read 57 better, 2 same and 1 worse:** the one worse is the stated rocksdb floor.

**Same quality (2):**
- **rocksdb `Env* base; base->IsInstanceOf( n )`:** the inherited `Customizable::IsInstanceOf` body replaces one subclass override. This is the stated floor: an inherited body is the static answer.
- **llvm `BinaryOperator& B; B.getOperand( 0 )`:** `DECLARE_TRANSPARENT_OPERAND_ACCESSORS` declares a hiding `getOperand` through a macro, which the index cannot see. Both answers are wrong.

**Found while sampling, out of scope:** Rule 2 binds no type for an UNQUALIFIED template-id (`SmallVectorImpl<T>& v`), because `writtenTypeOf` returns no type for `template_type`. Fixing that is an extraction change (a `kParserVer` bump) and should follow #256's template-family index. It is a separate lane.

## Gates

`test/narrowcheck.sh` arms 26–38, on generated fixtures:

| binary | result |
|---|---|
| pre-merge base `50129f8c` | **7 FAIL**: (26) (27) (28) (29) (32) (34) (35) |
| a merge build without the claim condition | **(36b) FAIL**: `prov=[final-segment]` |
| the merge `9ac655ed`, before the specialization fix | **(38) FAIL**: `[adt.h:7]` alone |
| this branch (`e5ca33b5`) | **53 PASS** (all of narrowcheck, including #254's arm 25) |

Arm 37 is a stated floor, which goes red if the behaviour moves. Arms 30, 31, 33 and (36c) are controls. Arm 33 was red on the intermediate build that dropped an aliased nested class.

Checks:
- **Full suite:**
  - on the head (binary stamped `e5ca33b5e`): `gates=635 pass=632 skip=3 fail=0`;
  - on `fe909208` (with the since-dropped join): `gates=635 pass=632 skip=3 fail=0`;
  - on the merge `9ac655ed`: `gates=635 pass=631 skip=3 fail=1` (recallevalcheck 6b, resolved below);
  - before the merge (`6a32bd5ca`): `gates=632 pass=629 skip=3 fail=0`.
  - The skips are environmental: `editchecknotecheck` and `argvdiffcheck` want a reference binary, and `g1freshcheck` has no local `asan/`.
- **quality-delta:** `gating="0"` over both `--quality-delta=bcd3b016..HEAD` (the merge-base with main) and, before the merge, `50129f8c..HEAD`.
  - The first run gated 8 rows; they were fixed in code, not acked:
    - `Narrower`'s verbosity went 409 → 800, so the identity logic moved into its own `IdentityNarrower`;
    - nine new functions over the complexity bar (and two over the params bar) were split;
    - the duplication and clone rows went by reusing `extent::inSet` with `kHeadRuleLangs`/`kExtentClassKinds`, `symbolsByFileInIdOrder`, and a flat sorted include index instead of new helpers.
  - Each refactor step was proven output-identical by `cmp` of census and map on rocksdb, llvm-project, the private corpus and `src/`.
  - The one ack is `rule2RecvVarType`'s 1 → 3 parameter contract change (the index and the chaUp graph).
  - Two non-gating rows remain: `IdentityNarrower` is a new 448-line struct, and `buildGraph`'s complexity goes up by 4 (780 → 784 on main, minor).
- **Generators:** `docs/gatecount_build.py --check` and `docs/limits_build.py --check` both clean.
- **ASan:** not run locally, per the machine rule. The stale `asan/` tree was removed, so `g1freshcheck` skips; CI's ASan legs cover it.

## `recallevalcheck` 6b now asserts the anchor's published promise (test-only, `fe909208`)

**What flipped.** On this branch arm 6b went red by 0.013 of score with no ranking change:
- It asserted a top-5 RANK for `--for="fix the virtual dispatch in test/chafix/cha.cpp"`, read off the live repo.
- The mention anchor promises a score: "lifted to within 5% of the top score" (`--help`, the `for` header). The anchored rows sit at exactly 0.95 × top.
- The rank therefore hung on whether an unanchored near-tie sits above that line:

| tree | top score | anchored (0.95 × top) | next unanchored row (`test/chaconefix/zoo.h::Creature`) |
|---|---|---|---|
| main `bcd3b016` | 18.676 | 17.743 | 17.622 (0.7% margin) |
| this branch | 18.393 | 17.473 | 17.486 |

This branch's comments and gate helpers use "virtual"/"dispatch", which lowers those terms' IDF.

**What 6b asserts now:** the anchor ran (`anchored` > 0), the fixture is in the top 50, and its best row scores ≥ 0.95 × top (0.001 tolerance for 4-decimal scores). `mentioncheck` already pins the rank behaviour on frozen fixtures.

**Evidence and scope:**
- Red with the anchor disabled (`RIPWIRE_NO_MENTION=1`): `score=absent from the top 50 … (anchored=0)`.
- Green here: `17.473 >= 0.95 x 18.3927 (anchored=4)`.
- **No default, flag or published promise changed.**

## Merged with train 2 (#254's `prov="final-segment"`)

`9ac655ed` merges `origin/main` once, with rerere off:
- **`RecvVarType`:** #254's namespace-level struct and this branch's nested one are now one struct, `{name, writtenQualified, declared}`.
- **Arm numbers:** #254's arm 25 keeps its number, and this branch's arms move to 26–35.
- **Provenance of claims:** an identity claim verified its one class before answering, so its edges carry no `prov="final-segment"` (`graph.h`, `&& !identityClaim`, line-neutral).
- **What stays disclosed:** a class-qualified step-1 narrow still carries `final-segment`, since that label means a qualified type was matched by name.
- **Arm 36:**
  - (36a) `hs::DiskHealth& d; d.fine()` claims the inherited `HealthBase::fine`;
  - (36b) that edge carries no `prov=`;
  - (36c) `Skip::Iterator& it; it.key()` keeps `prov="final-segment"`.
  - (36b) fails, as `prov=[final-segment]`, on a build without the `graph.h` condition.

## Floors, stated

- **Namespaces are evidence, not a model.** A same-named class in another namespace that the caller's file also includes stays a candidate.
- **A type alias is kept, not read through.**
- **An inherited body is the static answer,** as a class's own body always was; overriders join only for a method no ancestor defines.
- **A dispatch split is as wide as the interface's implementations:** up to 50 targets on rocksdb and 71 on the private corpus, every one disclosed by `amb=`.
- **A nested class's out-of-line member defined in a different file than the class has no known owner,** and is kept.

## Pins moved

None: no legend, byte, `printf_parity`, capture, gate-count or parser-version pin moved. `narrowcheck` gains arms 25–34.


joyful-ii-V-I:

PR #270, opened 2026-09-17
fix(uses): a member read through a std:: local or parameter pinned an in-repo namesake

joyful-ii-V-I:
## Defect

`--uses=Owner.field` pinned a read through a `std::`-typed local or parameter to an in-repo class that shares the type's final segment. Repro:

```cpp
// lib/pair.h
struct pair { int first; int second; };  struct Twin { int first; };
// app/use.cpp
int local() { std::pair<int, int> p; return p.first; }
int param( const std::pair<int, int>& q ) { return q.first; }
int mine() { pair r; return r.first; }
```

`ripwire <dir> --uses=pair.first --no-cache` answered `pinned="3"`: lines 1 and 2 were pinned to the in-repo `pair.first`, with no `owner_candidates=` and nothing disclosed. Only line 3 is that class.

`graph.h collectFieldUseSites` keeps its own `"<fromSymbol>#<var>"` → type table over the Type and ParamType binding records. That table keyed on `Binding::typeName`, the final segment, and never read `Binding::importedName`. Rule 2's flat table, its lexical lookup and the member-field table (#257) already refuse `std`. This table was the fourth reader of a written type, and #257's session filed it as the follow-up.

## Change

- **The std record tombstones.** The table records `""` for a type written in `std` (`resolve.h namesStdType`), the same tombstone Rule 2's table records. A same-named declaration of another type in that function is tombstoned too, not handed every site.
- **A tombstoned local no longer falls back to the enclosing class's same-named member.** The local lookup and the member fallback are one `receiverTypeOf` lambda now. A record that exists decides, even when it is `""`. This removes the two `if( type.empty() )` branches in the NamedVar and FieldOfVar cases.
  - Why: a local hides a member of that name, and the flat table cannot tell which declaration a site sees.
  - This also changes a non-std case. A local declared twice with different types, in a method of a class that has a member of that name, now answers the split.
- **Every other qualifier still narrows:** `store::Text t; t.len` pins.
- Ingest already carries the qualified text on both record kinds (`ingest_binds.h pushTypedBind` / `DeclType`), so there is **no kParserVer bump**, no qschemetrip change and no parity pin.

## Measured

Both comparisons are 50129f8c's binary against this change.

- **No field read moves on the three corpora.** Field reads are not covered by the pin census, so a scratch-only hook (never committed) dumped `collectFieldUseSites` for every field whose name is read through a receiver whose key has a `std::`-qualified record or two disagreeing records. That set contains every site this change can move. The dumps are byte-identical on:
  - `src/` (a frozen copy): 1,169 fields, 57,511 rows
  - rocksdb: 824 fields, 84,617 rows
  - a private C++ corpus: 1,811 fields, 393,336 rows
- **The call graph is untouched:** `--pin-census` and the default map are byte-identical on `src/` (31,318 census lines) and rocksdb (257,712).
- **Reach, on an adversarial copy of rocksdb** with `namespace shim { struct pair { int first; int second; }; }` and a second owner `Couple`:
  - The previous binary pins 110 reads to `shim::pair` (53 `.first`, 57 `.second`). All 110 are wrong, because no rocksdb code declares one.
  - This change turns all 110 into `owner_candidates=2` splits. It pins no new site and drops none.
  - Six were sampled across files, and all are `std::pair` locals or parameters: `db/builder.cc:300`, `env/env_chroot.cc:91`, `util/hash_map.h:35`, `db/version_set_test.cc:564`, `java/rocksjni/memory_util.cc:82`, `utilities/backup/backup_engine_test.cc:2789`.

## Gates

`test/fieldusescheck.sh` gets a new arm **J**, with its own fixture and hand-derived rows.

- **What the fixture covers:**
  - a `std::pair` local, a `std::pair` parameter and a `::std::pair` local
  - `std::string s; s.rep.first`, a receiver two hops deep
  - one function declaring both `std::pair v` and `pair v`
  - a `std::` local, a `std::` parameter and a `std::make_pair` local, each hiding a member `p`
  - a non-std double declaration in a method
  - controls that must keep pinning: `pair r`, the class's own member, `store::Text t` and `const store::Text& u`, against a second owner of `len`
- **Red first:** on 50129f8c, 3 rows fail. The `pair.first` rows, the `Twin.first` rows, and the arithmetic (`pinned=11`, want 2).
- **Each row can fail.** Three broken variants were built and run:
  - skipping the std record: fails line 6 (both sites) and lines 9/10/12
  - letting a tombstone fall back to the member: fails exactly lines 9/10/12 (measured before row 13 was added; 50129f8c also fails 13)
  - refusing every qualifier: fails only the `store::Text` / `Blob` controls
- **Green:** 83 PASS, on the plain build and on ASan (the ASan runs on the arm J fixture exit 0 with empty stderr).
- **Local full suite on 14bd4fcd:** `gates=632 pass=630 skip=2 fail=0 wall=897.5s` in one serialised run. The skips are the environmental editchecknotecheck and argvdiffcheck.
- **Related gates:** fieldnarrowcheck 44, narrowcheck 29, showcasecapturecheck 23, attrvocabcheck 43 and compactlegendcheck 174, all PASS. `gatecount_build --check` and `limits_build --check` are clean.
- **Quality-delta** (`50129f8c..HEAD`): `regressions="0" gating="0"`. `--edit-check=collectFieldUseSites`: signature unchanged, 0 incompatible callers.
- graph.h changes only below `rankGraphTeleport`, so the showcase capture's seed line holds.

## Since opening (head e3d04681)

- **`a81579f6`** merges `origin/main` (`bcd3b016`, train 2) once, with rerere off. It is clean with no conflicts, and no pins move. The PR diff against main is this lane's files only.
- **`e3d04681`** makes the member index fold its local types through `resolve.h recordFlatRecvType`, the helper train 2 gave Rule 2's flat table for the same tombstone rule, so one rule lives in one place.
  - Behaviour is unchanged. `--uses=<Owner.field> --limit=5000 --no-cache` is byte-identical between `a81579f6`'s binary and this one on 17 selectors, across the arm J fixture, `test/fieldusesfix`, the adversarial rocksdb copy and a `src/` copy.
- **On e3d04681:** fieldusescheck 83, fieldnarrowcheck 44, narrowcheck 36, showcasecapturecheck 23, localitycheck 15, attrvocabcheck 43 and versioncheck, all rc=0.
- **Quality-delta** (`bcd3b016..HEAD`): `regressions="0" minor="0" gating="0"`. `--edit-check=collectFieldUseSites`: unchanged, 0 incompatible. Both generator checks are clean (loop 621).

## Pins moved

None.

Lands via ripwire-67's queue: independent review, then direct on green or train 2b.


joyful-ii-V-I:
**Review disposition — head `f811baaa`**

CodeRabbit posted one actionable finding (`test/fieldusescheck.sh:262`), and it is **fixed in `f811baaa`**.

- **Why it was valid:** arm (J)'s warm-cache row compared a second `--cache` run with the cold output. That row passes when the cache was never written or never read, because the second run is then a cold parse. Arm (H) has the identical shape, so both are fixed.
- **The fix goes further than a file-exists check.** Each warm arm now also asserts that the second run READ the cache. The file must be non-empty, and the second run's `RIPWIRE_CACHE_STATS` line (the observable `artifactcheck` uses) must say `reparsed=0` and `reused=` every file.
- **Red first.** Two BIN shims were run over this binary: one turns `--cache=PATH` into `--no-cache`, the other gives each run a fresh cache path. Under both, the two new READ rows fail while the two old equality rows still pass, which reproduces the finding.
- **Green:** 85 PASS (was 83). gateexitcheck, manifestcheck and binoverridecheck pass.

No other findings. The walkthrough comment is informational.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

joyful-ii-V-I:
Commits after CodeRabbit's reviewed head, for the record:
- `a81579f6`: merges `main` `bcd3b016` once; clean, no pins moved.
- `e3d04681`: the member index's std:: tombstone now calls `resolve.h` `recordFlatRecvType` instead of a 6-line copy; `--uses` is byte-identical on 17 selectors.
- `f811baaa`: CodeRabbit's warm-cache finding; arms H/J now assert a real cache read.

All three are covered by the independent review.

PR #250, opened 2026-09-16
harden: git file-system-monitor policy, exclusive temp files, confined edit-plan reads

joyful-ii-V-I:
Hardening in three areas, each with a gate that fails on `main` and passes here, on the dev and ASan builds.

1. **git runs with the file-system monitor disabled.** Every git command ripwire runs goes through one prefix (`--no-optional-locks -c core.fsmonitor=false`, `src/gitcmd.h`). `githarden` resolves the governing git dir by asking git, and `--doctor` and `--help` state the policy.
   - Gate: `githardencheck` arms J/K/L/M.
   - J runs the git-backed verbs across several root shapes and through the MCP and `--plan-lint` entry points.
   - K uses a PATH shim to check that every git child carries the prefix.
   - L is a structural arm that refuses a git command spelled outside `rw::gitCmd()`.
   - M checks that the disclosure follows the governing repository.
2. **Temp files are created exclusively and do not follow links.** The three tmp+rename publish writers create their temp through one RAII helper (`O_CREAT|O_EXCL|O_NOFOLLOW|O_CLOEXEC`, random name, removed unless committed). File modes are unchanged, and two cache-directory writers now use the same helper.
   - Gate: `tempfilesymlinkcheck`, behavioural over the three publish writers and a source census over the two cache writers, plus the rescoped `sidecarsymlinkcheck` open census.
3. **The edit-plan payload is read through the path that was confined.** The confinement check and the read use the same canonical path, through one no-follow descriptor.
   - Gate: an `editplanpayloadconfinecheck` arm for a symlinked directory, plus an in-root control.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Git operations now avoid optional locks and filesystem monitoring side effects.
  * Atomic file updates no longer follow symlinks, reducing the risk of overwriting unintended files.
  * Edit-plan payloads are read only from paths that pass containment checks.
  * Repository configuration detection is more reliable across worktrees and alternate repository layouts.

* **Documentation**
  * Updated command help, changelog, and evaluation materials to describe the safety behavior.
  * Updated documentation and presentation materials to reflect 619 gate scripts.

* **Tests**
  * Added coverage for symlink-based file attacks, path traversal, and Git safety policies.

PR #271, opened 2026-09-17
test(churnjoincheck): ripwire and git read "2 weeks ago" at two moments, and the default window's oracle was the wall clock

joyful-ii-V-I:
## Defect

`test/churnjoincheck.sh` compared ripwire's churn with git's own count, but the two sides did not always ask about the same window.

- **The observed flake.** The live audit passed `--since="2 weeks ago"` to ripwire and `"2 weeks ago"` to the git oracle, and each side evaluated it when it ran. A commit on the edge fell inside one window and outside the other. #265 went red on `src/slice.h` churn 15 vs 14, from commit 16efdbcf at the two-week boundary. That was a rerun cause, not a defect.
- **The same bug, slower.** ripwire anchors its default churn window on HEAD's committer epoch minus 12 calendar months (`gitmine.h` `defaultWindowSince`, stamped `window="12mo@HEAD"`). All ten default-window oracles in this gate asked git for wall-clock `"12 months ago"` instead. That is a different window: it holds the same commits today, and it stops holding them on 2027-06-01, when the fixtures' 2026-06 commits leave the wall clock's year.
- **An inert arm next to those oracles.** The offset re-derivation (arm 10) still selected the retired `p="./…"` spelling. It read 0 rows and printed `all 0 rows … re-derive` as a PASS.

## Change (test-only, one file)

- **One clock read.** `NOW_EPOCH` is read once. `SINCE_2W` is derived from it as an ISO-8601 UTC instant, and that identical string goes to both ripwire and git. ripwire hands a date-shaped `--since` to git verbatim, and git reads `YYYY-MM-DDTHH:MM:SSZ` to the second in UTC; both facts were checked before relying on them.
- **The anchored start.** `defaultWindowStart REPO` computes the HEAD-anchored start using `monthsBeforeEpoch`'s rule, and every default-window oracle now uses it. The fixture arms no longer depend on the wall clock at all.
- **New arm 4b** pins both rules to the second:
  - A commit dated exactly on the two-week edge is inside on both sides (1/1) and outside on both sides one second later (0/0).
  - A commit exactly on the anchored edge is counted, and one a second earlier is not (1/1 and 0/0).
  - Two controls must stay red. "2 weeks ago" read by ripwire at N and by git at N+1 must disagree; this is staged with `GIT_TEST_DATE_NOW`, git's own clock override, which ripwire's git child inherits. The wall-clock oracle must drop the anchored edge commit that the default counts. If a git ignores `GIT_TEST_DATE_NOW`, the first control prints an arm-level `SKIP` with the reason.
- **Arm 10** now reads its rows whether or not they carry a leading `./`, and requires the row count to equal `ranked=`. It reads 163 rows, and all of them match git.

## Red first (main's gate vs this gate, the same stagings)

| staging | main's `churnjoincheck.sh` | this branch |
| --- | --- | --- |
| none | rc=0, 72 PASS (arm 10 reads 0 rows) | rc=0, **78 PASS, 0 FAIL, 0 SKIP** |
| #265's straddle: ripwire reads "2 weeks ago" at 8083ea9f's epoch + 14 d, the gate's git one second later | **rc=1, 4 FAIL**: `src/slice.h churn="12"` vs 11, `src/verbs_report.h` 36 vs 35, `src/infra/sortutil.h` 2 vs 1, "3 of 406 rows disagree" | rc=0, 78 PASS |
| git's clock at 2027-07-01 | **rc=1, 12 FAIL** (rename, spelling and offset fixtures) | rc=0, 78 PASS |

The staging runs used a copy of main's gate named `test/.gateprobe.*` (gitignored, removed afterwards), and a `RIPWIRE_BIN` wrapper that moves ripwire's clock only when it inherits the staged value.

## Sweep of `test/` for the same double reading

No other gate evaluates a relative date on both sides of a comparison.

- `sincecheck`, `hotspotsincecheck` and `modifierguardcheck` pass a relative `--since` and check only exit codes, refusals or the `window=` label.
- `recentscopecheck` arm 10e checks only paging attributes.
- `gitstampcheck` and `qddialscheck` compute their relative timestamp once, for a fixture date.
- `sincewindowcheck` and `sincecochangecheck` use absolute dates.

## Gates

- `churnjoincheck` as above.
- gateexitcheck 22 PASS, manifestcheck 16 PASS, skipclassifycheck 22 PASS, gatecountcheck 6 PASS.
- `docs/gatecount_build.py --check` (621) and `docs/limits_build.py --check` agree. No gate file was added and the loop is unchanged.
- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: `regressions="0" gating="0"`.

## Shared products

None: no loop change, shard weight, pin, kParserVer or qschemetrip change. **CHANGELOG is held**, as with #261, so this lands directly; its entry goes into train 1b's CHANGELOG union.

## Summary by CodeRabbit

* **Tests**
  * Improved churn comparison test reliability by ensuring date-window checks use a consistent reference time.
  * Added coverage for two-week and boundary-date scenarios in live repository audits.
  * Strengthened subdirectory audit validation to verify both matching results and expected record counts.


joyful-ii-V-I:
**Disposition at head `beee0834`.** No push: the one CodeRabbit round produced no change to make.

| # | finding | disposition | evidence |
| --- | --- | --- | --- |
| CR 4034076372 | `test/churnjoincheck.sh:384`: skip the wall-clock control "once `12 months ago` reaches" the fixed edge commit (2025-06-09T12:00:00Z), because `W_WALL` would become 1 | **DECLINE: the drift runs the other way** | The wall-clock window starts at now − 12 months, and that start moves *later* as time passes: 2025-09-17 today, 2026-07-01 on 2027-07-01. It moves away from the fixed 2025-06-09 commit, never onto it, so `W_WALL` stays 0 for every clock after 2026-06-09. A clock before that date would put every 2026-06-dated fixture commit in this gate in the future, and the whole gate would be wrong anyway, so a skip for this one arm would hide nothing real. Measured: with git's clock staged at 2027-07-01 (`GIT_TEST_DATE_NOW=1814400000`), this arm prints `PASS anchor control: the wall-clock oracle "12 months ago" names in.cpp in 0 commits where the anchored default counts 1`, and the gate is 78 PASS / 0 FAIL. The same staging on main's gate gives 12 FAIL. |
| CR pre-merge check | Docstring Coverage 71.43% < 80% | DECLINE | This is CodeRabbit's generic threshold, not a project rule, and the same warning is on #236, #241, #249, #250, #258 and #260. The helpers this PR adds (`isoUtc`, `defaultWindowStart`, `gitEdgeCount`) each carry a usage comment in the file's existing style. |

PR #273, opened 2026-09-17
fix(layout): a field's own alignas/attribute/decltype paren was taken for a function's parameter list, and the field vanished

joyful-ii-V-I:
`--layout` silently dropped a data member whenever its declaration carried a `(` that belonged to an `alignas`, `__attribute__` or `decltype` specifier, or sat inside a template argument list — `alignas(8) int x`, `int x __attribute__((aligned(8)))`, `decltype(1) x` and `std::function<void(int)> cb` were all misread as member functions and vanished, while the struct kept reporting `modeled="1"` with a size short by exactly that field's bytes.

`parameterListParen`'s candidate-paren scan now skips a `(` immediately preceded by `alignas`/`__attribute__`/`decltype`, and any `(` sitting inside a template argument list's `<…>`. Each shape now comes back honestly refused (`modeled="0"`, a named caveat) instead of silently missing — never a confidently wrong number.

Gate: `test/layoutcheck.sh` §14, confirmed red on the unfixed source before the fix landed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed layout analysis for fields using `alignas`, `__attribute__`, `decltype`, or function types in template arguments.
  * These fields are no longer silently omitted or incorrectly treated as member functions.
  * Unsupported cases are now clearly reported as refused with an explanatory caveat, while neighboring fields continue to be counted correctly.

* **Tests**
  * Added coverage for decorated fields and template-based function types.


joyful-ii-V-I:
On CodeRabbit's outside-diff finding (`src/layout.h:1298-1302`, postfix `__attribute__` field): this head already stops the silent drop. It does not yet model that field. Probe `struct S { int x __attribute__((aligned(8))); int y; };`:

- **main ea03af88**: `modeled="1" fields="1" size="4"`. `x` vanishes and the size is wrong. That is the bug class this PR fixes.
- **this head c01f8b2c**: `modeled="0" fields="1"` plus `<caveat k="unparsed-member" d="int x __attribute__((aligned(8)))"/>`. The aggregate is refused by name and the dropped declaration is quoted verbatim, so no offset or size is claimed.

Peeling postfix attribute groups so `x` gets its own refused `<f>` row would be a nicer rendering, not a correctness fix. It is noted for the next `--layout` lane and kept out of this PR so CI is not re-run for it.

PR #278, opened 2026-09-17
fix(resolve): assigning a variable from a function call erased the type it was declared with

joyful-ii-V-I:
## Defect

Rule 2 records a C++ **assignment** from a call as the variable's type, so that `x = Foo()` types `x`. A constructor call
and a function call are one grammar node, so the record held the callee's last name whatever it was:
`PHI = PHINode::Create( … )` recorded `Create`, `s = GetDBOptionsFromMap( … )` recorded `GetDBOptionsFromMap`, and
`t = llvm::cast<T>( y )` recorded `cast`. Rule 2's flat per-function table tombstones a variable whose records disagree.
So the non-type erased the declared `PHINode *PHI` / `Status s`, and `PHI->addIncoming( … )` / `s.ok()` fell to the
name ladder. On the fixture and on both corpora that ladder usually declines. The same record:

- tombstoned the field use-site index's table (`--uses=Target.count` lost its owner pin);
- made a **member** assigned from a call (`cur = ns::cast<Target>( y )`) read as a local, so Rule 2b refused the member's
  declared type.

graph.h's varType note said such a type "simply never produces a type::method hit". That holds for the narrow but not
for the tombstone. (The comment above `ctorNameNode` also claimed a class-name filter in buildGraph that did not exist.)

## Change

- **ingest** (`ingest_binds.h`): the assignment arm marks its record with `assignedTypeOf` →
  `RawBind` / `Binding::isFromAssignment`.
- **buildGraph** (`resolve.h assignmentNamesNoClass`): drops such a record unless a class of that name exists. It uses the
  class-name set Rule 2c already built, which is now built first (`classNameSet`). The record is dropped from:
  - Rule 2's `varType` table;
  - `collectFieldUseSites`' table;
  - the local-name set that Rules 2b/2c and the external veto read.
- **Declarations are unchanged.** A declaration's callee-read type (`auto t = makeFoo()`) is still recorded. That
  declaration exists with an unrecorded type, and its conflict with a sibling declaration keeps one block's type off
  the other block's calls (arm 48).
- **Format:** the bind record gains one u8.
  - `kCacheVersion` 22 → 23; `kMinBindRecordBytes` 26 → 27; `cachefuzzcheck`'s blob walker reads the new byte.
  - `kParserVer` 99 → **104** with both `quality.h` mirrors. 100–103 are declared by lanes queued ahead (train 2b: 102,
    #276: 103); the landing train assigns the number.
  - `test/qschemetrip.hash` re-pinned, with a RE-PIN LOG entry.
- `graph.h` stays line-neutral above `rankGraphTeleport` (showcase seed).

## Measured

Method: `--pin-census --no-cache`; C rows joined on (caller id, callee, line); main `fe28fd49` vs this branch.

| corpus | sites changing target | bound calls | newly bound | retargeted | lost |
| --- | ---: | ---: | ---: | ---: | ---: |
| rocksdb `0e2801ac3` | 1,871 | +1,562 | 1,544 | 327 | 0 |
| llvm-project `4d5358b1d` | 4,046 | +2,984 | 2,859 | 1,187 | 0 |

On rocksdb, 1,494 of the 1,871 are `Status::ok`.

**Blinded grading.** A seeded sample of 60 (seed 20260917; 25 rocksdb, 35 llvm-project) was graded against source by six
independent readers. Each saw the two answers as A and B, with the order randomised per site; the key was kept outside
the grading directory.

- **51 better:** 39 NONE→RIGHT, 5 WRONG→RIGHT, 4 PARTIAL→RIGHT, 2 NONE→PARTIAL, 1 WRONG→PARTIAL.
- **6 same:** 4 WRONG→WRONG, 2 RIGHT→RIGHT.
- **3 worse.** All three are resolver floors that the erased type had been hiding, not wrong recovered types:
  - rocksdb `Iterator*` narrowing onto the memtable's same-named `Iterator` classes (the floor #248 records);
  - two llvm-project calls where arity picked the wrong overload of the right class (`getFirstInsertionPt`, `StringRef::find`).
- Wilson 95%: better 73.8–91.9%, worse 1.7–13.7%.

**The local-name-set part on its own** (this branch vs a build without that filter): 23 rocksdb and 132 llvm-project
sites. 15 graded: 10 better, 2 same, 3 worse. One worse site is the same-named-class floor (`PHINode`). Two show a floor
this unmasks but does not cause: Rule 2c reads a member named like a class (`std::unique_ptr<ToolOutputFile> OutputFile;
OutputFile->keep()` → vfs `OutputFile::keep`). That is filed as a follow-up.

### Built and rejected

- **Dropping a declaration's callee-read type too.** It moves 89 more llvm-project sites (0 on rocksdb). 15 graded: 11
  better, 2 same, 2 worse (both the same-named-class floor). Rejected on mechanism: it hands one block's written type to a
  sibling block's `auto t = ns::cast<Decoy>( y )`, a precise edge to the wrong class (arm 48 is red on it).
- **Lifting #276's floor (narrowcheck arm 39d, the unqualified `Vec<T>()` constructor spelling).** Tested on a scratch
  composition of #276 `50b97b7b` + this change (never pushed). Reading that spelling moves 245 llvm-project sites: 170
  lost, 0 gained (779 lost without this change); rocksdb 0. The remaining losses are declaration conflicts, e.g.
  `auto *LI = cast<LoadInst>( … )` beside another `LI`, which this change keeps on purpose. **The floor stays.**

## Gates

`test/narrowcheck.sh` gains arms 44–51, with the fixture generated in-gate:

| arm | checks |
| --- | --- |
| 44 | Rule 2 narrow through `t = ns::cast<Target>( y )` |
| 45 | the same through a plain function call |
| 46 | a member assigned from a call keeps Rule 2b |
| 47a / 47b | controls: a class-named assignment still types and still conflicts |
| 48 | a declaration initialised by a call still tombstones a sibling declaration |
| 49 | `--uses=Target.count` pins |
| 50 | census mechanism is receiver-rule |
| 51 | cold == cold == warm |

Red first:

- **main `fe28fd49`:** 44, 45, 46 (`no edge`), 49 (`owner_candidates="2"`), 50 (`NO-CENSUS-ROW`); 47a/47b/48/51 green.
- **Variants of this change** (env toggles in a scratch build, never committed):
  - declarations dropped too → 48 red (`pick@lib/target.h:1`);
  - local-name-set filter off → 46 red;
  - new byte not persisted → 51 red;
  - guard off → 44/45/46/49/50 red.

Results:

- `narrowcheck` ALL PASS (46 rows).
- `qschemetripcheck`, `qextractionkeycheck`, `cacheidentitycheck`, `portablecachecheck`, `cachefuzzcheck`: ALL PASS
  (cachefuzz's ASan sweeps skip locally: no ASan tree, by instruction).
- rocksdb map byte-identical across two cold runs and warm vs cold; `xmllint --noout` clean; the same for this repo's own map.
- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: `gating="0"` after 3 acks written through the binary
  (`--ack-only=buildFieldNarrowTables,classNameSet`; the ledger diff is exactly three `+ack` rows). One minor verbosity
  row is left.
- Full suite, `python3 test/pargates.py . ./build/ripwire -j 6` at 379d9faa: `gates=639 pass=636 skip=3 fail=0`. The skips are
  editchecknotecheck, argvdiffcheck and g1freshcheck (no `asan/` tree). The first run, on a binary built before the commit,
  failed `versioncheck` alone (`built_from=fe28fd49e+dirty`): a true positive, fixed by rebuilding at HEAD, then the whole
  suite was re-run.

## Pins moved

- `kCacheVersion` 22 → 23, `kParserVer` 99 → 104, and both `quality.h` mirrors; the train assigns the final number.
- `test/qschemetrip.hash` `72f930b8` → `f7ffb332`.
- `.ripwire_quality_acks` +3.

## Train notes

- `test/narrowcheck.sh`: union arms 26–38 (#268), 39–43 (#276) and 44–51 (this).
- `src/ingest_binds.h`: the `ctorNameNode` comment conflicts with #276's rewrite; keep #276's text and add this lane's
  assignment sentence.
- `kParserVer` / qschemetrip: re-derive on the merged tree.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed type resolution for variables assigned from functions, constructors, and factory calls.
  - Prevented function names from being incorrectly treated as types when narrowing fields and local variables.
  - Preserved valid type narrowing for known class-based calls and declaration initializers.
  - Improved handling of conflicting local declarations and field-use analysis.
  - Added coverage for assignment-based narrowing, deterministic results, and cache behavior.


joyful-ii-V-I:
This PR lands through integration train 3 (with #268 and #276), built on train 1b (#277), after an independent review. I labelled it `train-member` and cancelled its full-matrix run: CI is the landing bottleneck right now, and the train PR runs the full matrix on the composed tree.

joyful-ii-V-I:

PR #279, opened 2026-09-17
feat(lsp): Phase 1 --lsp — read-only navigation LSP server over stdio

mpapis:
## What
`--lsp`: a read-only navigation LSP server over stdio — `definition`, `references`,
`documentSymbol` (member variables merged into the outline), workspace symbols, `hover` —
served off the same warm index `--mcp` uses. No second parser, no second process; refuses
`--mcp`/`--listen` (one protocol per stdin) and refuses knobs it would silently ignore.
Hover carries two clickable tiers: **Used at** (the name-matched call-role floor `--uses`
answers with) and **Referenced at** (Ruby constant-load directives).

## Provenance / review note
This work was generated fully by an LLM coding session. I (mpapis) am not fluent enough in
every touched area to personally vouch for its quality and adherence to the standard —
review accordingly. It is an experiment in whether ripwire can be versatile: not only MCP
for AI agents, but LSP for IDEs.

The experiment already paid for itself in a side effect: exercising the server against a
Rails app exposed weaknesses in **Ruby call-site extraction**. (1) Constant references like
`include Foo` exist only as `Include.isSymbolic` load directives, never as call edges;
(2) identifiers in argument position are not extracted as calls at all — `Result.new(
total_price, [])` on line 13 of a file defining `total_price` on line 18 yields
`--uses=total_price` → `count="0"` while the file names the method twice. Every Ruby
uses-count is a floor, and hover discloses which table each tier's answer came from. Fixing
extraction would touch the PageRank input and its gates — deliberately out of scope here,
but this is the evidence it deserves its own issue.

## Gate work
- `test/lspcheck.sh` — 16 arms, scripted client speaking real LSP framing (new: use-tier
  asserts; a Ruby fixture asserting clickable `#L` links).
- `test/mcpattrparitycheck.sh` — the analyze twin arm was grading top-200 window
  *membership* instead of dialect parity; now pinned to a 10-row whole-shown fixture with a
  non-vacuity guard.
- Design record `docs/LSP.md` (D7 covers the tiers); CHANGELOG updated.

## Evidence
- Full suite: 633 gates green under correct preconditions (the 4 DAC-premise gates pass as
  uid 1000 — environment artifact, not defect). macOS/ASAN PR legs per ci.yml.
- ASAN clean: full-tree and a live `--lsp` dialog. Determinism: byte-identical reruns;
  `xmllint` OK. `quality_delta`: gating 0 (advisories: new-symbol cx 22 / params 7; one
  dead-code report on a macro-dispatched trait method is a false positive).
- Hover latency 11–14 ms on a 2,046-file repo.

## Known floors
`Used at` is name-matched and call-role only; `Referenced at` tail-matches after `::`
(Zeitwerk-style) and renders only when non-empty; unsaved buffers are absent. All labelled
in place.

## Summary by CodeRabbit

- **New Features**
  - Added a read-only `--lsp` navigation server over standard input/output.
  - Supports definitions, references, document symbols, workspace symbols, and hover information.
  - Uses saved index data with UTF-8 positions and includes merged member fields in document outlines.
  - Rejects incompatible combinations with `--mcp`, `--listen`, and query-budget options.

- **Documentation**
  - Added LSP usage and design documentation, including command references and verification guidance.
  - Updated long-flag counts and related presentation materials.

- **Tests**
  - Added end-to-end coverage for LSP protocol behavior, navigation results, lifecycle handling, and transport restrictions.


joyful-ii-V-I:
Thanks for this, @mpapis, and for being upfront that an LLM wrote it; that made it easy to review on the merits. We put `--lsp` through an adversarial pass, and nothing crashed, hung, or read a byte outside the indexed root. The pass covered:
- path traversal through absolute paths, `..` segments, a symlink planted out of the root, and percent-encoding;
- `Content-Length` values that were negative, non-numeric, missing, oversized, duplicated or flooded;
- a dozen malformed JSON-RPC shapes.

Three things stood out as the right calls:
- **`lspFindFile`** matches the client's URI against a path that's already indexed and never opens the raw string. That's exactly the shape a server like this needs.
- **Reusing `mcpjson.h`'s parser** and the house resolvers, instead of a second copy of either.
- **Hover discloses** that its call rows are a name-matched floor. The Ruby `Referenced at` tier is a genuinely useful addition.

This is ready to land. It goes in with our next integration train, where we reconcile the shared pins (gate count, CHANGELOG). Please don't rebase. We'll add three small maintainer follow-ups on top, and your commits stay as they are:
1. Fold a few of those framing and traversal cases into `test/lspcheck.sh`, so this hardening stays proven in CI.
2. Reconcile the arm count, which reads 14 in the gate and 15/16 in CHANGELOG and docs.
3. Disclose the `workspace/symbol` 20-row cap. LSP has no truncation field there, so it'll likely go through `window/logMessage` or a raised cap.

A great first step onto the editor surface.

joyful-ii-V-I:

Issue #74, opened 2026-09-09
Java: a method reference (Type::method) is not a call site for --uses/--callers


davide94it:
### What happens

Two functionally identical Java methods, one written with a lambda and one with a method reference:

```bash
mkdir rw-mr && cd rw-mr
```

```java
cat > Util.java <<'EOF'
public class Util {
    public static String conv(Object o) { return String.valueOf(o); }
}
EOF
```

```java
cat > A.java <<'EOF'
import java.util.List;

public class A {
    public List<String> lambdaForm(List<Object> in) {
        return in.stream().map(item -> Util.conv(item)).toList();
    }

    public List<String> methodRefForm(List<Object> in) {
        return in.stream().map(Util::conv).toList();
    }
}
EOF
```

```bash
ripwire . --no-cache --uses=conv --format=rows
ripwire . --no-cache --callers=conv --json
```

--uses finds one call site out of two — only the lambda, at line 5:

```xml
<uses of="conv" defs="1" external="0" count="1" root="." graph_ambiguous="0"
 graph_unresolved="0" counts_floor="1" format="columnar">
  <paths>0=A.java</paths>
  <cols n="1" fields="path,line,role,in_id">
    <path>0</path><line>5</line><role>call</role><in_id>lambdaForm</in_id>
  </cols>
</uses>
```

--callers likewise lists only lambdaForm; methodRefForm does not appear:

```json
{"of":"conv","defs":1,"count":1,"root":".","hop_tested":0,"hop_untested":1,
 "graph_ambiguous":0,"graph_unresolved":0,"counts_floor":true,
 "callers":[{"t":"method","n":"lambdaForm","p":"A.java:4"}]}
```

Environment: ripwire 0.5.0 (Release, GNU 14.2.1, built_from=unknown), Ubuntu 24.04.4 LTS x86_64, each arm in a fresh directory with --no-cache.

### Why I'm reporting it rather than reading it as the documented callback caveat

--callers' legend lists what name-based extraction cannot see: dynamic dispatch, and callbacks bound to more than one function in scope. Util::conv is neither. The receiver is a type name and the member is a literal identifier in the source text, so the target is fixed at compile time — as fixed as the Util.conv(item) three lines above it, which does produce an edge. Delete conv and both forms stop compiling; only one of the two is visible to ripwire.

[...]

PR #280, opened 2026-09-17
fix(resolve): a C++ member named like a class was read as that class — on top of typedef/using alias bases

joyful-ii-V-I:
## The defect

Rule 2c reads `Cls.m()` through a class name as a call on that class. It refused only when a **local** of that name
exists. Inside a C++ member function, name lookup finds a member of the class or of a base before any namespace-scope
class, so a member named like a class is the member. Two llvm-project @ 4d5358b1d instances were graded while reviewing
`lane/ctor-inferred-type`. Each resolved to one precise, wrong edge:

- `llvm/lib/DebugInfo/LogicalView/Core/LVReader.cpp:175` `OutputFile->keep()`. The member is
  `std::unique_ptr<ToolOutputFile> OutputFile` (LVReader.h:35). It resolved to `llvm/lib/Support/VirtualOutputFile.cpp`
  `OutputFile::keep`.
- `llvm/lib/Transforms/IPO/SampleProfile.cpp:1962` `Reader->read()`. The member is the base-class
  `std::unique_ptr<SampleProfileReader> Reader` (SampleProfileLoaderBaseImpl.h:318). It resolved to msgpack `Reader::read`.

## Why this PR has two commits

The veto alone, as first specified, **graded net-worse** on llvm-project:

- It lost 1,772 edges.
- A seeded, blinded sample of 60 retargets graded **23 better / 3 same / 34 worse**.

33 of the 34 worse sites are one mechanism. clang's `CodeGenFunction` has a member `CGBuilderTy Builder`, and
`class CGBuilderTy : public CGBuilderBaseTy` is declared with `typedef llvm::IRBuilder<…> CGBuilderBaseTy;`. On main,
`Builder.CreateCall()` reached `IRBuilderBase::CreateCall` only by coincidence: a class `Builder : IRBuilder` in
`HexagonVectorCombine.cpp`. Once the member hides that class, Rule 2b types `Builder` as `CGBuilderTy`, and its base
walk dead-ends at the typedef.

So the first commit makes aliases walkable, and the second adds the veto on top. The owner chose this order in the lane
session.

### 1. `fb90a34a` fix(resolve): a base class or member type reached through a C++ typedef or using alias ended the base walk

**What records.** `ingest_relations.h captureTypeAlias` records a plain C/C++/ObjC alias of a named class. It handles
`typedef`, including multi-declarator typedefs, and `using X = Y;`. The record rides the compose shape:

- `composeRel "alias"`, with an empty `fieldName`
- `recvVar` = the alias
- `name` = the target class
- `qualifier` = the namespace the target was written in

It is collected on the side stream and appended after the walk, so `refs` keeps one writer per language during it.

**What reads it.** `resolve.h addTypeAliasBases` adds alias → target to **chaUp only**. It is refused, or not followed, in
these cases:

- A target written in `std`.
- An alias whose name is also a real class elsewhere: a class-like symbol at a line no alias record holds. A typedef is
  itself a Struct symbol at its name's line.
- graph.h's HAS-A loop refuses the record, so it adds no compose row.

**Also covered: alias templates.** `template <typename T> using SetTy = SmallPtrSet<T, 8>;` is an `alias_declaration` inside a template declaration, so a member typed `SetTy<Foo>` walks on to `SmallPtrSet` and its bases. The first version of this PR body did not disclose this; the independent review named it.

**Known floor (arm t11, from the independent review).** An alias records its target's class name without template arguments. `typedef SubT<marks> subtree;` with `marks` the enclosing template's own parameter walks to the primary `SubT::is_null` alone, dropping `SubT<true>::is_null` (rocksdb `omt_impl.h`, `subtree_templated<true>`). That is a lost candidate, not a wrong-class pin; main split over both. The control `typedef SubT<false>` narrows correctly.

**What never records.** Pointer, reference, array and function aliases; targets that name no class; self-aliases;
function-local aliases.

**Pins.** The cache format is unchanged. kParserVer goes to **111**, declared over main's 110 after train 3 merged (first declared 105); the landing train assigns the real one.

### 2. `36aa4f56` fix(resolve): a C++ member named like a class was read as that class

- `Narrower::memberFieldNames` builds `"<Owner>#<field>"` from the C/C++ field side table. That table holds every
  declarator shape, including the `std::unique_ptr<T>` members Rule 2b's compose table skips.
- `memberFieldHides` walks the caller's class and its chaUp bases.
- Rule 2c refuses a C++/ObjC receiver that names such a member. It also refuses when the 16-name walk stops with a base
  unvisited.
- The three name sets travel as `ClassNameRecvNames` (params 4, not 6).

graph.h stays line-neutral above `rankGraphTeleport` in both commits, so the showcase seed holds.

## Measured

Method: `--pin-census --no-cache` on each binary, C rows joined on (caller id, callee, line). A site is retargeted when
its target multiset differs.

| step | rocksdb @ 0e2801ac3 | llvm-project @ 4d5358b1d |
| --- | --- | --- |
| main → alias commit | 366 retargeted, +111 bound, **0 lost** | 2,649 retargeted, +1,146 bound, **0 lost** |
| alias commit → veto commit | 0 retargeted | 1,398 retargeted (1,346 changed, 52 lost) |
| main → this PR | 366 retargeted, +111 bound, 0 lost | 3,981 retargeted, +1,093 bound, 49 lost |
| *(rejected)* main → veto without the alias commit | 0 | 3,084 retargeted, **1,772 lost** |

**Grading.** Each sample is seeded and blinded. The A/B order is randomised per site, and the key was kept outside the
grading directory. Four independent graders took 15 sites each, graded against source with a fixed rubric.

| sample | better | same | worse |
| --- | --- | --- | --- |
| alias commit, 60 sites (25 rocksdb + 35 llvm) | **57** | 2 | 1 |
| veto on the alias base, 60 llvm sites | **51** | 7 | 2 |
| *(rejected)* veto without aliases, 60 llvm sites | 23 | 3 | 34 |

What the non-better grades are:

- **Alias commit, 1 worse.** A class template specialization (`subtree_templated<true>`) sharing the primary's name.
- **Veto, 2 worse.** One Rule 2b floor: `using CGBuilderBaseTy::CreateGEP;` re-exports base overloads that Rule 2b's
  own-class overload pick shadows.
- **Veto, 7 same.** WRONG→WRONG members whose types nothing records (`SmallVector<T, N>`, `std::` types).

**Walk cap.** A cap-4096 variant of the member walk produced a byte-identical census to cap 16 on both corpora (measured
on the veto before the alias commit). The cap decided no site.

**The two graded instances.** In a scratch composition of `lane/ctor-inferred-type` (379d9faa) plus this PR, neither
site has an edge. The ctor lane's own census pins them to `OutputFile::keep` and msgpack `Reader::read`.

## Gates

**Commit 1: `test/fieldnarrowcheck.sh` arm t (t1–t10), generated fixture.**

- **Red first.** t1–t4 are red on main (no edge). t5–t10 are green on main and stay green.
- **Mutation proof.** In a scratch build with one guard disabled at a time:
  - no std refusal → t5 red
  - no real-class guard → t9 red
  - no function-local skip → t10 red
  - no HAS-A refusal → t6 compose row red
  - unmutated → ALL PASS
- **Live controls.** Both t6 no-leak probes have a positive control: CGB's real base clause appears in
  `--uses role="extends"` and in `--lego`.

**Commit 2: `test/clsrecvcheck.sh` arms H–N, generated C++ fixture plus a Python control.**

- **Shapes.** Smart-pointer member (H); raw-pointer member, then typed by Rule 2b (I); base member (J); class template
  base member (K); member past the walk cap (L).
- **Controls.** M: the same call from a class without that member keeps the route. N: a Python `self.Interval` does not
  veto `Interval.validate(v)`.
- **Red first.** Red on main and on commit 1's binary: 9 FAIL rows (H ×2, I, J ×2, K ×2, L ×2), with M and N green.

**Targeted gates, all rc=0.** fieldnarrowcheck, clsrecvcheck, narrowcheck, localitycheck, chainguardcheck, identitycheck,
externalvetocheck, pincensuscheck, cachefuzzcheck, showcasecapturecheck, xmlwellformed, printffmtparitycheck,
compactlegendcheck, recallevalcheck, qschemetripcheck, qextractionkeycheck, cacheidentitycheck, manifestcheck,
gatecountcheck.

**Other checks.**

- Full suite: `python3 test/pargates.py . ./build/ripwire -j 6` on head 36aa4f56 → `gates=639 pass=636 skip=3 fail=0 wall=921.0s` (the three environmental skips: argvdiffcheck, editchecknotecheck, g1freshcheck with no asan/). An earlier run on the pre-amend head was 632/3/4 fail. All four were true positives, fixed by the amend: a stale binary for versioncheck, a duplicate `kFieldWalkCap` for limitstablecheck and readmedriftcheck (now one shared `Narrower::kFieldWalkCap`), and a one-line `&& ok || no` for gateexitcheck. The amended head's census is byte-identical to the graded binary on both corpora.
- Determinism: the repo map is byte-identical across two runs, and xmllint accepts it.
- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: `gating="0"`. The 3 minor rows are the side-stream arm
  idiom in `captureSideFacts` and `streamSideCaptures`, which were already over bar.

## Pins moved

- `kParserVer` → 111 (`src/ingest_cache.h`) and `kIngestParserVerMirror` (`src/quality.h`), over main's 110. kCacheVersion stays main's 23.
- `test/qschemetrip.hash` was re-pinned through `UPDATE_GOLDEN=1`, with a RE-PIN LOG entry in `test/qschemetripcheck.sh`.
- `docs/LIMITS.md` was regenerated by `docs/limits_build.py`: the `kFieldWalkCap` row note only. The two walks share one `Narrower::kFieldWalkCap`, so the cap count stays 217.

## Merge of train 3 (head e5f9a50a)

`origin/main` a2b3cd6e (#281) was merged with rerere off and resolved by hand. Every conflict was two sides inserting beside each other: graph.h (memberFields beside #74's javaNarrower), resolve.h (alias helpers beside classNameSet/assignmentNamesNoClass), kParserVer/mirror (111 over 110, cache 23), qschemetrip (re-derived), fieldnarrowcheck (arm t beside arm r), CHANGELOG. On the merged build: clsrecvcheck 29 PASS (H–N), fieldnarrowcheck 110 PASS (t1–t11), narrowcheck 76 PASS, and the pin, gate-shape and generator checks all green. Commit 3a6e6b83 adds arm t11 (the dependent-argument alias floor) and the alias-template disclosure from the independent review.

## Train notes

- **ctor-inferred-type.** `lane/ctor-inferred-type` (#278) also bumps kParserVer (104) and kCacheVersion (23), and adds
  `classNameSet` / `assignmentNamesNoClass` beside `fieldTypeWrittenInStd` in resolve.h. A merge needs the union of both
  resolve.h blocks, the higher cache version from that lane, and a fresh kParserVer plus qschemetrip re-derivation on the
  merged tree. Merge with `rerere.enabled=false`. My scratch composition resolved these by hand and was never committed.
- **CHANGELOG.** This PR adds two `### Fixed` entries under `[Unreleased]`.


joyful-ii-V-I:

PR #282, opened 2026-09-17
fix(ingest): a member held by std::unique_ptr or std::shared_ptr recorded no type, so calls through it guessed

joyful-ii-V-I:
## The defect

`src/ingest_relations.h` `captureFields` builds the S5-E HAS-A record that feeds Rule 2b's `fieldTypeByClass`
(`src/graph.h` `buildFieldNarrowTables`). It read a qualified type only when a `type_identifier` sat directly under the
`qualified_identifier`. In `std::unique_ptr<ToolOutputFile> OutputFile;` the name there is a `template_type`, so the member
recorded **no type at all**, and every `OutputFile->keep()` took the bare-name ladder (a split, a locality pick or no
edge). `std::string name_;` was recorded (qualifier `std`, refused later), but no smart-pointer member ever was.

## The change (decided by measurement)

- **Capture.** A member written `std::unique_ptr<T>` / `std::shared_ptr<T>` records T's final segment and T's written
  namespace (so `std::unique_ptr<std::string>` is refused like any std type), marked `viaArrow`. These record nothing,
  as before: a pointer to the smart pointer, `T[]`, a primitive, and a qualified-template pointee.
- **Receiver shape at resolve time.** It was not available: `RecvKind::NamedVar` covers both `.` and `->`. Every
  C++/ObjC call reference now records `viaArrow` (read off `field_expression`'s `operator`). Rule 2b requires it on an
  arrow-only field. `p->m()` narrows to `T::m`, and `p.m()` stays on the unchanged ladder, because it names the smart
  pointer's own `reset`/`get`.
- **Tombstone.** `FlatRecvType` gains `arrowOnly`, and `recordFlatRecvTypeFact` treats "T through `->`" and "T as the
  member" as different facts. Two same-named classes that disagree now tombstone, as two types already did.
- **`--uses=Owner.field`** reads the same record, so `w_->level` pins to the pointee's `level`. A std smart pointer has no
  data members, so a `.` read cannot mean anything else.
- **HAS-A output is unchanged.** The compose block skips the new records.
- **Format.** The ref record grows one u8: `kCacheVersion` 22 → 23, `kParserVer` 103 → 104 (the next free number over
  main; a train renumbers it), both `quality.h` mirrors, and `kMinRefRecordBytes` 39 → 40. Its runtime tripwire caught
  the stale value on the first run.

**Rejected by measurement:** recording every other `std::Tmpl<…>` member (vector, optional, unordered_map) as a std
type, which only tombstones same-named classes' members. On rocksdb and llvm-project that moved 6 sites and every one
got worse: 4 correct narrows lost (db_stress `Stats#hist_` against db_bench's `unordered_map` `hist_`, and llvm AMDGPU
`SIMachineFunctionInfo#ArgInfo` against its YAML twin's `std::optional`), 1 turned into a wrong split, and 1 moved to the
enclosing class's own method. It refused no wrong narrow.

## Measured

Method: `--pin-census --no-cache` on main `a5ce95e2` against this branch. C rows are joined on (caller id, callee, line);
a site counts as retargeted when its target multiset differs.

| corpus | retargeted | gained | changed | lost | bound |
| --- | ---: | ---: | ---: | ---: | ---: |
| rocksdb @ 0e2801ac3 | 809 | 448 | 355 | 6 | +442 |
| llvm-project @ 4d5358b1d | 793 | 498 | 295 | 0 | +501 |

rocksdb's 6 lost rows: five ObjectRegistry `parent_->Dump()`-style calls now resolve to their own method (a
`shared_ptr<ObjectRegistry>` parent, and the census omits self edges), and one `Rep::file->file_name()` is now
tombstoned where it used to hit the wrong class.

**Blinded grade.** A seeded (20260917), stratified sample of 60 retargets. Strata: rocksdb 12 gained / 12 changed / all 6
lost; llvm 15 gained / 15 changed. Answers were shown in random A/B order, the key was kept outside the grading directory,
and four independent graders read source under the previous lane's rubric (WRONG < NONE < PARTIAL < RIGHT):

| stratum | better | same | worse |
| --- | ---: | ---: | ---: |
| rocksdb gained | 11 | 0 | 1 |
| rocksdb changed | 8 | 2 | 2 |
| rocksdb lost | 6 | 0 | 0 |
| llvm gained | 15 | 0 | 0 |
| llvm changed | 13 | 2 | 0 |
| **all** | **53** | **4** | **3** |

The 3 worse sites are limits Rule 2b already had, now reached through a smart pointer. One is a pointee class name
(`Iterator`) shared with nested `MemTableRep::Iterator` classes. Symbol scopes drop namespaces, so `Iterator::key` finds
the nested ones. The other two are overload picks (`WritableFile::Append`, `Cache::Release`) that miss a default argument.

**Is the `->` bit worth a format byte?**
- Without it, 9 llvm sites bind the pointee's same-named method: `MC.reset(new MCContext…)`, `MII.get()`,
  `DT.reset(new DominatorTree…)` and more. All 9 are `.` calls on the smart pointer. rocksdb has none.
- A list of smart-pointer member names (refuse `reset`/`get`/… without knowing `.` vs `->`) would also refuse 38 llvm
  narrows this change makes through `->` (31 `->get()`, 7 `->reset()`).

Cost: the rocksdb cache grows 356,502 B (+1.15%). The warm census equals the cold one on rocksdb.

## The two named llvm sites: neither is reached by this change alone (measured)

- **`llvm/lib/DebugInfo/LogicalView/Core/LVReader.cpp:175` `OutputFile->keep()`** still has no edge. The capture now
  types `LVSplitContext#OutputFile` as `ToolOutputFile`, arrow-only. But two lines up,
  `OutputFile = std::make_unique<ToolOutputFile>(…)` is a C++ assignment, and it mints a local binding for `OutputFile`.
  Rule 2b's local-shadow veto (`localNameSet` takes every binding kind) then refuses the member. The proof: in a
  three-file copy, deleting that one assignment line makes the site `receiver-rule → ToolOutputFile::keep`. With a decoy
  `keep` added, the unchanged copy pins the decoy instead. In C++ an assignment never declares, so this is a separate
  veto defect.
- **`llvm/lib/Transforms/IPO/SampleProfile.cpp:1962` `Reader->read()`**: `Reader` is declared in the base
  `SampleProfileLoaderBaseImpl<FT>` (`SampleProfileLoaderBaseImpl.h:318`). Rule 2b looks fields up on the enclosing
  class only (`SampleProfileLoader#Reader`), so the base-class member is never found.

## Gates

`test/fieldnarrowcheck.sh` **arm p**, written first.
- Fixture: `std::unique_ptr<Widget>`, `std::shared_ptr<Widget>` and `std::unique_ptr<store::Blob>` members, each
  called next to a same-named decoy.
- Controls: `std::vector<Widget> v_; v_.size()`, an in-repo `Holder<Widget>` and an in-repo `util::Box<Widget>`.
- Collisions: a std pointee and a `.`/`->` pair, each in both record orders.

**Red on `a5ce95e2`:** p1–p4 (the narrow), p7 (`--uses`), all four p8 tombstones, and p9 (the warm cache).

**Mutations** (each built, run and reverted):

| mutation | turns red |
| --- | --- |
| ignore the `->` bit | p5 (`w_.reset()`) |
| read through any template | both p6 in-repo template controls |
| drop the arrow-only tombstone | a p8 row |

Also updated:
- `test/cachefuzzcheck.sh`'s record walker learns the new byte (Part 3 went red without it).
- `test/qschemetrip.hash` is re-pinned for both versions, with a RE-PIN LOG entry.

Results:
- Full suite, one run on `2e81e1dc` (the fix commit): `gates=646 pass=642 skip=2 fail=2 wall=1175.3s jobs=6`. The skips
  are the two environmental ones (`argvdiffcheck`, `editchecknotecheck`). Both failures were one arm:
  `localitycheck` (6), which `canoncheck` also runs. That arm used `std::unique_ptr<Target> rep_` as its example of "a
  member type Rule 2b cannot read", and this change makes that type readable, so `rep_->pick()` correctly narrowed. The
  second commit `ad7a9d56` keeps the arm's intent with `ns::Handle<Target> rep_` and adds arm 6b for the smart-pointer
  narrow (6b is red on `a5ce95e2`). After it, `localitycheck`, `canoncheck`, `gateexitcheck`, `manifestcheck`,
  `gatecountcheck`, `versioncheck` and `g1freshcheck` all pass. **No single run is all-green**: the second commit changes
  only test files, and its union with the full run is.
- Targeted and green on the fix binary: cache family, `chainguardcheck`, `clsrecvcheck`, `composelangcheck`,
  `cppqualcheck`, `cpptmplscopecheck`, `decltodefcheck`, `fieldnarrowcheck`, `fieldusescheck`, `importnarrowcheck`,
  `narrowcheck`, `narrowlangcheck`, `pincensuscheck`, `qextractionkeycheck`, `qschemetripcheck`, `rangecomposecheck`,
  `shadowcheck`, `showcasecapturecheck` (graph.h is line-neutral above `rankGraphTeleport`), `typerefcheck`.
- ASan+UBSan+LSan: `fieldnarrowcheck` ALL PASS under `asan/ripwire` with no sanitizer lines, and so were the shape probe,
  the LVReader slice and rocksdb.
- Determinism: the rocksdb and self maps are byte-identical across runs and well-formed (`xmllint`).
- `--quality-delta=$(git merge-base origin/main HEAD)..HEAD`: **gating="0"**, 4 minor rows. They are captureFields
  complexity 75→81 and verbosity 121→135, captureTagsFacts verbosity +1, and `recordFlatRecvTypeFact`'s added
  parameter. The working-tree form had also shown 3 gating `short-horizon-churn` rows, on the two version mirrors and
  `recordFlatRecvTypeFact`. All three are symbols this change has to edit, and none is in a foreign file.

**Train note.** `kParserVer` 104 and `kCacheVersion` 23 are claims. #278 also takes cache version 23, so the train
builder assigns the final numbers and re-pins `qschemetrip`.

## Pins moved

- `test/qschemetrip.hash`: `95a27416…` → `f7ffb332…` (kParserVer 104, kCacheVersion 23; log entry in `test/qschemetripcheck.sh`)
- `kMinRefRecordBytes` 39 → 40 (`src/ingest_cache.h`, derived by its own runtime tripwire)
- `test/localitycheck.sh` arm 6 fixture member `std::unique_ptr<Target>` → `ns::Handle<Target>`; new arm 6b

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved C++ member resolution for `std::unique_ptr` and `std::shared_ptr`.
  - Calls using `->` now resolve against the smart pointer’s pointee type.
  - Calls using `.` continue to resolve against the smart pointer type.
  - Improved handling of conflicting member types and qualified pointee types.

- **Tests**
  - Added regression coverage for smart-pointer access, cache compatibility, determinism, and unsupported template cases.


joyful-ii-V-I:
This PR is part of integration train 5, #284, and will show as merged when that train lands.

PR #286, opened 2026-09-18
rename(selfcheck): the self-check macro vocabulary lands — ASSUME/EXPECTS/ENSURES/DASSERT/UNREACHABLE/VALIDATE, and DEGRADED_PATH_ALERT becomes DISCLOSE

joyful-ii-V-I:
The self-check macros get the names the code has been asking for, and a degrade path can finally say what it degraded to.

| Old | New | Means |
| --- | --- | --- |
| `VERIFY` | `ASSUME` | a fact the optimizer may rely on |
| — | `EXPECTS` / `ENSURES` | a precondition / a postcondition |
| `DEBUG_ASSERT` | `DASSERT` | checked in debug builds only |
| — | `UNREACHABLE` | a branch that cannot be taken |
| (ad hoc) | `VALIDATE` | external input, always checked |
| `DEGRADED_PATH_ALERT` | `DISCLOSE` | a degraded answer, with a named reason |

`DISCLOSE( sink, why[, "msg"] )` is the minimal sink form: the reasons are a closed set in one declaration, so adding a reason without handling it is a `-Wswitch` error rather than a silent gap.

**Scope.** 839 renames across 179 files, applied by `rename_selfcheck.py` (committed evidence, not the tool: the run is idempotent and `--check` reports 0 remaining). On this branch `src/` has **0 `VERIFY(`** and **0 `DEGRADED_PATH_ALERT`** left, against 168 `ASSUME(` and 228 `DISCLOSE(`. `src/infra/Diagnostics.h` and `diagnostics.cpp` are hand-written and replace the old header whole.

**Hand work, beyond the script:**
- `gitmine.h` takes `VALIDATE`, and two sites from #241 convert to it.
- 25 disclose sites added since the classification was made (13 files) are classified by the same rubric; the table is in the lane report. No `src/` site converts to the sink form in this pass, by design.
- `docs/limits_classes.tsv` drops a stale `kNoticeByteCap` row (220 caps), and a README cap-count drift is corrected.

**The new gate.** `test/selfcheckcheck.sh` pins the vocabulary: a stray old spelling, an unhandled reason, or a `VERIFY( false )` on a degrade path fails it. Registered at all five sites; the gate count pin was re-measured on this tree.

**No behaviour change, proven.** `ripwire . --no-cache` is byte-identical between an `origin/main` build and this branch's build on three trees: this worktree, a main baseline checkout, and the cache fixture tree.

**Gates** (BIN passed as `$1`), all rc 0: selfcheckcheck, manifestcheck, gateexitcheck, gatecountcheck, `limits_build.py --check`, g1configcheck, binoverridecheck (651 gates, 0 false-greens), printffmtparitycheck, docscommandscheck, readmedriftcheck, ripwirepubliccheck, determinism, xmllint. Plain `--clean-first` build, 0 warnings. `--quality-delta` gating 0, with the rename's short-horizon-churn findings acked through the binary in one batch.

Not verified locally, and left to this run: gcc, AppleClang 16, clang-cl, ASan and Release.

## Summary by CodeRabbit

* **Changed**
  * Updated the self-check and diagnostic vocabulary across the application, including assumption, validation, assertion, and disclosure messages.
  * Release and debug builds now present the updated diagnostic categories and reporting behavior.
  * Added support for explicit validation of external input and structured disclosure sinks for recoverable paths.
  * Updated no-alias, thread-ownership, and unreachable-condition checks to use the current terminology.

* **Documentation**
  * Refreshed contributor guidance, architecture references, changelog entries, and build notes to reflect these changes.

* **Tests**
  * Added coverage enforcing correct self-check and disclosure usage.


joyful-ii-V-I:
Commit after the first push: **197db74a**.

The hand-written \`diagnostics.cpp\` in the prepared package had re-introduced a pattern main already fixed: a multi-insertion \`std::cerr <<\` reporter, whose notices tear under concurrent load. \`test/diagnoticecheck.sh\` exists on main for exactly that reason (18 flaky CI failures in 5,700 runs, a 32–73% tear rate when measured). The full suite caught it here; the required gate list had already passed.

197db74a re-derives the new \`diagnostics.cpp\` from main's one-write shape while keeping this PR's \`CheckKind\` banner and blame design, fixes the matching test harness, and fixes three gates the rename had reddened:
- \`infraportcheck\` — a rename fixup named "ripwire" inside a file required to stay project-neutral;
- \`fixedbufsweep\`;
- \`hazardpatterncheck\` — its classifier would have silently stopped flagging a swallowed catch that only calls \`DISCLOSE\`.

Re-verified after the fix: plain build with 0 warnings, determinism and xmllint clean, \`ripwire . --no-cache\` byte-identical to an \`origin/main\` build on three trees, and \`--quality-delta\` gating 0.

Follow-up, not in this PR: the same fixup bug should be fixed in \`rename_selfcheck.py\` itself, not only in its output.

joyful-ii-V-I:
Commit **d9fe7a9c**: the showcase caption was stale, not the recount. On this tree the top-10/50/100 figures recount as 92.5 / 87.1 / 86.4%, where the caption still read 91.8 / 85.2 / 85.5%. Only the caption string in `test/showcase_capture.py` changed; no capture was re-recorded, so nothing scouts the local repo for branch names. showcasecapturecheck, docscommandscheck, readmedriftcheck and determinism all pass with the binary passed as `$1`.

joyful-ii-V-I:
Commit **05852a1e**: the 8 names `selfcheckcheck` arm A flagged were `VERIFY`, `VERIFY_TEXT`, `VERIFY_DEBUG_ONLY`, `VERIFY_NOT_REACHED`, `VERIFY_SAME_THREAD`, `VERIFY_NO_ALIAS`, `DYNMAP_VERIFY` and `DEGRADED_PATH_ALERT` — all on one line of `test/showcase_capture.py`, in the caption that names what the rename changed. A legitimate historical mention, so it takes a `HISTORY_LINES` entry. The list is keyed by file and text, not by line; proved locally that inserting a line above it keeps the gate green and that an unrelated real mention still reds it.

joyful-ii-V-I:
Dispositions for this round. CI is green, 31 of 31.

**The 17 "use the sink form of `DISCLOSE`" findings** (atoms.h, dmm.h, docparse.h, editcheck.h, graph.h, ingest_crawl.h, ingest_parsepool.h, lanes.h, mcpindex.h, packtask.h, prcontext.h, search.h, serialize.h, slice.h, verbs_for.h and the rest): **deferred by design, not an oversight.** This pass renames the vocabulary and lands the sink form; converting call sites to it is the next pass, because each site needs its reason classified against the closed set, and a 179-file rename is the wrong change to bury that judgement in. The one-argument form keeps today's behaviour exactly. The classification table for the sites this pass added is in the lane report, and the conversion is tracked as a follow-up.

**`gitmine.h` ~2994, "keep the mismatch fallback reachable":** a real hazard, but **pre-existing and untouched here.** Lines 2994 and 2999 are byte-identical between `origin/main` and this branch — the same assumption and the same four-lines-later fallback are already on main. The old macro was already an optimizer assumption on both compiler families: the previous `Diagnostics.h` says so itself, with clang using `__builtin_assume` and GCC the `if(!expr) __builtin_unreachable()` idiom, "both compile to zero code, just a hint". `[[assume]]` is GCC's own documented spelling of that same contract, and clang's path is unchanged by the rename. Tracked as a follow-up: sweep for conditions that are assumed and then still tested, since that combination makes the fallback deletable.

**`darkflags.h:787` (preserve root-walk failure state) and `docdrift.h:2589` (`projected_drift="0"` after a mismatch):** both predate this branch; the rename only moved the macro name on those lines. Tracked as follow-ups.

**`test/kotlincheck.sh:496`, require disclosures in every build flavour:** a fair ask for the gate, and a follow-up. It belongs with the sink-form conversion, which is what gives every flavour something to assert.

Issue #225, opened 2026-09-14
wrap/skills install has no discovery path for binaries installed via mise/aqua-registry (or any channel outside the curl installer's layout)


s0undt3ch:
`ripwire` is now installable via mise/aqua-registry (https://github.com/aquaproj/aqua-registry/tree/main/pkgs/redhat-et/ripwire), which fetches the binary out of the release tarball and symlinks it onto `PATH` on its own. It does not reproduce the curl installer's layout contract of `<prefix>/bin/ripwire` + `<prefix>/share/ripwire/skills/`.

`wrapPrintSkillsLine` (`src/wrap.h:200-219`) only has two live discovery paths — (a) `./skills/install.sh` relative to cwd (a checkout), or (b) `<exeDir>/../share/ripwire/skills/install.sh` (the curl installer's staged copy). Any other install channel — mise/aqua, a distro package, a hand-copied binary — falls to case (c): a dead "clone the repo and run `skills/install.sh`" comment instead of a runnable command. `ripwire wrap claude` for a mise-installed binary never wires up skills, silently.

This will also recur on every future upgrade: skills aren't versioned with the binary once it's outside the curl-installer's own directory, so there's no path that keeps skills in sync with `ripwire --version` short of a full re-clone.

**Two directions worth a maintainer opinion on:**

1. **Embed the skills in the binary itself**, generated at build time (CMake step turning `skills/*/SKILL.md` into a baked-in header, ~384 KB uncompressed today), with a new subcommand (e.g. `ripwire skills install [--codex]`) that writes them to the standard agent skill dirs. Installer-agnostic by construction, and "update skills on upgrade" becomes "re-run the subcommand" since the skills are pinned to that exact binary build. Fits the G3 self-contained / G5 zero-dependency-CLI guardrails — no new install-path convention to maintain per channel.
2. **Extend the aqua-registry package** to also stage `skills/`/`hooks/`/`install.sh` next to the binary in a layout `wrap` can find. Weaker: it only fixes mise/aqua, leaves every other current and future distribution channel with the same gap, and aqua's registry format doesn't obviously support "also drop this sibling directory" for a `github_release` asset.

[...]

PR #293, opened 2026-09-19
Embed skills/hooks into the binary; unify activation as `ripwire skills install` (#225)

s0undt3ch:
## Summary

Implements #225: skills and hooks are now embedded into the `ripwire` binary at configure time (a generated, content-addressed header) instead of being staged as separate files a curl/source install had to copy around. Activation is now one command, `ripwire skills install`, which extracts the embedded set into a versioned, immutable-by-hash store and symlinks from there into each agent's discovery directory (Claude Code, Codex, Codex-legacy, Hermes, openclaw). `--doctor` reports install provenance (declared vs. live, manifest staleness against the binary's own embedded version) instead of ever calling a version-manager shim STALE by mtime.

This collapses two previously-separate mechanisms (`skills/install.sh`'s checkout-symlink installer, and the curl installer's `$prefix/share/ripwire/{skills,hooks}` staging step) into one binary-owned path, per the design already reviewed and approved by the maintainers in the #225 discussion.

## What changed

- **Embedding**: `CMakeLists.txt` globs `skills/` and `hooks/` at configure time into `generated/embedded_skills.h` (byte-identical across two configures, `CONFIGURE_DEPENDS` on both the glob and each file, gated by `test/selfcontainedcheck.sh`).
- **`ripwire skills install`**: extracts into `$RIPWIRE_DATA_HOME/{skills,hooks}/<version>-<hash8>/` (`O_CREAT|O_EXCL|O_NOFOLLOW`, atomic rename into place, immutable by content hash — a killed run can never leave a directory that reads back as "already extracted"), then symlinks into the destination agent's skill directory. Refuses to write through or replace a planted symlink or real file/directory even under `--force`; a dangling or foreign-but-live symlink is repaired; a link-loop failure is now counted and reported, never silent.
- **Codex `--hook` merge**: ported `skills/install.sh`'s pre-#225 `install_codex_hook` — PreToolUse nudge + SessionStart primer + UserPromptSubmit router merged into `${CODEX_HOME:-~/.codex}/hooks.json` in one `jq` call (Codex has one install channel, matched by exact command, unlike Claude's `isScript` basename match).
- **Claude's `--hook`**: now also registers the UserPromptSubmit prompt router (`install_claude_route`, previously never ported to the C++ path) and emits the substitution-meter disclosure banner.
- **`--hermes`**: now collects `skills/hermes/ripwire-*/` (Hermes-native content, e.g. `ripwire-repo-map`) in addition to the flat `ripwire-*` set, which it previously linked exclusively.
- **`pruneStale`**: no longer limited to entries the *previous* manifest tracked — it also scans the destination directory directly for `ripwire-*` symlinks outside the current set (a stray or dangling one, left behind by a hand-removed skill or an older installer), while still refusing anything outside its `ripwire-*` scope of authority.
- **Store integrity**: extraction no longer trusts the store directory's name (a content-hash key) alone; a byte-for-byte compare against the embedded reference catches a planted symlink, a partial extraction a killed run left behind, or on-disk tampering, and re-extracts from scratch when it doesn't match.
- **No `/tmp` fallback**: an unset `HOME` is now refused everywhere instead of silently redirecting writes to `/tmp`; `settings.json` and the manifest are written via temp-file-plus-atomic-rename instead of truncate-in-place, so a crash mid-write can no longer leave either empty.
- **`--doctor`**: skills rows carry `not_installed=`, `stale=` (manifest source vs. the binary's own embedded version), and a repair hint that names what's actually blocking (an untracked entry, not a command that can't fix it). The top-level binary-path row and the agent-specific rows both recognise a mise/aqua shim by location and compare byte-for-byte instead of by mtime.
- **`wrap`**: the printed `skills install` recipe line is shell-escaped (`rw::shSingleQuote`) — a resolved binary path containing `$(...)` or a backtick is now inert when pasted, not executable.
- Docs (`INSTALL.md`, `README.md`) and the release/source installers (`scripts/install.sh`, `install.sh`) updated to match; the curl installer's separate `$prefix/share/ripwire/skills` staging step is gone, replaced by the same `ripwire skills install --all` call the source build uses.

## Known, deliberately out of scope for this PR

- Windows is out of scope (blocked on the separate native port, #44).

## Testing

Two full whole-branch adversarial review rounds (building the binary, running the full `pargates` suite, and driving hand-built fixtures against `--doctor`/`skills install` under hostile environments — foreign symlinks, corrupt JSON, symlinked settings files, no `jq` on PATH, ambient env-var leaks) found and closed 3 Critical + several dozen Important/Minor findings across both rounds, including: an unrepairable v1→v2 manifest upgrade path, a manifest that recorded intent instead of outcome, an unescaped shell-injection vector in the printed recipe, two gates that were writing into the operator's real Codex/Hermes config during CI-style test runs, and several assertions that had silently stopped testing anything after an earlier refactor changed the output they grep for.

Rebased onto current `main` for this PR; the rebase surfaced (and this PR fixes) 18 raw-POSIX-call sites and one throwing-iterator use that predated `main`'s portability lane, caught by `osswitchcheck.sh`/`hazardpatterncheck.sh` once replayed against it.

The release matrix subsequently caught nine gates red across all 24 shards, tracing to several slices of #225's own plan left unfinished at that point (Codex `--hook`, Claude's route registration, Hermes-native skills, prune's dangling-symlink scan) — all fixed and covered by gates in this PR. A further maintainer review round added store-integrity verification, removed the `/tmp` fallback, made the manifest/settings writes atomic, tightened `pruneStale` to an ownership check, and split an over-complex `codexdoctor.h` function; a flaky idempotent-store test arm (comparing `stat -f`'s GNU filesystem-info output instead of falling back to `stat -c` as intended) was also fixed.

Full gate suite (661 gates) passes except two known, pre-existing/environmental flakes unrelated to this diff (`legendcoveragecheck.sh`, red on unmodified `main` too; `pargatescheck.sh`'s timing probe, which only fails under `-j6` parallel load and passes standalone).

## Disclosure

This PR was developed with substantial assistance from Claude Code (Anthropic's AI coding agent), including implementation, the adversarial review rounds described above, and this description. All work was reviewed and directed by a human throughout.


s0undt3ch:
This change is part of the following stack:

- #293 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

joyful-ii-V-I:
Thank you, @s0undt3ch — this has moved a lot since the last pass, and it moved in the right direction. The four
features `skills/install.sh` used to own — the Codex `--hook` merge (including the exact PreToolUse/SessionStart/
UserPromptSubmit timeouts and limits the old script used), the Claude `UserPromptSubmit` router, the opt-in
announcement/meter-disclosure banner, and Hermes-native skill linking — are all ported now and read like faithful
ports, not rewrites. Every raw POSIX call in `skillsinstall.h` now goes through `rw::os`, matching the seam
discipline the rest of the tree adopted. `skillsinstallcheck` (18 arms), `wrapverbscheck` and `selfcontainedcheck`
(including its skills+hooks embed cross-check) all pass clean in our own rebuild, and CI is green across all 32
jobs. The Ubuntu idempotent-store flake is fixed at the actual root cause — a `stat -f`/`stat -c` GNU-coreutils
trap, the same shape `cachehashcheck.sh` already documents — not papered over.

What's left is smaller than it was, and mostly concentrated in `skillsinstall.h` itself.

## Needed before merge

1. **Check the skill store's contents instead of trusting its name.** `extractGroup` (skillsinstall.h:142) still
   reads `if( std::filesystem::exists( storeRoot, ec ) ) { return { true, {} }; }` — no check that `storeRoot` is a
   real directory (not a link, not a partial extraction) and no byte comparison against the embedded files.
   **Fix:** `lstat` the store root and require a real directory; byte-compare each embedded file against
   `storeRoot/<relativePath>` (opened `O_NOFOLLOW`, size + `memcmp` — no hashing needed, the bytes are already in
   the binary). On a mismatch, re-extract to a fresh temp dir and rename over, or refuse with a disclosed reason.
2. **Don't fall back to `/tmp` when `HOME` is unset.** `homeDir()` (skillsinstall.h:61) is still
   `envOr( "HOME", "/tmp" )`. Refuse with a clear message when neither `HOME` nor `RIPWIRE_DATA_HOME` is set;
   `VALIDATE` that `RIPWIRE_DATA_HOME`/`CLAUDE_CONFIG_DIR`/`CODEX_HOME` are absolute.
3. **Write `settings.json` and the manifest atomically.** `skillsinstall.h` has zero calls to
   `rw::pathguard::createExclTempFile`/`randomTempSuffix` — `ingest_cache.h`, `quality.h`, `gitoracle.h` and
   `mcpedit.h` all use that helper already; this file still truncates-then-writes.
4. **Scope the symlink prune to our own links.** `pruneStale` (skillsinstall.h:288) correctly requires
   `S_ISLNK` before unlinking, but doesn't `readlink` the target and require it under `dataHome()/skills/` — a
   user's own unrelated symlink happening to be named `ripwire-<something>` is still fair game for deletion.
5. **Remove `.superpowers/sdd/embedded-skills-implementation-plan/fix-wave-2-report.md`.** Still in the tree.
6. **`--quality-delta` should gate 0** (we ran `./build/ripwire . --quality-delta=origin/main..HEAD` ourselves —
   rc=2, `gating="6"`):
   - `codexdoctor.h::skillsCheck` complexity 3→26 (bar 15) — still one function with the manifest-parity check,
     the staleness check and a 3-way outcome `if/else if/else` all inlined.
   - `envOr` is still byte-identical between `codexdoctor.h:34` and `skillsinstall.h` — move one copy to a shared
     header.
   - a *new* duplication pairing appeared: `notes.h::sortNotes` / `skillsinstall.h::symlinkOrRefuse`. This is not
     the same pairing your existing ack covers (`symlinkOrRefuse | extentsuspect.h::inSet`) — that ack won't
     silence this one; it needs its own ack or its own fix.
   - `installForAgent` (complexity 46) and `runSkillsInstall` (complexity 59) are both still one function each,
     ~120-150 lines. Splitting the per-entry decision out of `installForAgent`, and the argv-parsing / `--all`
     loop out of `runSkillsInstall`, should bring both under the bar.
   - one more gating row showed up that doesn't look like yours: `flipimpact.h::buildSiteLocator` /
     `notes.h::buildNoteIndex` duplication — neither file is touched by this PR. We think this is baseline
     attribution noise from the two trains that landed since your branch point, not something to chase on your
     side; we'll sort it out at train time.
7. **Self-checks, per CONTRIBUTING §3 "Self-check, don't throw."** Still none in `skillsinstall.h`. Good spots:
   `VALIDATE` the manifest's `version`/`skill=` shape and the environment paths that must be absolute; `EXPECTS`
   non-empty bytes in `writeStoreFile`; `ENSURES` a non-empty name out of `skillDirNameOf`; `DISCLOSE` the silent
   degrade paths (a lost rename race, a skipped foreign entry, an ambiguous manager layout).

## Worth doing

8. **The `--doctor` mise/aqua resolvers still look for a layout that doesn't match real installs.** We checked
   directly against your own `wrapverbscheck.sh` fixtures: `resolveMiseLayout` expects `<version>/bin/ripwire`,
   but the fixture builds `installs/ripwire/<version>/<archive>/ripwire` (no `bin/`); `resolveAquaLayout` expects
   `pkgs/redhat-et/ripwire/<version>/...`, but the fixture (and real aqua) nests under
   `pkgs/github_release/github.com/redhat-et/ripwire/...`. Both would report `managed_unverified` for every real
   install today.
9. **Create store directories with explicit modes (0755)** — every `create_directories` call in
   `skillsinstall.h` still relies on the umask.
10. **jq's stderr is already passed through** (in-flight, since `popen` only redirects stdout) — the code comment
    claiming it goes to `/dev/null` is stale and worth a one-line fix. The other half of the old ask still stands:
    jq is invoked as a bare `"jq"` token, resolved by the shell's `PATH`, never pinned to an absolute path.
11. **Usage text**: bare `ripwire skills` still maps the `skills/` directory as a repo (exit 0, dumps the corpus
    map) instead of printing usage; `skills install --help` still returns rc 2 ("unknown flag --help"); `--help`
    still doesn't mention `skills install`, so `docs/COMMANDS.md` never regenerates it.
12. **The "N skill(s) linked" summary** should only print once the links are verified — this falls out of item 1.
13. **`selfcontainedcheck`'s new hooks/ cross-check is a good addition but isn't recursive** (`glob`, not
    `rglob`) — a `hooks/lib/x.sh` would still slip past it unembedded and silent.

## Small items (optional)

- `FdGuard` (skillsinstall.h) still duplicates `rw::pathguard::OwnedFd`.
- `runShellCapture`'s `popen` still has no RAII holder.
- `shimBinaryCheck` still reports `ok=1` when `selfPath` is empty (codexdoctor.h:211) — should say unverified.
- This repo doesn't use `changelog.d/` fragments (that wasn't accurate in the last round) — it keeps one
  `CHANGELOG.md`. There's no `[Unreleased]` entry there yet for embedding skills/hooks into the binary; worth
  adding one in the PR's own voice, the way other entries in that file read.

## What's next

We had to rebuild `main` twice since the last round (two integration trains landed), so a merge of current `main`
into this branch will now pick up conflicts in `README.md`, `docs/EVALS.md`, `present/deck5_ripwire_build.js` and
`test/regression.sh` — all four are mechanical (gate-count/doc bookkeeping that regenerates on every landing), not
real conflicts, and we're happy to help resolve them when this rides a train. Once the items above are closed and
CI + a fresh review are green, this lands through our normal integration-train process, not a direct merge. Thanks
again for sticking with this — it's a substantial contribution and it's in good shape. Happy to answer anything
here as you go.

joyful-ii-V-I:
**Status, so this is not waiting on us silently.**

This is **not** a 0.6.2 blocker — 0.6.2 is cut on the graph-completeness, byte and Windows work — and it is **high priority for 0.6.3**. The refreshed review above still lists six items, the load-bearing ones being: verify the store's contents rather than trusting its name, atomic writes for `settings.json` and the manifest (the helper already exists in-tree), scoping the symlink prune to links we own, and self-checks in the ~1,000 new lines per CONTRIBUTING §3.

Two things changed underneath you while this sat, both worth a rebase rather than a surprise: the gate roster moved (several trains landed, including a new git-environment hermeticity gate and a centralised `stat` helper that your own PR #298 work seeded), and `kParserVer` is now 119.

No rush from our side, and nothing here is a criticism of the direction — the design is the one agreed in #225 and the parity ports read as faithful. Ping us when you push and we will re-review promptly.

s0undt3ch:
Pushed — rebased onto current main (through train-15, kParserVer 119; picked up the MSVC cl.exe
seam and the osswitch arm H hardening along the way).

On the four load-bearing items from the last review round:
- **Store contents verified, not trusted by name** — `storeContentsMatch()` byte-compares every
  embedded file against the store copy before extraction is skipped as "already done"
  (src/skillsinstall.h:180).
- **Atomic writes for settings.json and the manifest** — both go through
  `rw::pathguard::createExclTempFile`/`commit()` (src/skillsinstall.h:386, :842).
- **Symlink prune scoped to links we own** — `pruneTargetIsOurs()` requires manifest-tracking or a
  target resolving under our own store root before unlinking an untracked name
  (src/skillsinstall.h:396).
- **Self-checks per CONTRIBUTING §3** — VALIDATE/EXPECTS/ENSURES at the manifest-parse and
  environment-path boundaries.

Also folded in since: usage text for bare `skills`/`skills install --help`, the mise/aqua layout
fix so `--doctor` resolves a real install instead of reporting `managed_unverified`, 0755
directory creation regardless of umask, jq resolved via an actual PATH walk, `shimBinaryCheck`
no longer claiming a comparison it never ran, FdGuard/OwnedFd deduped, a RAII guard on the popen
handle, and a CHANGELOG entry. Full gate suite (651 gates) green modulo the same
pre-existing/environmental flakes noted before (the `scan-skills` legend-coverage gap, a couple of
timing-sensitive gates under parallel load).

Ready for re-review whenever you have a moment.

---
*Posted by Claude Code on behalf of @s0undt3ch*

joyful-ii-V-I (commented):
Thank you, @s0undt3ch. This round closes most of what we asked for, and we checked it by running fixtures against a
build of `436590f9`, not only by reading the diff:

- **HOME**: unset or empty HOME now refuses with a clear message and writes nothing (no `/tmp/.local`, no
  `/tmp/.claude`). A relative `RIPWIRE_DATA_HOME` or `HOME` refuses too, and the `VALIDATE` trace names the site.
- **Atomic writes**: we fed `settings.json` malformed JSON, then made its directory read-only, then SIGKILLed 20
  `--hook` runs at random points. `settings.json` stayed valid JSON every time, and no temp files were left behind.
  A 0600 file keeps its mode.
- **Store verification** works for a real directory: a truncated `SKILL.md` in the store is detected and
  re-extracted.
- **Prune**: a user's live `ripwire-mine` symlink, a real file and a real directory with `ripwire-` names, and a
  non-`ripwire-` link all survive. A dangling `ripwire-ghost` is pruned.
- **Modes**: under `umask 000` the store dirs are 0755 and the hooks are 0755.
- The `.superpowers/` notes are gone, self-checks are in, `kParserVer`/`kCacheVersion` are untouched, the usage
  text is fixed, jq is resolved from absolute PATH entries only, and `hooksembedrecursivecheck` pins the nested
  hook case. Nice work.

Three things are still needed before this can land, and all three are small:

1. **A symlink or a regular file at the store key is still trusted.** `extractGroup` only clears `storeRoot` when
   it is a real directory. When it is a symlink or a file, `rename(tmp, storeRoot)` fails with ENOTDIR, and the
   fallback `std::filesystem::exists( storeRoot )` follows the link and returns ok. We reproduced both cases from
   the first round on this head:
   - A symlink `~/.local/share/ripwire/skills/<key>` pointing at another directory: rc=0, "16 skill(s) linked",
     and `~/.claude/skills/ripwire-orient/SKILL.md` serves that directory's text.
   - A regular file at the key: rc=0, "16 skill(s) linked", and every link dangles.

   **Fix:** `lstat` the key first. If it is anything other than a real directory, `unlink` it before the rename. If
   the rename still fails, re-check with the same `lstat` + `storeContentsMatch` pair instead of `exists()`. A
   fixture arm for each case in `skillsinstallcheck.sh` would pin this.
2. **`--quality-delta` gates 1.** Running `./build/ripwire . --quality-delta=60b65f02..HEAD` gives rc=2. The
   finding is `duplication extentsuspect.h::inSet | skillsinstall.h::symlinkOrRefuse`, tokens=32. Your ack
   `a588e57c…` records 30, so it no longer matches. Re-ack with the emitted row, or fold the one-line wrapper away.
3. **The thin `skills/install.sh` can't find the binary next to it in a release archive.** `find_ripwire` checks
   `build/ripwire`, `build-release/ripwire`, then PATH. The release tarballs and the new Windows zip ship
   `ripwire` (or `ripwire.exe`) at the archive root, beside `skills/install.sh`. So `bash skills/install.sh` from an
   unpacked archive fails with "no built or installed ripwire binary found". We reproduced this with a
   tarball-shaped directory. Please probe `"$here/ripwire"` and `"$here/ripwire.exe"` first.

   On Windows there is a second gap: `os::symlink` returns ENOSYS. Train 19, which is about to land, adds a
   `windows-package` CI step that runs `bash skills/install.sh` (and `--codex`) from the unzipped folder under
   Git Bash. It also adds a README line telling Windows users to do the same, and it requires at least 10 readable
   `SKILL.md`s. On Windows, a copy fallback (copy the store file where the link would go, and record it in the
   manifest as ours) would keep that step green. Otherwise the install needs to refuse with a clear message and
   the README line needs to change.

Smaller items. None of these blocks the merge on its own; take them now or in a follow-up:

- A relative `CLAUDE_CONFIG_DIR` (or `AGENTS_HOME`) is still honoured for the *skills* destination.
  `resolveSkillsRoot` doesn't check that the path is absolute, so `CLAUDE_CONFIG_DIR=relcfg` installed into
  `./relcfg/skills`, rc=0. One `VALIDATE( skillsDest.is_absolute() )` in `installForAgent` closes it (`--hook`
  already refuses).
- `pruneTargetIsOurs` compares string prefixes. A target under `~/.local/share/ripwire/skills-mine/` counts as
  "ours". Compare path components, or append a `/` to the root.
- `storeContentsMatch` checks every embedded file, but not extra files in the store or a symlinked intermediate
  directory. For example, `skills/ripwire-orient -> elsewhere` holding identical bytes plus an extra `evil.md` is
  accepted, and the agent sees `evil.md`. Low risk, since it needs write access to your data home, but a listing
  check would make "verified" complete.
- The first-ever manifest, and a newly created `settings.json`, are created 0666 & ~umask. Consider 0644.
- A `.tmp-<pid>-<key>` directory left by a killed run with a different pid stays forever. Old store keys also
  accumulate.
- Carried over from `skills/install.sh`, so not something this PR introduced: when `settings.json` is a symlink
  (a dotfiles setup), the atomic rename replaces the link with a regular file, and the dotfiles copy never gets the
  hook. The hook `command` is also written unquoted, so a data home with a space in it gives a hook that can't
  run. We'd take either as a follow-up.

**How it lands.** Train 19 (#332) carries #328's hook fixes, and that interacts well with this PR: `hooks/` is
embedded at configure time, so a merge picks up the new `ripwire-nudge.sh`/`-route.sh` bytes (new store key)
with no edits on your side. We trial-merged `436590f9` onto train 19. The only conflicts were the gate count
(648/651 → 652 in README/EVALS/deck), the `regression.sh` gate list (a union; train 19 adds `rubyattrscheck`),
and `printf_parity.manifest`'s `help_all` row, which needs regenerating because both sides changed `--help`. It
built with no warnings.

Once train 19 merges, please rebase onto `main` and push the three fixes above. We'll re-run the fixtures and
gates and then carry the PR in the next integration train as-is, with your commits kept. No squash needed.

Thanks again. The design is right, and what's left is edge cases.

PR #298, opened 2026-09-19
test: centralize the GNU-vs-BSD stat trap; close a hermetic-isolation gap

s0undt3ch:
## Why this PR exists

While chasing an unrelated CI failure on #293, I found that `stat -f 'FMT' FILE 2>/dev/null || stat -c 'FMT' FILE` — a one-liner meant to be portable across GNU coreutils and BSD/macOS — silently does the wrong thing on Linux. GNU's `-f` is a *different, valid* flag (report on the filesystem, not the file), so it succeeds with a filesystem-info block instead of failing, and the `-c` fallback never fires. The caller ends up comparing junk (block/inode counts that legitimately drift under concurrent I/O) instead of the field it actually asked for.

That exact trap had already been independently rediscovered and fixed, by hand, in **twelve separate gate scripts** — each with its own near-identical comment explaining the same lesson. This PR centralizes the fix instead of leaving a thirteenth copy waiting to happen.

Separately, while auditing for a related class of bug (env vars read before `HOME`, capable of leaking a real write into the operator's machine — the same failure mode as a documented past incident), I found `codexpromptroutecheck.sh`'s `skills install --codex --hook` calls override `HOME`/`CODEX_HOME`/`AGENTS_HOME` but not `RIPWIRE_DATA_HOME`. That's inert on `main` today (the current shell installer doesn't read it), but becomes a real leak once the embedded-skills feature (#225) lands and starts reading it. Fixed proactively, in the same PR, since it's the same category of gap this cleanup is already touching.

## What changed

- **`test/lib/statcompat.sh`** (new): detects GNU vs. BSD/macOS `stat` once, provides `mtime_of`/`inode_of`/`mode_of`/`size_of`/`inode_mtime_of` as ready-to-call functions. Callers just source it — no per-file branching on the flavour.
- 12 gates (`cachehashcheck.sh`, `cachesplitcheck.sh`, `clonecachecheck.sh`, `evictioncheck.sh`, `g1freshcheck.sh`, `headsnapcachecheck.sh`, `mcpeditmodecheck.sh`, `portablecachecheck.sh`, `prcontextcheck.sh`, `qsnapcachecheck.sh`, `qsnapprefetchcheck.sh`, `statgatecheck.sh`) source it instead of hand-rolling their own detect-once block. Two of them keep a one-line local wrapper for their own caller-specific fallback string (`|| echo 0`, `|| echo "MISSING"`).
- `mcpeditmodecheck.sh`'s locally-named `file_mode` was the exact same function as `mode_of` under a different name — call sites renamed rather than keeping two names for one thing.
- `statgatecheck.sh` had a dead, uncalled `mtime_ns` helper (a leftover from before this cleanup) — removed rather than migrated.
- **`test/lib/clean-env.sh`** (new): unsets the agent-home-relocating env vars (`CODEX_HOME`/`AGENTS_HOME`/`HERMES_HOME`/`CLAUDE_CONFIG_DIR`/`RIPWIRE_DATA_HOME`) before a gate varies `HOME=` per invocation. `codexpromptroutecheck.sh` now sources it and passes `RIPWIRE_DATA_HOME` explicitly at its `skills install --codex --hook` call sites.

## Verification

- Full `test/pargates.py` run, twice, plus once more under added CPU/I-O contention (20 `yes` processes + 10 concurrent gate copies) — the exact conditions that made the original stat bug reproducible in the first place.
- `--quality-delta` gating=0, determinism (`diff -q` on two runs) clean, `xmllint --noout` clean.
- Three gates fail under `-j6` parallel load on this machine regardless of this branch (`legendcoveragecheck.sh`, `pargatescheck.sh`, `rootrelemitcheck.sh`) — all pass individually, confirmed pre-existing/environmental (reproduced identically on unmodified `main`), not touched by anything in this diff.

## Summary by CodeRabbit

* **Tests**
  * Improved portability of test checks across GNU/Linux and BSD/macOS environments.
  * Centralized file metadata handling for timestamps, inodes, permissions, and sizes.
  * Added shared environment cleanup to make integration tests more isolated and reliable.
  * Expanded installer-hook validation to confirm repeated runs register exactly one hook.
  * Removed obsolete, duplicated platform-detection logic from individual tests.


s0undt3ch:
This change is part of the following stack:

- #298 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

joyful-ii-V-I:
Thanks for tracking this down — the coreutils `-f`/`-c` disagreement is a genuinely nasty trap (it fails
by *succeeding* with the wrong data, which is worse than a clean error), and centralizing it once other
scripts can just source is the right shape of fix.

**What's good, with evidence**

- `test/lib/statcompat.sh` is a faithful, function-for-function extraction of the 12 gates' own
  detect-once blocks — I diffed each migrated gate's old inline block against the new shared function
  and they're identical logic, just one copy instead of twelve. All 13 touched gates
  (`cachehashcheck`, `cachesplitcheck`, `clonecachecheck`, `codexpromptroutecheck`, `evictioncheck`,
  `g1freshcheck`, `headsnapcachecheck`, `mcpeditmodecheck`, `portablecachecheck`, `prcontextcheck`,
  `qsnapcachecheck`, `qsnapprefetchcheck`, `statgatecheck`) pass green against this branch's binary.
- The `mcpeditmodecheck.sh` `file_mode`→`mode_of` rename and the `statgatecheck.sh` dead `mtime_ns`
  removal both check out — `mtime_ns` has zero remaining references after the diff, confirmed by grep.
- `manifestcheck.sh`, `gateexitcheck.sh`, `gatecountcheck.sh` all rc=0 — no gate-registration fallout,
  as expected since no new *gate* is added (the two new `test/lib/*.sh` files are sourced helpers, the
  same shape as the existing `cxxflags.sh` / `gatehttp.sh` / `headbinlib.sh` in that directory).
- `--quality-delta` against this PR's actual base (its merge-base with main, 92bc055e — see below for
  why that's the number that matters) is clean: 0 regressions, 0 gating.
- The `RIPWIRE_DATA_HOME` env-leak fix for `codexpromptroutecheck.sh` is honestly scoped in the PR body
  itself ("inert on main today... becomes a real leak once #225 lands") — I confirmed `RIPWIRE_DATA_HOME`
  has zero references anywhere in `skills/install.sh` or `src/` today, so the fix is exactly the
  proactive, low-risk thing it's described as.

**Before merge**

1. **The stat-trap sweep missed four gates that have the identical pattern.** `cacheisolationcheck.sh`
   (line 33), `qsnapproducercheck.sh` (lines 221-223), `sidecarsymlinkcheck.sh` (lines 537-540), and
   `tempfilesymlinkcheck.sh` (lines 66-68) all hand-roll the same `if stat --version; then stat -c ...;
   else stat -f ...; fi` detect-once block this PR extracts from the other twelve — correct code, but a
   fifth, sixth, seventh, eighth copy of it, which is exactly what this PR's own description says it's
   eliminating. Worth folding into the same cleanup (or a fast follow) so `statcompat.sh` is genuinely
   the one place, not one of two. (I also checked `hookcheck.sh`'s `stat -f %m ...` hit — that one's a
   literal test-fixture string fed to a classifier, not live gate logic, so it's not a miss.)

2. **The hermetic-isolation fix doesn't cover the gate most exposed to it.**
   `claudeconfigdircheck.sh` exists specifically to test `CLAUDE_CONFIG_DIR` relocation, including an
   "UNSET IS UNCHANGED" arm (G) that deliberately calls `skills/install.sh` with `HOME=` set but
   *no* `CLAUDE_CONFIG_DIR=` override — relying on the variable being absent from the environment. I
   reproduced the exact failure mode this PR's write-up warns about: with `CLAUDE_CONFIG_DIR` exported
   ambiently (simulating a developer whose real Claude Code config is relocated, which is precisely what
   this gate is testing the relocation *for*), running `claudeconfigdircheck.sh` unmodified wrote 19
   real files (`settings.json`, 15 skill dirs, a manifest) into that ambient directory, and the gate
   itself then went red (4 FAILs) because its own "unset" baseline run absorbed the same ambient value.
   That's a live leak today, not a future one. Since this PR already introduced `test/lib/clean-env.sh`
   for exactly this purpose, sourcing it at the top of `claudeconfigdircheck.sh` (before its `HOME=`-only
   calls) looks like the natural fix — worth doing in this PR or a same-day follow-up, since it's the
   same root cause the PR already fixed once.

   Two smaller instances of the same shape, lower priority: `skillinstallcheck.sh` already has its own
   ad hoc `unset CODEX_HOME AGENTS_HOME HERMES_HOME` (line 22) predating this PR — functionally the same
   job as `clean-env.sh` but missing `CLAUDE_CONFIG_DIR`/`RIPWIRE_DATA_HOME` and not consolidated onto
   it; and `hermesinstallcheck.sh`'s `HOME="$CLAUDE_HOME" bash "$SK/install.sh"` call (line 79) is
   unguarded for `CLAUDE_CONFIG_DIR`.

**Smaller items**

- Worth a one-line CHANGELOG entry naming the new `test/lib/statcompat.sh` / `clean-env.sh` helpers and
  the fixed gates, if this project's CHANGELOG covers test-only changes — not blocking either way.
- No behavior/build/runtime changes here (test-only), so `--quality-delta` against the PR's own base is
  the right gate and it's clean; just make sure whoever runs the train-level `--quality-delta` uses the
  merge-base, not a blind `origin/main..HEAD`, or the 100-commit gap will paint this PR with train 8-11's
  unrelated src/ movement (see above).

**What happens next**

This rides the next integration train once the two "before merge" items are addressed (or explicitly
deferred with a reason) — no PR of your own to open, the orchestrator batches it in.

---

## Detail for the orchestrator (not part of the posted text)

### Gates run (all against this PR's binary, `$PWD/build/ripwire`)

| gate | rc |
|---|---|
| cachehashcheck | 0 |
| cachesplitcheck | 0 |
| clonecachecheck | 0 |
| codexpromptroutecheck | 0 |
| evictioncheck | 0 |
| g1freshcheck | 0 (SKIP — asan/ not configured in this worktree, expected) |
| headsnapcachecheck | 0 |
| mcpeditmodecheck | 0 |
| portablecachecheck | 0 |
| prcontextcheck | 0 |
| qsnapcachecheck | 0 |
| qsnapprefetchcheck | 0 |
| statgatecheck | 0 |
| manifestcheck | 0 |
| gateexitcheck | 0 |
| gatecountcheck | 0 |
| claudeconfigdircheck (ambient CLAUDE_CONFIG_DIR probe, see finding 2) | 1 (RED, as expected — reproduces the leak; PASSES clean with no ambient var set) |

### quality-delta

- `--quality-delta=origin/main..HEAD`: `gating="12" preexisting-worse="12"`, all 12 findings in src/
  files this PR does not touch (compactlegend.h, dynamic_map.hpp, testmap.h, crossref.h, mergescout.h,
  flipimpact.h, gitoracle.h, notes.h, renamemine.h, slice.h, taskroute.h) — a measurement artifact of the
  PR's 100-commit-stale base, confirmed by re-running against the true merge-base.
- `--quality-delta=92bc055ee3550ef695308c0b6ed7d48414e9b986..HEAD` (the PR's actual merge-base with
  origin/main): `regressions="0" minor="0" preexisting-worse="0" gating="0"`. This is the number that
  actually describes what the PR did.
- `git status --porcelain --untracked-files=no` empty; `target_ref` == `git rev-parse HEAD` ==
  af88bfdab3f9f83edd0bc4fce68ab77c223de6fb in both runs.

### Train interaction (trains 8-11: compact-legend-default, MCP affected/rank_by, gate-list churn)

- No textual conflict: `git diff --name-only 92bc055e..HEAD` (PR's own changes) touches 15 test/ files;
  none overlap the lines trains 8-11 changed in the two files they share
  (`prcontextcheck.sh` — trains added `--legend=full` to several invocations for the new compact
  default, in the OUT=/NTOUT=/PRC_FAULT lines; the PR's own edit is near the top, adding the
  `statcompat.sh` source line and switching `file_mode`→`mode_of`-equivalent reads — different regions,
  clean merge) and `sidecarsymlinkcheck.sh` (trains added `--legend=full` to `readArchCheck`; the PR
  doesn't touch this file at all — it's one of the four missed stat-trap gates, see finding 1).
- `test/regression.sh`'s single-line gate-absorb loop grew by three names on main since the PR's base
  (`forhdrshapecheck`, `forsectioncollapsecheck`, `notesdegradecheck`) — the PR doesn't edit that line,
  so it's a non-issue for merge, but it's more evidence the PR is meaningfully behind main and should
  rebase before landing rather than merge blind.
- No semantic conflict found: the PR's changes are pure test-infrastructure refactors orthogonal to the
  legend-default/MCP-tool changes trains 8-11 shipped.

## Dogfood gaps

- `--for="GNU BSD stat compat"` / `--grep=stat -f` were not tried — this is a test/*.sh-only PR (shell,
  not one of ripwire's indexed source languages for symbol-graph purposes), so `--for`/`--expand`/
  `--callers`/`--impact` don't apply to gate scripts the way they do to src/. Fell straight to `grep -rl`
  and `git diff`/`git log` for the whole review, which is the correct tool for shell-script diffing, not
  a fallback from a verb that should have worked. Logging as a genuine "N/A for this artifact type"
  rather than a gap: ripwire's own graph has no notion of bash function call edges across `test/*.sh`
  and `test/lib/*.sh`, so "who sources `mode_of`" had no ripwire verb to ask — `grep -rl` was the only
  route to the four missed gates in finding 1. That's a real product gap if this repo's own gate-script
  reuse is a class of question worth answering (bash isn't in ripwire's language list per CLAUDE.md).

s0undt3ch:
@joyful-ii-V-I thank you for the review. It's ready for another go.

joyful-ii-V-I:
Hi @s0undt3ch — this was in great shape by the time we got to it; you'd already pushed a follow-up
addressing both items from the review (the four more `statcompat.sh` gates, and the
`claudeconfigdircheck.sh`/`skillinstallcheck.sh`/`hermesinstallcheck.sh` hermetic-leak closes) before we
started, so we mostly verified and carried your own two commits across.

We did extend the sweep one step further, per the task: past the gates the review named,
`hookcheck.sh`, `routehookcheck.sh`, `agenttablecheck.sh`, `codexinstallhonestycheck.sh`,
`meterdisclosurecheck.sh`, and `releaseinstallcheck.sh` all call an installer varying only `HOME=`,
leaving `CLAUDE_CONFIG_DIR` (and friends) free to leak the same way `claudeconfigdircheck.sh` did — we
reproduced that exact scenario against `hookcheck.sh` (real files written into an ambient
`CLAUDE_CONFIG_DIR`, then the gate failing its own sandboxed-path assertion) and closed all six with
your own `test/lib/clean-env.sh`.

Landed on `lane/t13-contrib-finish` with your authorship preserved on both original commits. Thank you
for the centralization — `statcompat.sh` is genuinely the one place now, not one of two.

Issue #289, opened 2026-09-18
--expand=SYM buries the requested body behind a full ranked map when the name is ambiguous and its file exceeds the pack budget


mariadb-KyleHutchinson:
## Summary

`ripwire --expand=SYM` is designed to save an agent from reading a whole file just to get one symbol's body. That works well when `SYM` resolves to exactly one match — `exactNameExpandDefault` (`main.cpp:2046-2052`) forces `mapTopK=0` and the response is body-only.

But when `SYM` matches **more than one** definition *and* the containing file exceeds `--pack-budget-bytes` (default 64KB, `cli.h:35`), the response degrades to `mode="bundle"` and the ordinary default map (up to 200 symbols, whole-repo ranked) rides along — `main.cpp:2280-2288` even discloses this on stderr:

> `ripwire: note — the ranked top-{N} map rides along with your requested bodies; add --top-k=0 for the bodies alone (or --top-k=1 for a minimal map)`

The problem: the map is emitted *before* the requested `<bodies>` (main.cpp's §H7 appended-sections block, `main.cpp:2407-2412`, runs after the map serialization at `main.cpp:2337-2368`). On a real ambiguous name, the actual payload ends up **75-85% of the way through the document**, and the escape hatch is stderr-only — invisible to anything reading only stdout (the common case for a tool-calling agent).

## Repro (ripwire's own source, no external code needed)

```
$ ripwire . --expand=write   # matches src/pathguard.h and src/serialize.h
```

- Document size: 25,501 bytes.
- The real `<bodies shown="2" total="2" capped="0">` tag starts at byte **21,679** — **85%** into the document.
- The map (`<r ...>` with `shown="200"`, ~10.4K est_tokens) occupies bytes ~900-21,679.
- The escape hatch (`--top-k=0`) is only ever printed to stderr.

We also independently reproduced this against an unrelated, large proprietary C++ codebase, across three different real symbols that each happened to be ambiguous names in oversized files — same `mode="bundle" reason="whole-file ...B over pack-budget ...B"` root, same body-buried-at-75-85%-through-the-document shape each time. This is also what caused a real agent session to give up on `--expand` for a specific-symbol lookup and fall back to plain grep+`Read` — it never noticed the body was in the response at all.

## Proposed fix (proof-of-concept described below, not a polished PR)

Two independent, low-risk changes:

[...]

PR #302, opened 2026-09-20
docs(readme,skills): the --top-k/--for inertness trap, first-run anti-patterns, and what ripwire does not replace

llvm-x86:
## Why

Field report from a first-time agent session using ripwire against a 759-file Flask repo: the binary was fine (cold parse ~0.8 s); the friction was discovering the right verb/flag shape. Three of the reported traps trace to documentation gaps this PR closes.

## What changed

1. **README budget item scoped.** It said `--top-k=N` caps the rows with no qualifier — but on `--for` the flag is read by *nothing*: the run prints the stderr note (`--top-k is not read by --for`) and emits the full bundle anyway, so the agent believes it narrowed the output and did not. The item now names the flags `--top-k` actually shapes and points `--for` users at `--signatures-only` / `--token-budget=N` / `--detail=N`.

2. **New WRONG → RIGHT table** beside the budget guidance, one row per reported first-run failure:
   - `--top-k` on `--for` (inert, warns, full bundle)
   - `--query` as a default lens (the binary's own help calls it debug; `skills/ripwire-navigate` already frames it as vocabulary hunting)
   - `--expand` without `--top-k=0` (the ~200-symbol ranked map rides along)
   - `--callers` on a framework route handler (empty by design — the decorator reaches it, not project code)
   - a directory-of-repos as the map root

3. **New "what ripwire does not replace" table**: route→handler lookup from a URL, templates/i18n/migrations, one exact string in a named file, UX flow through frontend event handlers.

4. **Skill files**: `ripwire-navigate` gets the budgeting-trap warning at the `--query`-vs-`--for` section (where an agent meets `--for` first); `ripwire-orient` step 6's `--top-k=50` tip now scopes itself to the default map.

## Explicitly not changed

The DISCLOSE-not-refuse behavior in `cli.h` (`noticeShapingFlagIgnored`) is deliberate — the help text itself says "documented, not fixed — a real fix is a behavior change." This PR documents the truth; it does not touch the behavior.

## Verification

- Every quoted string (`--top-k is not read by --for`, `raw BM25 ranking (debug); use --for`) confirmed verbatim in `src/cli.h` at HEAD.
- `test/readmedriftcheck.sh` failure count **identical before and after** this change (18 arms), run against the installed 0.3.8 binary — all 18 are pre-existing version skew between that binary's `--help` layout and HEAD's gates, not this diff. Local toolchain is g++-11 (too old for `<format>`), so a full HEAD build wasn't possible here; CI's own build should run the gates properly.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how `--top-k` affects default maps, queries, candidate formats, recall, and graph queries.
  * Documented that `--top-k` does not limit `--for`; instead, `--for` supports `--signatures-only`, `--token-budget`, and `--detail` for narrowing results.
  * Added guidance on common command-line mistakes, including query modes, expansion, route handlers, and multi-repository directories.
  * Added recommendations for when to use grep or direct file reading.


joyful-ii-V-I:
Hi @llvm-x86 — thank you for this. A field report from a real first-time session is exactly the kind of
evidence that should drive doc changes, and most of what's here holds up: I rebuilt HEAD and ran every
claim against the actual binary rather than taking the prose on faith.

**What's solid, with evidence:**
- The `--top-k`/`--for` inertness trap is real and exactly as described. `--for="…" --top-k=5` produces
  byte-identical output to the same call without `--top-k` (verified: 9202 bytes both ways), and the
  stderr note matches your quoted string verbatim. The suggested alternatives all actually narrow the
  bundle with no warning: `--signatures-only`, `--token-budget=N`, `--detail=N` all measured smaller and
  clean.
- The `--callers` route-handler row is correct. I built a small Flask fixture to check it independently:
  `--callers` on an `@app.route`-decorated handler returns `count="0"` by design, while a plain function
  the handler calls correctly shows up with `count="1"`.
- The `--query` vs `--for` guidance matches `--help` verbatim.

**What needs to change before merge (the doc would teach agents something false about the CURRENT binary):**

1. **The `--expand=SYM` row (README's WRONG→RIGHT table, third row) and the matching sentence in
   `ripwire-navigate/SKILL.md`'s new "Budgeting trap" paragraph are stale against today's `main`.** Since
   the `--expand` reorder (PR #289, already in the `main` this PR is built on), an *unambiguous* single-match
   `--expand=SYM` already defaults to `--top-k=0` on its own — disclosed as `topk_default="0"` in the
   header — specifically because the tool decided the ~200-row map is pure overhead once you've named an
   exact symbol. I measured it: `--expand=FILE:NAME` for an unambiguous symbol is 831 est_tokens by default
   vs 772 with an explicit `--top-k=0` — the map is already gone. The map only still rides along when SYM
   is *ambiguous* (a bare name with multiple definitions) — I measured that case too (`--expand=run`:
   25,641 tokens by default vs 16,475 with `--top-k=0`, a real ~9K gap matching your quoted figure), and
   for that case the binary already self-discloses the fix in its own output note. Please scope the row and
   the SKILL.md sentence to "when SYM is ambiguous" rather than presenting `--expand=SYM` unconditionally
   as a mistake — as written, it tells an agent to add a flag that's already the default for the common
   case, which undersells a fix that already shipped.

2. **The "directory of repos" row (README's WRONG→RIGHT table, fifth row) overstates what the tool does.**
   I pointed a built HEAD binary at a scratch directory containing two independent git repos as
   subdirectories (one positional root, the shape the row warns about) — it ran cleanly (`rc=0`), produced
   a normal merged map, and `--doctor` reported no failing check. There's no refusal, no warning, nothing
   in the source that matches "wrapper guards exist to refuse" — I grepped for it and found no such guard
   anywhere. The tool just silently crawls and merges everything under the root as one corpus, which is
   arguably worse than "refused," not better. Please either soften the "why" column to describe what
   actually happens (a silent over-broad merge, not a refusal) or drop the "wrapper guards exist to refuse"
   clause. Also worth a one-line distinction from the tool's real, deliberately-supported multi-root
   feature (`ripwire dir1 dir2 dir3 ... <verb>`, documented in `--help` and README:783 as an explicit,
   *multiple positional argument* invocation merging 2..16 checkouts) — as worded, a reader could come away
   thinking multi-root itself is the discouraged shape, when it's a first-class capability with a different
   invocation form (N explicit roots vs. one root that happens to contain nested `.git` dirs).

3. **CodeRabbit's actionable comment on this PR is a real, narrow precision gap worth folding into the same
   edit.** The shared rule as stated in README item 1 and the SKILL.md paragraph doesn't note that it's
   scoped to a *positive, explicit* `--top-k`: `--for --top-k=0` is a separate refusal (exit 1, a distinct
   "payload only" message, not the notice+full-bundle path), and `--for --format=candidates --top-k=N`
   consumes the flag without the notice. I verified both. The WRONG→RIGHT table row itself is fine as-is
   (it uses `--top-k=5`), but the unqualified prose above it should pick up the same qualifier CodeRabbit
   is asking for.

**Smaller items:**
- `readmedriftcheck.sh` and `docdriftcheck.sh` both pass clean on a proper build of this PR's head (I
  rebuilt and re-ran them rather than relying on the PR description's comparison against an old installed
  0.3.8 binary, which predates months of change and isn't meaningful evidence either way — worth mentioning
  since a future verification note on a doc PR should build HEAD if at all possible, even on an older
  toolchain, since g++-13+/clang with `<format>` is what CI actually gates on).
- No duplication or contradiction found against existing README.md, docs/COMMANDS.md, or other skills —
  these are genuinely new sections.
- The "what ripwire does not replace" table held up on every row I could check (no URL→handler verb exists
  anywhere in the CLI surface); nothing there looks like overclaiming or underselling.

**What happens next:** please update the two flagged rows (and the CodeRabbit-flagged qualifier) to match
current behavior, and this is in good shape — the underlying trap-hunting work is real and well-evidenced,
it just needs to track two things that changed the same week it was written. Happy to re-check the updated
wording once it's pushed.

llvm-x86:
Thanks — all three findings hold up, and I verified each against `src/` at this PR's head rather than
re-asserting the prose. Pushed as `e46349f`.

**1. `--expand=SYM` row was stale (your item 1).** Confirmed: `main.cpp:2185` gates
`exactNameExpandDefault` on `cfg.expand.size() == 1 && expandNodes.size() == 1 && !cfg.topKExplicit`, and
`cli.h:1285-1290` documents it as the EXACT-NAME DEFAULT with `topk_default="0"` on the root. So the row as
written told an agent to pass a flag that is already the default for the common case. Scoped the row and the
`ripwire-navigate` sentence to the **ambiguous bare name** case (which does keep the map — the same
"something to disambiguate" the help text names), and added the cheaper right-hand option of naming the
symbol exactly (`--expand=FILE:NAME`).

**2. "the wrapper guards exist to refuse" was wrong (your item 2).** Confirmed by grep: the only matches for
repo-of-repos language anywhere in the tree are the README row itself, an unrelated nested-`.gitignore`
comment in `ingest_crawl.h:1266`, and a `multirootcheck.sh` comment about fixture layout. There is no guard.
(The clause was leaking a habit from a *local* wrapper script on my box, which is not this project's
behavior and had no business in this README.) The row now says what actually happens — a silent walk and
merge of the nested checkouts into one corpus — and carries the distinction you asked for: that is ONE root
containing checkouts, not the first-class multi-root form of N **explicit** positional roots (2–16, README:2758).

**3. CodeRabbit's qualifier (your item 3).** Confirmed in `noticeShapingFlagIgnored` (`cli.h:3992-3996`):
`isTopKConsumedBeside = c.candidates || c.topK == 0` suppresses the notice, and the comment block above it
documents both suppressions as fixes for false notices. `--for --top-k=0` instead lands on the payload-only
refusal at `cli.h:4957-4962` (`c.ok = false`, distinct message), so it is an exit-1, not warn-and-emit.
README item 1 and both skill paragraphs now scope the rule to a *positive, explicit* `--top-k` on *plain*
`--for` and name the two neighbours that behave differently.

**On the verification note.** Fair hit, and I've kept the correction honest rather than restated it: this box
has g++-11 and clang-14 only, so `<format>` still puts a HEAD build out of reach here and
`readmedriftcheck.sh` / `docdriftcheck.sh` correctly refuse without `build/ripwire`. I am therefore **not**
claiming a green gate run for this push — the evidence above is source-level at HEAD (line-cited), plus your
own clean run of both checks on the head this commit sits directly on top of, and the diff is prose inside the
same three sections. Worth flagging because the failing shape you diagnosed — comparing against an installed
0.3.8 binary and presenting the count as evidence — was the actual defect in my first verification note, and
substituting a different unverified claim would repeat it.

Ready for your re-check.

joyful-ii-V-I:
Hi @llvm-x86 — by the time we picked this up you'd already pushed the exact follow-up the review asked
for: the `--expand=SYM` row scoped to the ambiguous case, the directory-of-repos row corrected to
describe the silent merge instead of a nonexistent guard, and the CodeRabbit qualifier folded into both
the README prose and the SKILL.md paragraph. We rebuilt HEAD and re-ran `readmedriftcheck` and
`docdriftcheck` against your corrected text rather than taking it on trust — both ALL PASS.

Landed on `lane/t13-contrib-finish`, your authorship preserved on both commits (the original addition
and the correction). Thank you for building the Flask fixture and chasing the `--expand` default down to
the disclosed `topk_default="0"` — that's exactly the kind of field-verified doc fix this project wants.

PR #306, opened 2026-09-20
Integration train 13: three contributions completed, four correctness fixes

joyful-ii-V-I:
Integration train 13. Three outside contributions carried to completion, and four correctness fixes.

**Contributions** — each contributor's commits are preserved with their authorship; our completions are separate commits.
- **#291, @pt-act** — the advertised MCP verb roster is now pinned by DERIVING it from `tools/list` (33 verbs), which catches a swap the existing count check cannot see: rename a verb while holding the count and the old arm stays green. Verified by planting an add, a remove and a count-preserving rename.
- **#298, @s0undt3ch** — the GNU-vs-BSD `stat` trap centralised into one helper the gates share (it fails by *succeeding* with the wrong data, which is worse than an error), plus the ambient-`HOME` leak: the review reproduced a gate writing 19 real files into a developer's own config directory, and the same shape is now closed in six more gates.
- **#302, @llvm-x86** — a field report from a real first session. Two claims were corrected against the built binary: `--expand=SYM` now defaults to the focused form by itself, and the "directory of repos" row claimed guards that do not exist.

**Correctness fixes**
- **`--dead-code` no longer stamps `confidence="high"`.** It was a hardcoded literal on every finding — a claim the code does not support. Removed rather than decorated, with four docs and skills that asserted it corrected; a skill is agent-consumed ground truth, so a stale one teaches the wrong thing.
- **`reuse-decline` gains the two demotions its sibling reporter already had** (the all-test-script skip and idiom → minor), so it stops firing on shell-gate harness clones — the exact class `duplication` exempts.
- **An `import` is no longer a caller in Java and Kotlin.** `queries/{java,kotlin}/tags.scm` captured it as `@reference.call`, so `--callers` and `--impact` counted an import as a call site. It is now `@reference.import`: still visible as an import, never an edge. **`kParserVer` 118 → 119**, and a cache written by the previous binary is proved refused cold and warm. The Kotlin fixture now reports zero file-scope owners, which is correct — a `.kt` file has no top-level executable code, and the phantom owner existed only because of this miscapture.
- **`--slice-flow=both` discloses `flow_redundant`** when unseeded, where it provably reaches no line the flat rows do not.

**Train work.** One conflict, in `queries/java/tags.scm`, where train 12 had recorded this exact question as open beside the line; resolved toward the fix, with train 12's note kept as the record of what it answers. Goldens re-derived through their own procedures and listed with reasons. Gate selection used train 12's rule — 222 candidate gates from the merge's 48 files, 169 run locally, 7,733 assertions green — which is why this train needed no CI round to find its own gaps.

**Checks:** determinism byte-identical, `xmllint` clean, ASan clean over 7 invocations, `--quality-delta=origin/main..HEAD` gates 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Java and Kotlin imports are now reported as dependency edges rather than call-graph references, improving caller and impact results.
  * Dead-code findings no longer claim a fixed “high” confidence level; reports identify the supporting evidence instead.
  * Unseeded bidirectional slice-flow results now disclose when the flow adds no information.

* **Documentation**
  * Clarified `--top-k`, `--for`, `--signatures-only`, `--token-budget`, and `--detail` behavior.
  * Updated guidance for dead-code analysis, safe deletion, import relationships, troubleshooting, and slice-flow output.


joyful-ii-V-I:
Review pass answered.

**Post-review commits on this PR** (after the review of 34ef6332):
- `dd9ca946` — clear inherited Git location variables in the shared test helper; new gate `test/gitenvhermeticcheck.sh`; two gates re-pinned with reasons.

**The one finding was correct, and it generalises.** `git -C` selects a directory, not a repository, so with `GIT_DIR` set the gate's `git init` never created `$REPO/.git` at all — the fixture's commit landed in the ambient repository and `rev-parse HEAD` read that foreign sha back, while the gate still reported ALL PASS. **155 of the 157 repo-building gates were exposed.** The two that were not had each hand-rolled a *different* partial list, neither carrying the object-directory names, so the clearing now lives in `test/lib/clean-env.sh` (the helper #298 introduced) and both copies fold onto it. The new gate proves the defect is live, proves the helper fixes it, and sweeps the tree for any repo-building gate that does not source it.

The wider sweep caught two more reds: `qdrefpaircheck` (E) re-pinned 9 → 8, where the row that left is exactly the all-test-script clone group this train's `reuse-decline` fix now exempts; and a docs gate reading a bare flag in the changelog prose.

CI was green at 34ef6332 (31 passed, 2 skipped). On the fix head: 272 gates green with 11,749 assertions, determinism byte-identical, `xmllint` clean, ASan clean, `--quality-delta` gates 0.

